PR #24, opened 2020-01-02
Allow re-uploading recordings from clients

tabjy:
Adds client functionalities for rh-jmc-team/container-jfr#55.

Blocks rh-jmc-team/container-jfr#90.


tabjy:
@andrewazores I've updated to move the upload button to the archived recording page and trigger a refresh upon the upload successes.

Issue #23, opened 2019-12-24
Auth token dialog should be customizable for deployment types


andrewazores:
The current auth token dialog contains a message like "for example, if this is an OpenShift deployment", and a blurb about how to get an appropriate access token to paste in. This message should come from container-jfr and be tailored to the deployment platform.

andrewazores:
In progress: #29

andrewazores:
Closing as obsolete

PR #29, opened 2020-01-13
#23: Auth dialog messages should have tailored messages

tabjy:
I'm not very familiar with angular and reactive programming. Please advise if I did anything against the convention. Thank you!

Blocked by rh-jmc-team/container-jfr#101

### DIFF
--- a/src/app/documentation-messages.service.ts
+++ b/src/app/documentation-messages.service.ts
@@ -56,15 +56,15 @@ export class DocumentationMessagesService {
     return locales
       .filter((v, i) => locales.indexOf(v) === i) // remove duplicates
       .map((l, i) => {
-      const q = 1 - i * 0.1;
-      if (q === 1) {
-        return l; // ';q=1' can be omitted
-      } else if (q <= 0) {
-        return l + ';q=0';
-      } else {
-        return l + `;q=` + q;
-      }
-    }).join(',');
+        const q = 1 - i * 0.1;
+        if (q === 1) {
+          return l; // ';q=1' can be omitted
+        } else if (q <= 0) {
+          return l + ';q=0';
+        } else {
+          return l + `;q=` + q;
+        }
+      }).join(',');
   }
 
   isReady(): Observable<boolean> {
--- a/src/app/documentation-messages.service.ts
+++ b/src/app/documentation-messages.service.ts
@@ -16,33 +16,14 @@ export class DocumentationMessagesService {
     private http: HttpClient,
     private notifications: NotificationService,
   ) {
-    this.setLocale(navigator.language);
+    this.setLocale(navigator.language); // use the UA language by default
   }
 
-  setLocale(locale: string = navigator.language) {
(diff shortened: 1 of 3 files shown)


src/app/documentation-messages.service.ts
```diff
@@ -0,0 +1,69 @@
+import { Injectable } from '@angular/core';
+import { HttpClient, HttpErrorResponse } from '@angular/common/http';
+import { NotificationService, NotificationType } from 'patternfly-ng';
+import { BehaviorSubject, Observable } from 'rxjs';
+import { filter, map } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root'
```
andrewazores: Neat, I hadn't seen this feature before. Probably worth updating the other services to do the same thing. I'll create an issue for it.

src/app/documentation-messages.service.ts
```diff
@@ -0,0 +1,69 @@
...
+  private readonly ready = new BehaviorSubject<boolean>(false);
+
+  constructor(
+    private http: HttpClient,
+    private notifications: NotificationService,
+  ) {
+    this.setLocale(navigator.language);
+  }
+
+  setLocale(locale: string = navigator.language) {
```
andrewazores: Hmm, setting a default value on `locale` seems funny. Why would a consumer of this API need to call `setLocale()` with no args? Is that intended as a "reset" call?

src/app/documentation-messages.service.ts
```diff
@@ -0,0 +1,69 @@
...
+        locales.push(lang);
+        locales.push(lang.split('-')[0]);
+      }
+    });
+    const range = locales.map((v, i) => {
+      if (i > 10) {
+        i = 10;
+      }
+      return `${v};q=${1 - i * 0.1}`;
+    }).join(',');
```
andrewazores: I think this section above needs some comments explaining what it's doing and why, or just to be split into some smaller functions so that the function names can serve as documentation. I'm assuming it's because you can have locale names like "en-CA" and "en-US", and you're trying to process both of those into "en"? But beyond that I'm not really sure what's happening.

src/app/documentation-messages.service.ts
tabjy: Ah, when writing javascript, I have a (potentially bad) habit of setting default values that make little sense. Thank you for pointing out.

In the future, if we ever need the ability to reset to browser's default language, a dedicated `reset()` function should be introduced. Or just explicitly call `.setLocale(navigator.langauge)`. :)

src/app/documentation-messages.service.ts
tabjy: The confusing part is for assigning the q-factor weightings, which are the priorities, to each locale.

I've updated to extract this process into a separate function with proper documentation.

andrewazores:
Looks good. I'll hold off on merging this until I review the corresponding container-jfr PR.

andrewazores:
Closing as obsolete

PR #43, opened 2020-03-13
Detect auth scheme and display auth prompt accordingly

### LINKED ISSUE #38: HTTP Basic auth should be supported
Currently the API service and command channel service both only support token auth. That is, the API service only knows how to send `Authorization: Bearer FOO` headers, and the command channel only knows the `base64url.bearer.authorization.containerjfr.FOO` subprotocol. They should support `Authorization: Basic` headers and `basic.authorization.containerjfr` subprotocols, respectively. Related: #37 , since these two additions also require base64 encoding of the credentials. This is also related to #23 since there should be some tailored message instructing the user how to input their basic auth credentials.

andrewazores:
Use X-WWW-Authenticate response header sent by container-jfr on POST
/auth request to determine Basic or Bearer auth scheme, and display an
authorization prompt to the user according to this scheme. Also adjust
API and Command Channel auth protocols for the detected scheme.

Fixes #38

### DIFF
--- a/src/app/command-channel.service.ts
+++ b/src/app/command-channel.service.ts
@@ -1,6 +1,6 @@
 import { Injectable, OnDestroy } from '@angular/core';
-import { Subject, BehaviorSubject, Observable, ReplaySubject, combineLatest } from 'rxjs';
-import { filter, first, map } from 'rxjs/operators';
+import { Subject, BehaviorSubject, Observable, ReplaySubject } from 'rxjs';
+import { filter, first, map, combineLatest } from 'rxjs/operators';
 import { HttpClient } from '@angular/common/http';
 import { NotificationService, NotificationType } from 'patternfly-ng/notification';
 import { ApiService } from './api.service';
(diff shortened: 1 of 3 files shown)


src/app/command-channel.service.ts
```diff
@@ -68,9 +68,18 @@ export class CommandChannelService implements OnDestroy {
...
     this.api.getToken()
-      .pipe(first())
-      .subscribe(token => {
-        this.ws = new WebSocket(clientUrl, `base64url.bearer.authorization.containerjfr.${btoa(token)}`);
+      .pipe(
+        combineLatest(this.api.getAuthMethod()),
+        first()
+      )
+      .subscribe(auths => {
+        let subprotocol = undefined;
```
tabjy: This feels a bit weird... but it makes sense. Just to point out for the 2nd arg of `WebSocket` constructor, `null`, `undefined` and `""` are equivalent.

andrewazores: Yea I felt a bit weird about it too, but using the empty string or null seemed like it might look like a mistake on later review. `undefined` was weird enough to look intentional, I guess.

PR #45, opened 2020-03-23
Allow finer-grained specification of recording duration

### LINKED ISSUE #35: Support user-inputted times
The UI allows selection from a list of times. Instead it should allow users to input a time; something like:

time := \<number\>\<unit\> | 'unlimited'

number in set of Integers greater than 0
unit in {s, m, h, d}

e.g.

5s
30m
2h
1d
etc.

andrewazores:
Fixes #35

### DIFF
--- a/src/app/create-recording/custom-recording/custom-recording.component.html
+++ b/src/app/create-recording/custom-recording/custom-recording.component.html
@@ -3,23 +3,31 @@
     <label for="recordingName">Name</label>
     <input id="recordingName" type="text" class="form-control" [(ngModel)]="name" name="recordingName" placeholder="Name" (keyup.enter)="submit()"/>
   </div>
-  <div class="form-group">
-    <label for="recordingDuration">Duration</label>
-    <select id="recordingDuration" [(ngModel)]="duration" name="recordingDuration" class="form-control">
-      <option value="-1">Unlimited</option>
-      <option value="5">5 seconds</option>
-      <option value="30">30 seconds</option>
-      <option value="60">60 seconds</option>
-      <option value="300">5 minutes</option>
-    </select>
+  <div class="form-group row">
+    <div class="col col-xs-6">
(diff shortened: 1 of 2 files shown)


andrewazores:
Before:
Screenshot from 2020-03-23 14-25-23

After:
Screenshot from 2020-03-23 14-21-59
Screenshot from 2020-03-23 14-22-20

andrewazores:
Also, since the controls in that modal dialog are set up within a form, the user can navigate between fields with `<Tab>` (and `<Shift>+<Tab>` to step back). When the numeric input has focus the user can simply type in their numeric value with the keyboard, or they also have the option of using the graphical buttons to increment/decrement the value. Tabbing over or clicking the units control, when it is enabled (for all durations > 0) allows the user to select from `Seconds|Minutes|Hours`. This can again be done by clicking the graphical dropdown elements, or by using the keyboard to type a case-insensitive prefix match for the unit desired (ie pressing 'h' will select `Hours`).

src/app/create-recording/custom-recording/custom-recording.component.html
```diff
@@ -3,23 +3,31 @@
...
-    <label for="recordingEvents">Events <span class="pficon-info" [tooltip]="tooltipTemplate" placement="right"></span></label>
+    <label for="recordingEvents">Events <span class="pficon-info" [tooltip]="eventTooltipTemplate" placement="right"></span></label>
     <input id="recordingEvents" type="text" class="form-control" [(ngModel)]="events" name="recordingEvents" placeholder="events" (keyup.enter)="submit()"/>
   </div>
   <button type="submit" class="btn btn-primary" (click)="submit()">Submit</button>
 </form>
 
-<ng-template #tooltipTemplate>
+<ng-template #durationTooltipTemplate>
+  <span>&lt;= 0 specifies an "unlimited" duration, continuous recording</span>
```
jiekang: Does this require any API change on the backend?

andrewazores: No, that's already how it worked.

PR #48, opened 2020-03-30
Templates

### LINKED ISSUE #47: UI for event templates
Related to #34 and #36

Some kind of UI should be implemented allowing for users to easily discover which templates are available and what those templates might be useful for, along with some way to easily select these templates when creating a new recording. This should still allow the user to input their own customized event specifier strings as well.

andrewazores:
This PR adds some UI for event templates. The new UI is a table within a tab, belonging to a tabset which now contains the previously existing event types listing table. The new table displays the name, description, and provider of known event templates, and allows for filtering. Templates can be either manually specified in the existing recording creation dialog using `template=Foo` syntax, or they can be created by clicking on the `+` button to the left of the template name in each table row. When this button is clicked the recording creation modal dialog is summoned, displaying only the relevant "Custom Recording" tab, and with the `template=Foo` specifier pre-filled.

Fixes #47

container-jfr-web-template-ui


src/app/command-channel.service.ts
```diff
@@ -112,6 +113,15 @@ export class CommandChannelService implements OnDestroy {
           NotificationType.WARNING, 'WebSocket Error', JSON.stringify(evt), false, null, null
         ));
 
+        // FIXME handle the case of a failed 'connect' command
```
jiekang, line 116: What's the impact and workaround if any of a failed connect command? Would it be warranted to open an issue to track this?

andrewazores: It'll just be basically silently swallowed, and not reflected by the `isConnected` Observable. I need to come back around to this and think about how the whole `isReady/isConnected` is handled, because this error handling also exposes the need for an `isErrored` kind of status, but that's all going to be a bit awkward for components to actually consume. Especially because currently `isReady` encapsulates the state of the WebSocket command channel connection itself, whereas `isConnected` is about the state of the connection from container-jfr to the target, so the error state would need to differentiate between these connections.

Maybe what I should do is refactor it again (later) to be an `Observable<ConnectionState>`, and `ConnectionState` should be an enum of all the types of states the connection could be in, ie `no command channel connection | command channel connected but no target | command channel and target connected | command channel error | target connection error`. Not really sure, needs more thought. I'll file an issue for it.

andrewazores: Filed: #49

jiekang: Okay, thank you!

andrewazores:
I just noticed that the connection state refactor I already made to pull that logic into the command channel service, instead of being mostly in the recording list component, has introduced a new bug. If you have an active connection between container-jfr and the target JVM, and then refresh the client page (or open a new client instance), it will properly connect to container-jfr and detect the existing target connection but will *not* update the UI to display the recording list or event tables components. I'm looking into it now.

andrewazores:
Fixed that bug in the second-to-last commit that I just pushed. Turned out to be my misremembering of how I implemented the `is-connected` command, actually, so a very simple fix.

jiekang:
@andrewazores When testing locally, on connection to a previously unconnected jvm, I see the warning "disconnect: command disconnect unavailable". Any ideas what causes that?

jiekang:
Scenario:
When starting from the initial state of a connected JVM, the templates are visible. As a user, if I wish to see events, I click on the Events tab, but wait, I need to click again to show the table.

Now that there are two tabs, I don't think it makes sense for the Events table to start closed. Maybe neither of them need to have that open/close feature at all? What do you think?

PR #50, opened 2020-04-09
React re-write

andrewazores:
This PR is an initial effort to rewrite the web client using Patternfly 4/React, replacing Patternfly 3/Angular. There are many missing features in this initial PR - not at all at parity with the previous Angular implementation. However, I thought it made sense to get it into a very basic runnable and loosely usable state and then enhance it further with subsequent PRs to bring it up to parity.

container-jfr-web-react


jiekang:
Seems functional as described.

Given the scope of the changes here and the feature difference, is there any value in running a temporary upstream feature branch for react that can be worked against. Maybe call the branch 'react-migration', or similar. What do you think? Once it is matured, we can switch merge it on top of master. I guess we don't have an actual release yet so it's fine to drop this into master now?

As well, we should discuss the overall UX for the features we want to support before doing more. The workflow for this new UX is a bit strange to me with the Target Selection now separated. It's a lot more clicks to switch between active connections from the Recordings page. I don't want to impose anything on this initial work though as I agree it makes sense to get some code in.

andrewazores:
Yea, I'd be willing to keep this PR open as the eventual meta-PR for the React migration. Further enhancements/new features/etc. can be done by PRs into this branch, and this branch would only be merged once we think it's ready to replace the Angular implementation.

Agreed on the UX planning front. I wanted to use the nice premade patternfly-react left side nav drawer, which is also just generally the right way to do nav in Patternfly, and thought it made sense to put features like the recordings list, event types list, event templates list, etc. into that nav - certainly better than just infinitely appending new features to the bottom of a vertically-scrolling page as the Angular implementation does. But then we do get the weirdness you point out where target selection is also a separate nav item, which does feel weird. Maybe each feature page (ex. recordings list, event templates list) should also have a target selection widget at the top? They all share global state through the command channel service, so that should work fine.

PR #51, opened 2020-04-09
Add TargetSelect widget on RecordingList

andrewazores:
container-jfr-web-react-targetselect

This PR modifies the TargetSelect component so that it can be reused, reinstates the Dashboard main page component containing a TargetSelect, and adds a TargetSelect to the RecordingList. The intent is to allow users to quickly switch targets while remaining on the same content view, without needing to click back and forth between the previous standalone target selection view.


andrewazores:
Here's a video of the latest version. I think I'm done committing on this PR now.

container-jfr-web-react-targetselect-2

PR #52, opened 2020-04-14
Split Login component into React Form components

andrewazores:
Basic login form asks for separate username and password, Bearer form
asks for just auth token

Additionally, use Card styling

Before:
container-jfr-web-react-login-before

After:
container-jfr-web-react-login-after


andrewazores:
Nope, no specific plans there. That's just part of the seed project, which I haven't gotten around to removing.

PR #54, opened 2020-04-24
Initial Events view

### LINKED ISSUE #55: View event types
Supported event types should be viewable in a list or table view.

https://projects.invisionapp.com/share/AZX0VIK5EKM#/screens

### LINKED ISSUE #56: View event templates
Supported event templates should be viewable in a list or table view.

https://projects.invisionapp.com/share/WGX0VI6SPBM#/screens

andrewazores:
This PR adds an initial pass at the Events views outlined in #55 and #56 .

The Templates view is missing the ability to select rows and create recordings from these templates, since the view for creating recordings in general is not yet implemented (see #59).

~~The Types view needs to be made scrollable or to gain pagination, since the number of events is quite long and currently they are all rendered into an extremely large table.~~

image
image
image

Fixes #55 
Fixes #56

### DIFF
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -24,10 +24,11 @@ export const EventTemplates = (props) => {
 
   const getTemplates = () => {
     let filtered;
-    if (!filter) {
+    if (!filterText) {
       filtered = templates;
     } else {
-      filtered = templates.filter((t: EventTemplate) => t.name.toLowerCase().includes(filterText) || t.description.toLowerCase().includes(filterText) || t.provider.toLowerCase().includes(filterText));
+      const ft = filterText.trim().toLowerCase();
+      filtered = templates.filter((t: EventTemplate) => t.name.toLowerCase().includes(ft) || t.description.toLowerCase().includes(ft) || t.provider.toLowerCase().includes(ft));
     }
     return filtered.map((t: EventTemplate) =>
       [ t.name, t.description, t.provider ]
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { filter, map } from 'rxjs/operators';
+import { TextInput, Toolbar, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
 import { Table, TableBody, TableHeader } from '@patternfly/react-table';
 import { ServiceContext } from '@app/Shared/Services/Services';
 
@@ -13,6 +14,7 @@ export const EventTemplates = (props) => {
   const context = React.useContext(ServiceContext);
 
   const [templates, setTemplates] = React.useState([]);
+  const [filterText, setFilterText] = React.useState('');
 
   const tableColumns = [
     'Name',
@@ -21,10 +23,16 @@ export const EventTemplates = (props) => {
   ];
 
   const getTemplates = () => {
-    return templates.map((t: EventTemplate) => {
-      return [ t.name, t.description, t.provider ];
-    })
-  }
+    let filtered;
+    if (!filter) {
+      filtered = templates;
+    } else {
+      filtered = templates.filter((t: EventTemplate) => t.name.toLowerCase().includes(filterText) || t.description.toLowerCase().includes(filterText) || t.provider.toLowerCase().includes(filterText));
+    }
+    return filtered.map((t: EventTemplate) =>
+      [ t.name, t.description, t.provider ]
+    );
+  };
 
   React.useEffect(() => {
     const sub = context.commandChannel.onResponse('list-event-templates')
@@ -40,11 +48,18 @@ export const EventTemplates = (props) => {
     context.commandChannel.sendMessage('list-event-templates');
   }, []);
 
-  return(
+  return(<>
+    <Toolbar>
+      <ToolbarGroup>
+        <ToolbarItem>
+          <TextInput name="templateFilter" id="templateFilter" type="search" placeholder="Filter..." aria-label="Event template filter" onChange={setFilterText}/>
+        </ToolbarItem>
+      </ToolbarGroup>
+    </Toolbar>
     <Table aria-label="Event Templates table" cells={tableColumns} rows={getTemplates()}>
       <TableHeader />
       <TableBody />
     </Table>
-  );
+  </>);
 
 }
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { filter, map } from 'rxjs/operators';
 import { TextInput, Toolbar, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
-import { Table, TableBody, TableHeader } from '@patternfly/react-table';
+import { Table, TableBody, TableHeader, TableVariant } from '@patternfly/react-table';
 import { ServiceContext } from '@app/Shared/Services/Services';
 
 export interface EventTemplate {
@@ -56,7 +56,7 @@ export const EventTemplates = (props) => {
         </ToolbarItem>
       </ToolbarGroup>
     </Toolbar>
-    <Table aria-label="Event Templates table" cells={tableColumns} rows={getTemplates()}>
+    <Table aria-label="Event Templates table" cells={tableColumns} rows={getTemplates()} variant={TableVariant.compact}>
       <TableHeader />
       <TableBody />
     </Table>
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -1,0 +1,50 @@
+import * as React from 'react';
+import { filter, map } from 'rxjs/operators';
+import { Table, TableBody, TableHeader } from '@patternfly/react-table';
+import { ServiceContext } from '@app/Shared/Services/Services';
+
+export interface EventTemplate {
+  name: string;
+  description: string;
+  provider: string;
+}
+
+export const EventTemplates = (props) => {
+  const context = React.useContext(ServiceContext);
+
+  const [templates, setTemplates] = React.useState([]);
+
+  const tableColumns = [
+    'Name',
+    'Description',
+    'Provider',
+  ];
+
+  const getTemplates = () => {
+    return templates.map((t: EventTemplate) => {
+      return [ t.name, t.description, t.provider ];
+    })
+  }
+
+  React.useEffect(() => {
+    const sub = context.commandChannel.onResponse('list-event-templates')
+      .pipe(
+        filter(m => m.status === 0),
+        map(m => m.payload),
+      )
+      .subscribe(templates => setTemplates(templates));
+    return () => sub.unsubscribe();
+  }, []);
+
+  React.useEffect(() => {
+    context.commandChannel.sendMessage('list-event-templates');
+  }, []);
+
+  return(
+    <Table aria-label="Event Templates table" cells={tableColumns} rows={getTemplates()}>
+      <TableHeader />
+      <TableBody />
+    </Table>
+  );
+
+}
--- a/src/app/Events/EventTypes.tsx
+++ b/src/app/Events/EventTypes.tsx
@@ -132,7 +132,7 @@ export const EventTypes = (props) => {
       <SplitItem isFilled />
       <SplitItem>
         <Pagination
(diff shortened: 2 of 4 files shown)


jiekang:
Does lint/format still work out for these code additions?

`npm run lint` has a lot of output atm, and I think `npm run format` does change some things.

src/app/Events/EventTypes.tsx
```diff
@@ -0,0 +1,140 @@
...
+        setOpenRow(rowKey);
+      } else {
+        setOpenRow(rowKey > openRow ? rowKey - 1 : rowKey);
+      }
+    } else {
+      setOpenRow(-1);
+    }
+  };
+
+  // TODO replace table with data list so collapsed event options can be custom formatted
```
jiekang, line 115: Is this TODO still relevant? I'd like to see an issue updated about this, or filed for this if appropriate. Maybe this TODO can link the issue, up to you for that part.

andrewazores: It is still relevant. I'll file a new issue for it, along with start working on a mockup of what the expanded rows could look like.

andrewazores:
I just pushed a commit that includes the small changes made by `npm run format`. There are lot more linting issues that are not fixed by that, so I'll tackle those in a separate PR if that's okay.

jiekang:
Yeah sounds good

src/app/Events/EventTypes.tsx
```diff
@@ -0,0 +1,140 @@
...
+      if (includesSubstr(t.name, filter)) {
+        return true;
+      }
+      if (includesSubstr(t.typeId, filter)) {
+        return true;
+      }
+      if (includesSubstr(t.description, filter)) {
+        return true;
+      }
+      if (t.category.some(c => includesSubstr(c, filter))) {
```
jiekang, line 89: This filter setup doesn't seem to work for targeting multiple categories, for example, both 'Java Virtual Machine' and 'GC'. My attempts were "Java Virtual Machine, GC" && "Java Virtual Machine GC". The categories are displayed as a comma separated list, so I'd hope the comma separated list filter would work too.

src/app/Events/EventTemplates.tsx
```diff
@@ -0,0 +1,65 @@
...
+    'Description',
+    'Provider',
+  ];
+
+  const getTemplates = () => {
+    let filtered;
+    if (!filter) {
+      filtered = templates;
+    } else {
+      filtered = templates.filter((t: EventTemplate) => t.name.toLowerCase().includes(filterText) || t.description.toLowerCase().includes(filterText) || t.provider.toLowerCase().includes(filterText));
```
jiekang, line 30: I think the lowerCase setup is mixed up. Filter works for "profiling" but not "Profiling". The name is displayed as "Profiling" so I'd expect the capitalized one to work, and then ideally both to work.

src/app/Events/EventTypes.tsx
andrewazores: Right, it doesn't. It's just a simple substring match, so you get an implicit `or` of categories anyway, ie filtering will narrow down the list as long as the event has at least one matching category.

src/app/routes.tsx
```diff
@@ -41,11 +42,18 @@ const dynamicRoutes: IAppRoute[] = [
...
     label: 'Recordings',
     path: '/recordings',
-    title: 'Flight Recordings'
+    title: 'Recordings'
   },
+  {
+    component: Events,
+    exact: true,
+    label: 'Events',
+    path: '/events',
```
jiekang, line 51: Maybe this is local only has to do with CORs or something? I tested this with container-jfr pointing to your work @ https://github.com/rh-jmc-team/container-jfr/pull/148

If I visit / and click events, it brings me to /events fine

If I take the url from /events (ip-addr:port/events) and visit it in another tab, it shows "Resource not found"

I'd expect it to open up the events page. Any ideas what's going on?

andrewazores: Not sure, can't reproduce it. Initially the Events view, as well as the Recordings view, should produce a "Resource not found" because at startup they really aren't defined, because there is no command channel or target connection. But once the application establishes a command channel connection and sees that there is also a target connection then the routes should be made available. That's specifically implemented in `routes.tsx` and seems to work on my end.

src/app/Events/EventTypes.tsx
andrewazores: Are you trying to get it to do an `and` filtering? That would probably require a more complex filtering setup than the current simple text box. It could be filed as a future enhancement.

jiekang: Hmm, well the category list displays text e.g. 'Java Development Kit, Security'

It's unclear to a user why the same text in the filter would not match anything. I think the most simple pass would also check against the text displayed, such that the comma-separated list would work. Maybe you can construct that for checking against or save it somewhere to be reused?

jiekang: Basically, without any use of explicit && || type operators in the text box, I'd expect it to be a text search, with the text that is displayed to me being searched.

andrewazores: Sure, I could do that, but it wouldn't match "Security, Java Development Kit" if some other event happens to have that list of categories (same members, different order).

andrewazores: (although searching for each substring separately would of course still work)

src/app/routes.tsx
jiekang: Huh weird... reproducer steps should be, on:
https://github.com/rh-jmc-team/container-jfr/pull/148
with web-client @ this pr.

$ MAVEN_OPTS="-Dcom.sun.management.jmxremote.port=9091 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false" mvn clean prepare-package exec:java

(Using Google Chrome)
* Visit HTTP service @ e.g. http://192.168.0.27:8181 and target container-jfr
* Click on events to get to: http://192.168.0.27:8181/events
* Copy-paste http://192.168.0.27:8181/events into a new tab and hit enter. See "Resource not found"

I notice if I vist http://192.168.0.27:8181 in a new tab, it is already set to be connected to container-jfr and changes to the target are seen by both tabs. Would visiting Events view directly miss some of the code that gets the target?

src/app/Events/EventTypes.tsx
jiekang: I think that sounds fine for now. The category text looks sorted fortunately, so at least there won't be event type with Security & JDK, but in different order.

jiekang: That would match how basic text search works anyways I think?

src/app/routes.tsx
andrewazores: Ah - I see. It's broken in the way you describe when you do an actual client build and serve it from ContainerJFR, but it works as I intended when using Webpack Dev Server. That's unfortunate. I'll have to do some research to see how to get the web client running and preserve the path to pass to its internal routing. I'll file a separate issue.

andrewazores: In other words - this is actually a ContainerJFR webserver problem, not a ContainerJFR Web problem

jiekang: Okay, sounds good. Please resolve this conversation when an issue is filed.

andrewazores: #64

src/app/Events/EventTypes.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  // TODO replace table with data list so collapsed event options can be custom formatted
+  return(<>
+    <Split>
+      <SplitItem>
+        <TextInput name="eventFilter" id="eventFilter" type="search" placeholder="Filter..." aria-label="Event filter" onChange={setFilterText}/>
+      </SplitItem>
+      <SplitItem isFilled />
+      <SplitItem>
+        <Pagination
```
jiekang, line 127: The pagination code seems to mess up on specific switches between number of items to display.

E.g.

1. Default is 10. Page twice to items 21-30 out of 138
2. Switch to show 100. It states showing items 101-138 but no items are visible.

andrewazores: Good catch. I'll work on this.

andrewazores: Hmm. The issue seems to be that I'm trying to calculate the offset for which events to display on the single rendered page, and doing this reactively when any of the parameters (current page number, count per page, etc.) change. The problem is that when the changed parameter is the count per page we only know what the new value is, so the calculation of the current offset overshoots because it's already using the new count per page.

A simple solution is to simply reset the view to no offset (ie items 1-n) whenever the count per page is changed. The user will lose their place if they change the count per page after paging over. A more complex solution means adding some more state and calculation to figure out what changed, map the index of the previous first recording in view to its new index in the resized page, and shift to that offset. Do you think it's worthwhile implementing all that?

andrewazores:
Rebased.

src/app/Events/EventTypes.tsx
jiekang: To clarify, in the case of:

CPP = 10 : Page 2

When a user switches to CPP = 100, we don't know that the previous CPP was 10, just that we're on page 2?

andrewazores: Pretty much. The rendered page is calculated in a React hook, and that hook is called whenever any of those parameters is updated. The hook also only knows the current values of each parameter. So if we're on Page=2 and CPP=10 and the user goes to the next page, we calculate an offset based off of Page=3 and CPP=10 and it works as expected. But if they change the CPP then we try to give them a view based on that CPP and the previous page number - if we have 138 items then Page=3 and CPP=100 doesn't make sense, it's beyond the range of the actual backing array, so we end up rendering nothing.

andrewazores: Maybe we can hack it by checking if the offset exceeds the length of the backing array, and if so, set the current page number to the last one that would be in range?

jiekang: Is it possible to add state for a single int to store CPP as updates occur? Then if it exists, we can use it as the previous CPP to calculate the correct page.

andrewazores: That would basically just be another `useState` hook like the existing one for CPP, except we would have to do something hackish like only set its value in the update hook, and make the update hook not depend on that parameter's value, even though it does actually use it. Not great IMO, and goes against the best practices recommended by React since we would be intentionally using a piece of state within the update hook and not updating when that piece of state changes.

jiekang: This sounds weird. It's natural for state change results depend on what the previous state was. Usually state change actions includes (from, to)...

jiekang: Not sure I followed the last line "not updating when that piece of state changes.". We'd be running the same update hook, so only acting when a piece of state changes, and in this hook we'd check the previous state stored, run our calculation, store the new state and continue as before.

andrewazores: > This sounds weird. It's natural for state change results to depend on what the previous state was. Usually state change actions includes (from, to)...

Yea, you *can* get that with React state, but I would need to rewrite this entire component as a class component instead of a functional component.

> Not sure I followed the last line "not updating when that piece of state changes.". We'd be running the same update hook, so only acting when a piece of state changes, and in this hook we'd check the previous state stored, run our calculation, store the new state and continue as before.

Right, but then if the update hook depends on a piece of state and also updates that state, then you end up updating every time you update.

jiekang:
Some notes for issues that I can think can be addressed in a future PR:

* At the moment, switching back and forth between the two event tabs has default initial state each time. I think from a user experience perspective, the state should be kept.
* When thinking about the create recording work and a custom configuration, it's natural for me to look at the event types list and think "it would be nice to have some kind of UX integration where I can use the list to help me build the recording configuration"

andrewazores:
> When thinking about the create recording work and a custom configuration, it's natural for me to look at the event types list and think "it would be nice to have some kind of UX integration where I can use the list to help me build the recording configuration"

I have that planned for next week. First will be linking from the Event Templates view to the Create Recording view with the template pre-selected. Second will be improving the Event Types view to make the expandable options somehow editable, and a workflow for pre-filling those also into the Create Recording custom definition string.

> At the moment, switching back and forth between the two event tabs has default initial state each time. I think from a user experience perspective, the state should be kept.

That's probably a good call. I'll need to think about how best to do that. Since they are both independent components that are rendered within a tab system, the components are unmounted (and destroyed?) when not in view.

Issue #58, opened 2020-04-24
Create recordings


andrewazores:
Users should be able to create recordings with variable durations, user-provided names, and user-specified event types/templates. Snapshot recordings should also be supported.

https://invis.io/T8WWKRMDPN5

Issue #57, opened 2020-04-24
List recordings


andrewazores:
Users should be able to view a list of recordings including details such as name, start time, duration, and state. Users should also be able to archive active recordings. The list view should be able to display either active or archived recordings, or possibly both.

https://invis.io/3MWWKS9DYAP

PR #59, opened 2020-04-24
"Create Recording" view

andrewazores:
This PR adds an initial pass at #58 and an associated minor enhancement to #57 .

A new "Create Recording" view is added, which is routed as a child of the Recordings List view. The Recordings List gains a
"Create" toolbar button which links to the new view.

The new view is just a form with a bit of logic to determine what command name and arguments to use for the recording
request sent to ContainerJFR. When the request is successfully submitted, the view routes back to the Recordings List.
If the user cancels out of the view then they are returned to their previous location - currently this can only be the
Recordings List, but other links are possible in the future.

Not yet handled:
~~1. input validation~~
~~2. any system for notifying users of what went wrong if the request fails (invalid input, duplicate recording name, etc.)~~
~~3. the app Nav bar will not indicate any location when the user is in the Create Recording view. It should probably highlight the parent path as active~~
~~4. breadcrumbs at the top of the view~~

image
image
image

image

### DIFF
--- a/src/app/CreateRecording/CreateRecording.tsx
+++ b/src/app/CreateRecording/CreateRecording.tsx
@@ -10,6 +10,12 @@ export interface CreateRecordingProps {
   eventSpecifiers?: string[];
 }
 
+export interface EventTemplate {
+  name: string;
+  description: string;
+  provider: string;
+}
+
 export const RecordingNamePattern = /^[\w_]+$/;
 export const TemplatePattern = /^template=([\w]+)$/;
 export const EventSpecifierPattern = /([\w\\.\$]+):([\w]+)=([\w\\d\.]+)/;
@@ -23,7 +29,7 @@ export const CreateRecording = (props: CreateRecordingProps) => {
   const [continuous, setContinuous] = React.useState(false);
   const [duration, setDuration] = React.useState(30);
   const [durationUnit, setDurationUnit] = React.useState(1);
-  const [templates, setTemplates] = React.useState([]);
+  const [templates, setTemplates] = React.useState([] as EventTemplate[]);
   const [template, setTemplate] = React.useState('');
   const [eventSpecifiers, setEventSpecifiers] = React.useState('');
   const [eventsValid, setEventsValid] = React.useState(false);
@@ -52,9 +58,9 @@ export const CreateRecording = (props: CreateRecordingProps) => {
     setEventSpecifiers(evt);
   };
 
-  const getEventSpecifiers = () => !!template ? template : eventSpecifiers;
+  const getEventSpecifiers = () => !!template ? `template=${template}` : eventSpecifiers;
 
-  const getEventString = () => !!template ? template : eventSpecifiers.split(/\s+/).filter(Boolean).join(',');
+  const getEventString = () => !!template ? `template=${template}` : eventSpecifiers.split(/\s+/).filter(Boolean).join(',');
 
   const handleRecordingNameChange = (name) => {
     setNameValid(RecordingNamePattern.test(name));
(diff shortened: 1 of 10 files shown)


jiekang:
@andrewazores I noticed you ended up addressing 1,3 and 4. Are you planning to address 2 here too?

andrewazores:
Just edited the initial text to cross off that issue as well actually. It was addressed by the last three commits in the series.

jiekang:
Oh nice! My bad oops.

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -16,26 +17,26 @@ export class CommandChannel {
...
-        (err: any) => console.error('Client URL configuration', err) // TODO add toast notification
+        (err: any) => this.logError('Client URL configuration', err)
       );
 
     fromFetch(`${this.apiSvc.authority}/grafana_datasource_url`)
       .pipe(concatMap(resp => from(resp.json())))
       .subscribe(
         (url: any) => this.grafanaDatasourceUrlSubject.next(url.grafanaDatasourceUrl),
-        (err: any) => console.error('Grafana Datasource configuration', err) // TODO add toast notification
+        (err: any) => this.logError('Grafana Datasource configuration', err)
```
jiekang, line 34: In my test I see the error as just: 

Danger alert:Grafana Datasource configuration
{}

Is the empty brackets expected, or just to do with how I'm running it through cjfr?

src/app/CreateRecording/CreateRecording.tsx
```diff
@@ -0,0 +1,262 @@
...
+    context.commandChannel.onResponse(command).pipe(
+      filter(m => m.id === id),
+      first(),
+      )
+      .subscribe((msg) => {
+          if (msg.status === 0) {
+            notifications.success('Recording created');
+            history.push('/recordings');
+          } else {
+            notifications.danger(`Request failed (Status ${msg.status})`, msg.payload)
```
jiekang, line 101: The payload looks like:

Danger alert:Request failed (Status -2)
FlightRecorderException: com.redhat.rhjmc.containerjfr.core.templates.TemplateService$UnknownEventTemplateException: Unknown event template "def"

Can it show a more user-friendly message? I'd recommend "Create recording failed: Unknown event template "def".  Just dropping the stack trace would be fine for a start.

src/app/Shared/Services/CommandChannel.service.tsx
andrewazores: That's the normal result right now. The response has a 500 status code but no payload, so that's what you see in the toast notification currently (unfortunately).

src/app/CreateRecording/CreateRecording.tsx
andrewazores: I could do something like that, although what you see there is just it dumping what it receives from ContainerJFR into a string. There would need to be some basic parsing or at least regex match to get it to process the message and drop the stack trace, or else ContainerJFR would need to be modified to send a different response.

src/app/Shared/Services/CommandChannel.service.tsx
jiekang: Okay, maybe we can open an issue to target the error notifications in general, seems some need cross-project work. Then we can resolve them later. What do you think?

andrewazores: Sounds good to me.

jiekang: Okay, let's do that. Can you resolve this after an issue is created?

andrewazores: #65

src/app/CreateRecording/CreateRecording.tsx
```diff
@@ -0,0 +1,262 @@
...
+                      }
+                    </FormSelectOptionGroup>
+                  </FormSelect>
+                </SplitItem>
+                <SplitItem>
+                  <Tooltip
+                    position={TooltipPosition.auto}
+                    content={
+                      <Text component={TextVariants.p}>
+                        Templates are selected with the syntax <i>template=Foo</i>. Event names and options are specified with the syntax
```
jiekang, line 238: For whatever reason, when I view the tooltip I see
"with the syntax*namespace.EventName...*"

I guess you need an explicit space here?

andrewazores: I reformatted the tooltip. It should be easier to read now and that odd lack of whitespace probably won't happen - although I'm not sure why the .tsx template would have interpreted the text that way anyway.

jiekang: Nice!

PR #63, opened 2020-04-30
Recording List overhaul as Data List

andrewazores:
This is based on top of #59

Mockup: https://projects.invisionapp.com/d/main#/console/19772434/413969810/preview

This PR reimplements the Recording List as a Patternfly Data List, rather than as a Table. This allows more flexibility in the rendering of the cells - for example, the download and report URLs are now clickable links.

This also opens the door to more advanced custom styling options, such as making the rows expandable and containing more custom rendered elements (as opposed to an expandable table, which can only contain plain text). This could include copying the ContainerJFR Web Angular UI where the expanded rows contain an inline recording report. The Data List also allows for action buttons and other customized elements to be placed within the cells, which will likely be used for (re)implementing the "View in Grafana" feature.

In addition, checkboxes are added to each row of the table, as well as a header checkbox which acts as a select-all. These checkboxes are used in conjunction with the new Stop and Delete buttons, allowing these actions to be performed on recordings or batches of recordings.

image
image

### DIFF
--- a/src/app/RecordingList/RecordingList.tsx
+++ b/src/app/RecordingList/RecordingList.tsx
@@ -1,8 +1,7 @@
 import * as React from 'react';
 import { useHistory, useRouteMatch } from 'react-router-dom';
 import { filter, map } from 'rxjs/operators';
-import { Button, Card, CardBody, CardHeader, PageSection, Text, TextVariants, Title } from '@patternfly/react-core';
-import { Table, TableHeader, TableBody, textCenter } from '@patternfly/react-table';
+import { Button, Card, CardBody, CardHeader, DataList, DataListItem, DataListItemRow, DataListItemCells, DataListCell, PageSection, Text, TextVariants, Title } from '@patternfly/react-core';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { TargetView } from '@app/TargetView/TargetView';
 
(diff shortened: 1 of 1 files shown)


jiekang:
Just a quick thought on the UX. Maybe for Download/Report, we want to replace the column with Actions, be at the right-most and then have two clickable actions with descriptive text "Download Recording", "View Analysis Report", or something.

andrewazores:
Yea, there are a lot of options. I like what we have in the Angular client, the expandable rows which opened up to display the automated analysis reports inline, but adding an Actions column which contains actions for downloading, viewing the report (in a new tab maybe), and uploading to Grafana would make sense and also be similar to what we have in Angular.

jiekang:
When I tested https://github.com/rh-jmc-team/container-jfr-web/pull/59 and created a recording "hi" and downloaded it, it gave me the file "hi". I think the filename should be extended with .jfr. Entering hi.jfr initially is an invalid name so we should be fin to do that. We can do that in this PR maybe?

andrewazores:
Rebased.

src/app/RecordingList/RecordingList.tsx
```diff
@@ -74,16 +104,98 @@ export const RecordingList = (props) => {
...
+    return (
+      <Toolbar>
+        <ToolbarGroup>
+          <ToolbarItem>
+            <Button variant="primary" onClick={handleCreateRecording}>Create</Button>
+          </ToolbarItem>
+        </ToolbarGroup>
+        <ToolbarGroup>
+          <ToolbarItem>
+            <Button variant="secondary" onClick={handleStopRecordings} isDisabled={!checkedIndices.length}>Stop</Button>
```
jiekang, line 162: Can `isDisabled` be expanded to also check if any of the selected recordings are in the RunningState?

andrewazores: Sure. What should the behaviour be if multiple recordings are selected and are in different states? I think it makes sense to enable the Stop button if at least one recording is currently running, and then I suppose to only send a stop message for those such recordings.

jiekang: Yeah, that sounds right to me!

PR #69, opened 2020-05-01
Add "Archived Recordings" view tab

andrewazores:
Adds a conditionally-displayed Archived Recordings tab to the Recordings List card component. If the ContainerJFR instance does not support recording archiving then the UI is unchanged - no tabs are displayed on the card, only the Active Recordings table. If Archives are supported then the active recording list gains an Archive button, and the new view allows a list of archived recordings to be seen, downloaded, and deleted, as well as reports viewed by link URL.

image
image
image

### DIFF
--- a/src/app/RecordingList/ActiveRecordingsList.tsx
+++ b/src/app/RecordingList/ActiveRecordingsList.tsx
@@ -1,0 +1,181 @@
+import * as React from 'react';
+import { useHistory, useRouteMatch } from 'react-router-dom';
+import { filter, map } from 'rxjs/operators';
+import { Button, DataList, DataListCheck, DataListItem, DataListItemRow, DataListItemCells, DataListCell, Text, TextVariants, Title, Toolbar, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
+import { ServiceContext } from '@app/Shared/Services/Services';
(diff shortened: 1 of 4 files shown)


andrewazores:
You mean documentation about what "archiving" means?

There's `COMMANDS.md` which explains `save` and related commands: https://github.com/rh-jmc-team/container-jfr/blob/master/COMMANDS.md

Otherwise, not really. Is there anything specific you would like to see documented?

jiekang:
@andrewazores How to deploy with/without archiving enabled; I guess that's for cjfr side of things.

andrewazores:
Yea, I could make a note in the ContainerJFR README about that. All that's required is that there is a directory `/flightrecordings` accessible to the ContainerJFR process. In the main -operator use-case that's automatically configured by one of the operator controllers - it sets up a PersistentVolumeClaim and mounts it to that location in the container.

jiekang:
I think that information is valuable enough to note in the readme!

andrewazores:
That `/flightrecordings` path could also be made configurable if we get any users that care to use ContainerJFR as a local Java process, although right now it's a hardcoded path. IMO that's fine for right now when the expected use-cases are primarily just the -operator and otherwise running the same OCI container image elsewhere, in which case a volume mount to that path is also nothing unusual.

PR #73, opened 2020-05-07
Expandable recording rows for inline reports

andrewazores:
This PR makes the Active and Archived recording list rows expandable. When expanded, an inline recording reports view is displayed, showing the HTML rules report contents generated by ContainerJFR (using upstream JMC automated analysis).

Additionally, the Download action is moved from a simple link in the recording columns to an action overflow menu on each row. Clicking the "Download" button within this overflow menu will download the recording to the user's local machine. (Note: if testing using a CORS setup, an active recording's filename will be saved without the `.jfr` extension. This is a CORS limitation - the client cannot change the name of the downloaded file from that in the request path. When deployed normally, ie the client is served by ContainerJFR and CORS is disabled/not needed, then the `.jfr` extension will be properly appended).

There is also a "View Report ..." button in the same overflow menu, which opens the same rules report for the recording but in a new browser tab or window.

Screenshot_2020-05-07 Recordings
Screenshot_2020-05-07 Recordings(1)


andrewazores:
Just realized that both the `iframe` and new tab link for viewing reports do not know to pass credentials to ContainerJFR, so if any authz are enabled then reports are not viewable. Looking into how to solve this.

jiekang:
CORS, jeeze...

andrewazores:
The Authorization header problem mentioned above has been solved, and I also noticed a bug where hidden reports were still being updated periodically in the background, which I also fixed.

andrewazores:
I think this is really done now. Other than the sort of lazy-loading I mentioned above (keeping the `ReportFrame` out of the DOM until the row is expanded, to avoid "background" report requests), things are still the same as initially set out. Just now authz and CORS are properly supported.

jiekang:
I played around with the expanded reports UI and noticed some behavior:

* For multiple archived recordings. Expanding item one. On completion of report loading, expand item two. Both item one and two go through loading again.
* For recordings, expand item one. Every so often, after alt-tabbing around to other screens, the report loading for the item seems to trigger again. Also applies when multiple recordings are loaded.
* For recordings, expand item one (stopped recording). On load completion, switch between archived or active recordings. Items go through loading state again. I think maybe in the case of a continuous recording, I'd expect it to reload? Otherwise it doesn't feel right from user perspective when these aren't two different pages, but tabs on the same page.
* For multiple archived recordings. Expand and load one item. See time taken. Expand and load ten items at once. Time taken is dramatically high.

jiekang:
Also:
* Open a report. Click on the ellipsis. Report reloads. Exit ellipsis form. Report reloads.

andrewazores:
Points 1, 3, and 5 are all the same thing I think - React is seeing a change in the view state (expanding/collapsing rows, changing tabs, etc.) and because of this, it's triggering a re-render of the view. It should be something that can be worked around, but I'd need to look into it in more detail. This probably goes to a larger architectural issue we should consider around the whole front-end client state and exploring options like Redux.

Point 2 is somewhat related as well. There is a background timer that refreshes the list periodically, and when it re-populates the list, the reports are also refreshed due to the re-render (again, because React sees that the list contents "changed" - even if everything still has equal values, if I understand what's happening correctly).

For point 4, could you elaborate? How large are your recordings? Generating a report can be a fairly intensive operation on the ContainerJFR backend, so if you're trying to do ten of them at the same time, it isn't surprising that it might be a bit slow. This sounds more like a backend performance issue than a frontend client issue?

jiekang:
Bleh I'll use numbered points next time.

Uh... for point 4 I think as you say it's a consequence of the report generation time and it's probably just exacerbated from the UI due to the underlying issue from the other points.

For me, the UX experience is so bizarre I would strongly prefer resolving underlying render/refresh issues before adding this. I'm also surprised a render/refreshing takes so long atm, e.g. for opening/closing a row; the input is mostly HTML right? Don't suppose that's cached in the "back-end" of the front-end?

What do you think?

andrewazores:
Yes, it's just rendering additional HTML to an `iframe` in the page, so that part shouldn't be very expensive. But there is no caching of that HTML report at all, so every time the `iframe` is loaded, it triggers a brand new report load and generation in the backend. That's where all the expense is. Some backend caching could be done, but that would only really work for recordings that are already stopped or archived - an ongoing recording, which is our primary targeted use case (in particular for these automated reports), obviously can't have its reports served from a cache.

I'm okay with holding off on merging this PR until some work is done (#79) to figure out better client-side state management, but that means that React can't reach feature-parity with Angular until then. I would really rather have the React client merged sooner rather than later, even if it is still a bit rough around the edges, or perhaps missing a minor feature here and there. We're already at a point where the `master` branch Angular client is incompatible with the latest ContainerJFR backend, and it's only going to get worse in the coming weeks with the connectionless API changes etc. There performance issues are also present in the Angular client anyway - although it does happen to avoid spurious refreshes/re-renders, but mostly unintentionally.

jiekang:
Ah right... the Angular stuff... 

I think you can view the reports in a new tab without this PR right? I think the react stuff can become mainline anytime now but this refresh and re-render of stuff is just a bit too janky to me at the moment. 

If I recall in OSIO, you had to be careful with parent child things and causing render changes to ripple up/down the tree of components unnecessarily? That was Angular though right? I need to spend more time in with React ugh, sorry.

Redux seems like a potential help for our state. I don't have enough experience in React to give an opinion at the moment. I'll get back to you...

andrewazores:
Yes, without this PR you still have the ability to download recordings and reports, and you can choose to open the report HTML doc in your same browser instance rather than saving it somewhere (well, it probably still goes into /tmp or something). 

Here's what the current plain `react` branch looks like in this view:
image

OSIO was Angular so there are some differences there in how change detection works for sure. I think there are also some differences between React's functional components and class-based components as far as which hooks are available etc, so maybe we can get away with rewriting some pieces of this PR using classes instead and avoid some of these rerendering/lifecycle issues, without having to go all the way to using a separate state management library like Redux?

Anyway - from my side I'm happy to merge the `react` branch into `master` as-is, and this PR can be held for later. The overflow menu item part can probably be pulled out into #75 and that can be merged separately as well, so the only missing feature would be the in-line report frames which this PR would provide eventually.

jiekang:
That plan of action sounds good to me. Over lunch I realize I expanded a bit too far ahead of things here.

Specifically for this expandable row work I'd like to see, potentially addressed separately if that makes the most sense:
1. Opening/closing row A does not cause all other open rows B,C,... to reload.
2. When the list is periodically refreshed, open rows that don't have changes are not reloaded

andrewazores:
Sounds good to me. Those are some good metrics we can use to judge whether any state-management patches applied are effective and useful if we apply those patches on top of this PR's patches. I think that'll be a good path forward for this feature.

PR #82, opened 2020-05-20
Connectionless API

andrewazores:
Rebased and reopened version of #80, after `react` branch merge to `master`.

The goal here is simply to patch the React client to work with ContainerJFR's upcoming non-connection-oriented/connectionless API changes ( https://github.com/rh-jmc-team/container-jfr/pull/161 ). It is intentionally minimal and provides support for the backend changes only, not changing any client layout or behaviour.

The `CommandChannel` front-end "service" object gains an RxJS Subject which maintains a piece of state representing the "connected" target JVM, emulating the previous stateful connection-oriented API. This is just to maintain compatibility with the rest of the structure of the React client and will be removed at a later time, so that the web application itself also works in a "connectionless" manner.

Otherwise, this simply adds a frontend distinction between messages sent being "control" messages or not. "Control" messages simply being those that do not specify any `targetId` - they perform some kind of "control" operation on the ContainerJFR backend instance itself (ex. delete a saved recording in ContainerJFR disk space, ping the instance). Other, non-control messages do specify a `targetId`, and perform operations upon a specific JVM target (ex. start/stop recording, delete recording, list recordings - all occur on a specific JVM). This distinction will also likely be refactored or removed in the future when the various components are modified to work in a connectionless manner.

### DIFF
--- a/src/app/Shared/Services/CommandChannel.service.tsx
+++ b/src/app/Shared/Services/CommandChannel.service.tsx
@@ -1,6 +1,6 @@
 import { from, Subject, BehaviorSubject, Observable, ReplaySubject, combineLatest } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
-import { concatMap, distinctUntilChanged, filter, first, map, tap } from 'rxjs/operators';
+import { concatMap, filter, first, map, tap } from 'rxjs/operators';
 import { ApiService } from './Api.service';
 import { Notifications } from '@app/Notifications/Notifications';
 import { nanoid } from 'nanoid';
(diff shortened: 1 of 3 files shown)


src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -34,25 +35,22 @@ export class CommandChannel {
         (err: any) => this.logError('Grafana Datasource configuration', err)
       );
 
-    this.onResponse('disconnect').pipe(
-      filter(msg => msg.status !== 0),
-    ).subscribe(msg => this.logError('Command failure', msg));
+    fromFetch(`${this.apiSvc.authority}/api/v1/grafana_dashboard_url`)
```
jiekang, line 38: When I view file, it looks like this fromFetch is duplicated from above; likely not intended?

andrewazores: One is dashboard and one is datasource ;-)

jiekang: Oh, you're right. My mistake, sorry.

PR #83, opened 2020-05-20
View in Grafana

andrewazores:
Rebased and reopened version of #75 after `react` branch was merged into `master`

~~Based on top of #84~~

### DIFF
--- a/src/app/RecordingList/ActiveRecordingsList.tsx
+++ b/src/app/RecordingList/ActiveRecordingsList.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
+import { NotificationsContext } from '@app/Notifications/Notifications';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { Button, DataListAction, DataListCell, DataListCheck, DataListItemCells, DataListItemRow, DataToolbar, DataToolbarContent, DataToolbarItem, Dropdown, DropdownItem, DropdownPosition, KebabToggle, Text } from '@patternfly/react-core';
 import { useHistory, useRouteMatch } from 'react-router-dom';
-import { filter, map } from 'rxjs/operators';
+import { filter, first, map } from 'rxjs/operators';
 import { Recording, RecordingState } from './RecordingList';
 import { RecordingsDataTable } from './RecordingsDataTable';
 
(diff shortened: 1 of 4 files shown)


andrewazores:
This has been rebased onto `master` rather than #84. The reports and recordings download links and removed and these actions moved into a dropdown action menu on each row of the active and archived recordings lists, and a "View in Grafana..." action button has also been added to this menu.

andrewazores:
Rebased, had some conflicts after merging #85 but generally just imports being cleaned up.

jiekang:
@andrewazores is it intentional that report is only downloadable and not viewable in another tab anymore?

andrewazores:
It seemed redundant to offer two different ways to do it, so I suppose it's intentional. You can click "Download Report" and then when your browser download prompt appears simply choose to open the document in the same browser again, which will probably have the effect of opening it in a new tab.

src/app/RecordingList/ActiveRecordingsList.tsx
```diff
@@ -192,3 +183,97 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
...
+}
+
+export const RecordingActions: React.FunctionComponent<RecordingActionsProps> = (props) => {
+  const context = React.useContext(ServiceContext);
+  const notifications = React.useContext(NotificationsContext);
+  const [grafanaEnabled, setGrafanaEnabled] = React.useState(false);
+  const [uploadIds, setUploadIds] = React.useState([] as string[]);
+
+  React.useEffect(() => {
+    const sub = context.commandChannel.grafanaDatasourceUrl()
```
jiekang, line 201: I think this would be on the container-jfr side, but the jfr-datasource supports liveness on it's root url. Maybe a quick http test can be used so if someone does supply a url but it doesn't actually work, we still have it disabled? Does that make sense to you? I can file an issue and maybe look at it too for once

andrewazores: Yea, makes sense to me. I've been meaning to implement some kind of `/health` for container-jfr itself for a while, so maybe what we can do is have that return an HTTP status to indicate its own health, and the response body can be some JSON that includes details on whether the datasaource or dashboard URL etc. are configured and are reachable from ContainerJFR as well.

PR #90, opened 2020-06-11
Include target connection port number, if any

### LINKED ISSUE #89: Since connectionless API changes, target port numbers are not included in command args
The CommandChannel/TargetSelect interaction in #82 is somewhat broken. The target connection port is not included in the connection identifier state set in the CommandChannel, so commands causing connections to targets using non-default ports will always fail.

andrewazores:
Fixes #89

### DIFF
--- a/src/app/TargetSelect/TargetSelect.tsx
+++ b/src/app/TargetSelect/TargetSelect.tsx
@@ -42,7 +42,11 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props) 
     if (isPlaceholder) {
       context.commandChannel.setTarget('');
     } else {
-      context.commandChannel.setTarget(selection.connectUrl);
+      let identifier = selection.connectUrl;
+      if (selection.port) {
+        identifier += `:${selection.port}`;
+      }
+      context.commandChannel.setTarget(identifier);
     }
     // FIXME setting the expanded state to false seems to cause an "unmounted component" error
     // in the browser console


vic-ma:
My understanding of the issue this fixes is that if cjfr connects to non-default port (anything other than 9091?), and you try to run, for example, a record command in the web client, then that command will fail, because the command always runs on the default port of the specified connection. Is this about right?

I am having trouble figuring out how to test this. With the `run.sh` script, the only mention of `9091` that I see is in the podman command. But even if I change that to use `-p 9098:9098` instead, when I do a recording, I can see in the recording urls and cjfr output that it is still running on port `9091`. I see some mentions of `9091` in `pom.xml` but if I were to change it there I imagine it would just change the default port number?

I also tried testing this with cjfr as a local process, but when I change the target JVM to use a port other than `9091` (it works on `9091`), cjfr can't find it anymore. I see in the README that it says for non-Kube deployments it expects targets to be on port `9091`, so I suppose this is expected. Does this just mean that testing for this issue has to be done in a container environment?

andrewazores:
Yes, that's right. A `dump`/`start` command, or `list-event-templates`, etc. in the web-client will currently always fail for targets listening on non-default port numbers.

The default port number of 9091 actually comes from `-core`: https://github.com/rh-jmc-team/container-jfr-core/blob/master/src/main/java/com/redhat/rhjmc/containerjfr/core/net/JFRConnectionToolkit.java

There are a bunch of steps that everything goes through before getting to that point however, but the gist of it is that if you use just `foo` as the target ID for connecting, then eventually the assumed port number is 9091. If that doesn't match the actual JMX port that the target is listening on then the connection will fail, obviously.

I'm not sure how you set things up to test it with the `run.sh` command, but there would be a few different things that need to be altered. First, the JVM flags set for the target application would need to be set to listen to JMX on a different port number than 9091 (ex. `-Dcom.sun.management.jmxremote.port=9093`). Then, podman or whatever container platform is running the target would also need to expose that port for that container so that it can accept connections. If you only change the port exposed by podman, the JVM process doesn't know about that and is still trying to listen for JMX connections on the port it was configured to, even if that port isn't exposed by podman.

It should be possible to test this as a local process by setting the `com.sun.management.jmxremote.port` JVM flag to some other number than 9091 also setting `-Dcom.sun.management.jmxremote.autodiscovery=true`. Send a `scan-targets` command (may need to be done twice, since JDP discovery is async and may take a second to find itself after it's first started). Sending a `{command:list,args:[localhost]}` should then fail with an exception about being unable to open the connection, but `{command:list,args:["localhost:9093"]}` should work.

For reference, here is the command line I just used for testing this as a local process:

```
MAVEN_OPTS="-Dcom.sun.management.jmxremote.port=9093 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.autodiscovery=true" mvn -DskipTests=true -Dcontainerjfr.minimal=true clean spotless:apply prepare-package exec:java
```

It can also be tested with the Operator setup by deploying both of the sample apps. One of them listens on 9091 and is detected that way, the other listens on 9093 and is detected by named port detection, but fails because the port number is not included in the target ID (the subject of this PR).

vic-ma:
I ran cjfr as a local process using all the jmx flags you used, and after calling `scan-targets`, I get this:
```
{"connectUrl":"service:jmx:rmi:///jndi/rmi://victor-work:9093/jmxrmi","alias":"org.codehaus.plexus.classworlds.launcher.Launcher","port":0}]}
```

Then, I am able to select this new target in the web client, and when I start a recording I can see that this is the command that gets run:
```
[INFO] (192.168.0.107:45520): CMD {"id":"1YulumKRKXKeGU9a_BqcN","command":"dump","args":["service:jmx:rmi:///jndi/rmi://victor-work:9093/jmxrmi","foo","30","template=Continuous"]}
```
And the command doesn't give an error.

I can see that this is because the `targetId` is in the form of a JMX URL rather than `host:port`. I also had cjfr connect to a separate `HelloWorld` JVM, and it also uses the JMX URL. Sending commands in `host:port` form like `{command:list,args:["localhost:9093"]}` work.

So I'm not sure why my cjfr is using JMX URLs or how to change this in order to reproduce the error.

andrewazores:
Ah, you're getting JMX URLs because it's local discovery using JDP. JMX URLs are generally the preferred connection method when available, and they're readily available with JDP but not necessarily with other discovery methods. So the web client sidesteps the bug when you're using it with a local ContainerJFR process.

The way I stumbled across this bug to begin with is the other example I gave before - using the Operator and deploying both of the sample apps. That's probably the most straightforward way to reproduce the bug, since there is no particular setup required other than starting up crc and then running `make deploy sample_app sample_app2`. It does take a bit more legwork to verify the fix in this setup, however.

vic-ma:
I'm trying to run the Operator in a CRC.

After I run the `make image` command, I can see that the image appears in `podman images`:
```
[vma@victor-work ~]$ podman images
REPOSITORY                                    TAG      IMAGE ID       CREATED          SIZE
quay.io/some-user/container-jfr-operator      latest   4692ffd6f3dd   42 seconds ago   126 MB
quay.io/rh-jmc-team/container-jfr             0.19.0   c0f9f3dded8a   5 days ago       614 MB
...
```

But when I then run the podman image push command I get this:
```
[vma@victor-work ~/container-jfr-operator]$ podman image push $IMAGE_TAG
Getting image source signatures
Error: Error copying image to the remote destination: Error trying to reuse blob \
sha256:e0a4651703ad7a4a7a8772a8242e22c605f0474579299d9142bc58e846dee2d2          \
at destination: Error checking whether a blob                                    \
sha256:e0a4651703ad7a4a7a8772a8242e22c605f0474579299d9142bc58e846dee2d2          \
exists in quay.io/some-user/container-jfr-operator: unauthorized:                \
authentication required
```
If I run it as superuser, it can't find the image:
```
[vma@victor-work ~/container-jfr-operator]$ sudo podman image push quay.io/some-user/container-jfr-operator
Swipe your finger across the fingerprint reader
Error: unable to find 'quay.io/some-user/container-jfr-operator' in local storage: no such image
```

Is this error maybe just because I've already pushed the image, and it sees that I'm trying to push it again? I can't remember if it gave me an error the first time I ran the command.

When I run `make deploy` afterwards, the pod does not start successfully:
```
[vma@victor-work ~]$ oc get pods
NAME                                      READY     STATUS             RESTARTS   AGE
container-jfr-operator-648df49667-dzcvs   0/1       ImagePullBackOff   0          3m38s
```
```
Events:
  Type     Reason     Age              From                         Message
  ----     ------     ----             ----                         -------
  Normal   Scheduled  3m               default-scheduler            Successfully assigned default/container-jfr-operator-648df49667-dzcvs to crc-m27h4-master-0
  Normal   Pulling    1m (x4 over 2m)  kubelet, crc-m27h4-master-0  Pulling image "quay.io/some-user/container-jfr-operator"
  Warning  Failed     1m (x4 over 2m)  kubelet, crc-m27h4-master-0  Failed to pull image "quay.io/some-user/container-jfr-operator": rpc error: code = Unknown desc = Error reading manifest latest in quay.io/some-user/container-jfr-operator: unauthorized: access to the requested resource is not authorized
  Warning  Failed     1m (x4 over 2m)  kubelet, crc-m27h4-master-0  Error: ErrImagePull
  Warning  Failed     1m (x6 over 2m)  kubelet, crc-m27h4-master-0  Error: ImagePullBackOff
  Normal   BackOff    1m (x7 over 2m)  kubelet, crc-m27h4-master-0  Back-off pulling image "quay.io/some-user/container-jfr-operator"
```
---
After writing the above, I realized that I was probably supposed to change the `some-user` in `$IMAGE_TAG` to my username, but even after doing this I got the same error.

andrewazores:
Have you done `podman login quay.io`? Podman needs to be authenticated to Quay in order to push images to its repository. When you do `sudo podman push`, Podman is looking to push an image from the root user's local repository to Quay, so even if you're logged in that won't work/isn't the right thing.

vic-ma:
I'm having trouble with pushing the image to Quay. I tried once yesterday, and then deleted everything and tried again today, but in both cases, when I first run the command, the upload bars progress for some time, and then it fails with the same error message:
```
Error: Error copying image to the remote destination: Error writing blob: \
Put "https://quay.io/v2/vic_ma/container-jfr-operator/blobs/uploads/95e4de75-1a0e-4e7b-a56a-a830a02a6d29?digest=sha256%3A60bb516266585c5efce9502cafb76e629e54a4efbaa59ca5d4193cc0e5b6f2cc": \
net/http: TLS handshake timeout
```
I've worked out that this is probably just because I have a slow internet connection so it was not able to complete the upload before the handshake expired or something.

If I run the command again afterwards, it'll continue the upload and eventually complete without error, but on quay.io it says the image size is 50.0 MB even though it should be around 100 MB. And the pod doesn't start successfully if I try to `make deploy`.

So I'm not sure if the upload actually resumed and completed successfully and it's just something else wrong with my setup, or if after that error the upload was just broken.

andrewazores:
Huh. You can try setting up your Podman to directly access the CRC registry then, there are some hints how to do this here:

https://github.com/code-ready/crc/issues/147

I've never done it myself, so I'll need to try it out too.

vic-ma:
Hmm I'm having trouble figuring out the internal registry stuff. Reading through that PR and the openshift docs, I don't think I know enough about openshift to be able to follow the instructions adequately. I can take the time to learn it if you think it's worth doing, but since none of this will be an issue once I have access to a decent internet connection, I'm inclined to think that it isn't necessary.

Maybe @Alexjsenn could take a look at this PR instead?

andrewazores:
Sure, sounds fine to me. I don't think it's particularly worth spending time to figure out right now - if it really becomes a useful piece for us to figure out then either Elliott or I will probably take a look into it, since we're more familiar with it.

PR #91, opened 2020-06-11
UI better conceptually conforms to connectionless API

andrewazores:
This removes the "dynamic routing" from the UI, making all views routable at all times. For the nested recording creation form subviews, they also gain a target selection widget so that they continue to make sense in context. This allows a user to visit any in-app URL without encountering the 404 page that was previously found, since no target would have been selected in the initial application state.


andrewazores:
I'm a little conflicted on the TargetSelect widget being visible in the recording creation forms. It makes sense that all views that have to do with a specific target should have that widget available, but on the other hand the recording creation forms are only reachable in-app when the user has already first selected a target and visited another view. But without the selection widget being there, the user can directly visit the recording creation forms by URL and end up in a weird situation where they are at the forms but the creation will always fail because there is no select target.

Also, I still need to implement a better indication for why the view is blank when the user visits the Recordings or Events views and does not have a target selected. This should be distinct from an "empty state" for the list views that should normally be populated into those views when a target is selected.

jiekang:
Code wise I think it looks okay. UX wise I guess this is our foray into how we want things to end up. I agree with your concerns. Maybe target connection is something that should be done in an off-page widget on the side where Dashboard, etc. are. or even added to the header.

andrewazores:
I think either of your suggestions sound like a good UX, so I'll take a look and see which one better fits with Patternfly's UX guidelines and see about implementing that in a later PR.

andrewazores:
Maybe the page header "context selector" is the right thing for this? https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/navigation-system

It seems like it's meant for primarily for changing the layout and components on the page, not necessarily just changing the information within the components. But it seems like it would probably work well and more or less fits the intent of the design.

Also, I rebased this PR, so it should be ready for merge once it's approved.

jiekang:
The context selector you linked looks like something we can try out.

PR #95, opened 2020-06-18
Remove timed background command channel ping

andrewazores:
The periodic ping was initially used for development testing to check
whether communication errors were due to closed connections or the
backend not responding, but this is no longer a common concern and so
the ping serves no practical purpose. It should be removed to prevent it
from clogging backend logs.

### DIFF
--- a/src/app/Shared/Services/CommandChannel.service.tsx
+++ b/src/app/Shared/Services/CommandChannel.service.tsx
@@ -17,7 +17,6 @@ export class CommandChannel {
   private readonly grafanaDatasourceUrlSubject = new ReplaySubject<string>(1);
   private readonly grafanaDashboardUrlSubject = new ReplaySubject<string>(1);
   private readonly targetSubject = new ReplaySubject<string>(1);
-  private pingTimer = -1;
 
   constructor(apiSvc: ApiService, private readonly notifications: Notifications) {
     this.apiSvc = apiSvc;
@@ -90,16 +89,12 @@ export class CommandChannel {
           protocol: subprotocol,
           openObserver: {
             next: () => {
-              this.pingTimer = window.setInterval(() => {
-                this.sendControlMessage('ping');
-              }, 10 * 1000);
               this.ready.next(true);
             }
           },
           closeObserver: {
             next: () => {
               this.ready.next(false);
-              window.clearInterval(this.pingTimer);
               this.notifications.info('WebSocket connection lost');
             }
           }


vic-ma:
Not directly related to this PR, but whenever I start my cjfr now, the webpage asks me for an authorization token and says that the cjfr server is configured with Bearer token authentication.

I've tried using multiple different commits in multiple different local repos (including a new one I cloned just now), and I've also tried using both the `run` script and `exec:java`, but the login page always pops up. The only thing I know I changed from before I had this issue is adding in the environment variables you mentioned on IRC, but even when I unset these, the login page is still there.

I'm really not sure what could be causing this. If I run `env | grep CONTAINER`, nothing pops up, so it isn't like I accidentally set `CONTAINER_JFR_AUTH_MANAGER` or anything. If I do manually set that variable to `com.redhat.rhjmc.containerjfr.net.BasicAuthManager` or `com.redhat.rhjmc.containerjfr.net.NoopAuthManager` or random gibberish, nothing changes.

If I type in random characters or no characters into the Token field and press Login, nothing happens. Also, strangely, the Grafana error message pops up twice now.

andrewazores:
Hmm. I hadn't noticed, but I see the same thing too if I try to use the web client built into the container-jfr image. Is that how you're running it? When I run a separate -web dev instance using `npm run start:dev` then that works as expected. I'll have to take a look into what's going on there.

> If I type in random characters or no characters into the Token field and press Login, nothing happens. Also, strangely, the Grafana error message pops up twice now.

Is it one for the dashboard and one for the datasource? That's currently normal if so.

andrewazores:
Screenshot_2020-06-19_14-46-42

That would do it. Now to see why the backend is doing that.

vic-ma:
>I see the same thing too if I try to use the web client built into the container-jfr image. Is that how you're running it?

I think so...I'm just building the `container-jfr` repo and then from there using the `run.sh` script or running `exec:java`. I'm definitely not doing `npm run start:dev` or anything.

andrewazores:
Yea, so that's using the client that is built as part of the container-jfr built and baked into the container image. For some reason that's breaking due to CORS rejections, but a separate Webpack devserver (`npm run start:dev` in the -web repo) is working normally. Weird.

andrewazores:
Oh, weird. Maybe I had some browser caching issue. I did a hard refresh of the container-jfr built-in client (local process, mvn exec style) and it's working as expected, with the container-jfr on current master branch and its web-client submodule on v0.10.0 - same as is in master.

andrewazores:
Hmm. Going into the web-client submodule and checking out the remove-ping branch from this PR, doing a clean build, and a hard refresh in the browser is also working as expected.

vic-ma:
I see the same thing in my auth request header as what you posted above. I tried rebuilding just now, but still get the issue. I also tried opening `localhost:8181` in firefox and chromium, which I haven't used before, but they both have the same issue.

>I did a hard refresh of the container-jfr built-in client (local process, mvn exec style) and it's working as expected

By hard refresh, do you just mean rebuilding, or is it something else?

andrewazores:
It depends on the browser, but in Firefox for example it's pressing Ctrl+Shift+R, which causes the browser to refresh the page and not use any cached assets - as opposed to a normal Ctrl+R or an F5, which refreshes but uses cached assets when available. The web client is often not cached properly, so even if the asset on the server has changed due to a rebuild, the browser can't determine this and reuses an old cached asset.

vic-ma:
Hmm it still asks me for a token, but now the only tab I see on the left is "Dashboard" whereas before there were two others, and this time only one Grafana error popped up. This is the case when I restart CJFR as well.

andrewazores:
Have you tried a different browser? @Alexjsenn had some weird issues that sound similar as well, but after switching browsers it started working normally again. I'm not sure what's going on there.

vic-ma:
Yes I've tried in multiple browsers and did Ctrl+Shift+R in all.

vic-ma:
Could it be an issue related to fonts at all? I downgraded my Pango recently, so maybe that? I see a few log messages which I don't think were there before, like:
```
[INFO] (127.0.0.1:42274): GET /fonts/overpass-regular.woff2 403 0ms
```

andrewazores:
I don't think so, the fonts there should all be coming from the web-client build and then served out by the container-jfr webserver. They aren't coming from your system. It's weird that you're getting a 403 on those requests even, though. Does it also say it was from CORS rejection?

vic-ma:
No, when I open/refresh the webpage, it just prints these logs:
```[INFO] (127.0.0.1:44358): GET /api/v1/grafana_datasource_url 500 0ms
[ERROR] io.vertx.ext.web.handler.impl.HttpStatusException: Internal Server Error

[INFO] (127.0.0.1:44362): GET /api/v1/grafana_dashboard_url 500 2ms
[INFO] (127.0.0.1:44362): POST /api/v1/auth 403 0ms
[INFO] (127.0.0.1:44362): GET /fonts/overpass-regular.woff2 403 0ms
[INFO] (127.0.0.1:44358): GET /fonts/overpass-semibold.woff2 403 0ms
[INFO] (127.0.0.1:44356): GET /fonts/overpass-light.woff2 403 0ms
[INFO] (127.0.0.1:44356): GET /fonts/overpass-regular.woff 403 0ms
[INFO] (127.0.0.1:44358): GET /fonts/overpass-semibold.woff 403 0ms
[INFO] (127.0.0.1:44356): GET /fonts/overpass-light.woff 403 2ms
[INFO] (127.0.0.1:44356): GET /fonts/overpass-regular.ttf 403 0ms
[INFO] (127.0.0.1:44358): GET /fonts/overpass-semibold.ttf 403 0ms
[INFO] (127.0.0.1:44358): GET /fonts/overpass-light.ttf 403 0ms
```

vic-ma:
Actually, maybe. In chrome, I see this in the console:

```
localhost/:1 GET http://localhost:8181/fonts/overpass-regular.woff2 net::ERR_ABORTED 403 (CORS Rejected - Invalid origin)
```

andrewazores:
I meant in your browser's devtools console, does it mention the CORS rejection? The backend logs show just the 403 status code, but not the reason string for the 403.

I've just pushed an image to Quay tagged as `quay.io/andrewazores/container-jfr:0.19.0-SNAPSHOT`. Can you try running that and see how it behaves for you? It's working as expected for me.

vic-ma:
Here are the logs from Chrome from my local CJFR; also I upgraded Pango and rebooted but it didn't seem to make a difference:

```
:8181/api/v1/auth:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
2webpack-internal:///190:4593 SyntaxError: Unexpected token D in JSON at position 0
CommandChannel.logError @ webpack-internal:///190:4593
overpass-regular.woff2:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
overpass-semibold.woff2:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
overpass-light.woff2:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
overpass-regular.woff:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
overpass-semibold.woff:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
overpass-light.woff:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
overpass-regular.ttf:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
overpass-semibold.ttf:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
overpass-light.ttf:1 Failed to load resource: the server responded with a status of 403 (CORS Rejected - Invalid origin)
www.patternfly.org/c…t/img/favicon.ico:1 Failed to load resource: the server responded with a status of 404 ()
```

When I ran your image, there is still a login page but this time it is basic authentication. It asks for a username and password. This time in the logs, there is this:
```
:8181/api/v1/clienturl:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
webpack-internal:///190:4593 TypeError: Failed to fetch
:8181/api/v1/grafana_datasource_url:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
CommandChannel.service.tsx?9341:194 TypeError: Failed to fetch
:8181/api/v1/grafana_dashboard_url:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
CommandChannel.service.tsx?9341:194 TypeError: Failed to fetch
:8181/api/v1/auth:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
webpack-internal:///190:4235 {}
overpass-regular.woff2:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
overpass-semibold.woff2:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
overpass-light.woff2:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
overpass-regular.woff:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
overpass-semibold.woff:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
overpass-light.woff:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
overpass-regular.ttf:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
overpass-semibold.ttf:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
overpass-light.ttf:1 Failed to load resource: net::ERR_CONNECTION_REFUSED
```

andrewazores:
That's really, really odd. I don't know why the detected auth scheme would be different, or why you would be getting connection refusals when trying to load the fonts from the server.

vic-ma:
Here are the backend logs:
```
[INFO] Logger level: INFO
[INFO] container-jfr started. args: []
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[WARN] No available SSL certificates. Fallback to plain HTTP.
[INFO] HTTP service running on http://10.0.2.100:8181
[INFO] Selecting platform default AuthManager "com.redhat.rhjmc.containerjfr.net.NoopAuthManager"
[INFO] low memory pressure streaming disabled for web server
[INFO] low memory pressure streaming disabled for web server
[INFO] Max concurrent WebSocket connections: 2
```

andrewazores:
Those logs suggest that it didn't even see any incoming connections...

vic-ma:
>Oh, I forgot to also add -p 8181:8181 to the podman command. But without that you shouldn't even have been able to load the client at all to see the basic auth login page, let alone see it failing to load resources after that.

Hmm I didn't add that to the command...

andrewazores:
Ah. You probably got a cached client then, which then failed to connect to the server because the port wasn't exposed.

vic-ma:
Did that and now there is no more login page, but I get these two new messages in the webpage:
```
Danger alert:WebSocket error
{"isTrusted":true}
```
and
```
WebSocket connection lost
```

andrewazores:
Okay, great. That's the same result that I see now with that image. I need to figure out why that is, but at least we've managed to get on the same page.

vic-ma:
I guess what's most strange is that my webpage was working fine just a few days ago. I can see in my browser history that I visited it this Monday. So it must've been something that I changed on my machine from then to now, right?

andrewazores:
It would seem that way, but I don't know what it would be other than maybe the browser, but you've tried several and I doubt that they all somehow are broken at the same time for this specific application. But then it works (more or less) with an image that I built, so that suggests maybe something in your repo. But then you tried from a freshly cloned one...

andrewazores:
Okay so the "websocket connection lost" and "websocket error" messages actually come up *because of this PR*. Whoops. But only when it's in a container image - as a local process it seems to work normally? I need to figure out why that is. I'll mark this PR as a draft in the meantime.

andrewazores:
Oh, nevermind. Use `run.sh` and this PR works fine, so it's just a misconfiguration causing the websocket connection to fail immediately.

andrewazores:
Oh, nevermind. Use `run.sh` and this PR works fine, so it's just a misconfiguration causing the websocket connection to fail immediately.

vic-ma:
When I run `npm run start:dev` in `web-server/` and then go to `localhost:9000`, it has the basic authentication login page, if that means anything. This is without cjfr running or anything.

andrewazores:
That would be expected. If you don't have a container-jfr instance running then the web-client has nothing to connect to, so it fails auth detection (and fails no-auth login) and sits you at the login page.

andrewazores:
When you use `npm run start:dev`, the webpack dev server configures things so that the client should be attempting to connect to a container-jfr backend at `http://localhost:8181/`. That could be a local process or an appropriately configured Podman instance, so long as it's at that location. You can configure that by editing the `.env` file in the web-client.

vic-ma:
This is quite strange...

So this morning, everything was still not working. I tried on master, with both the `run` script and as a local process.

But then, in another copy of the repo, I checked out the connectionless API commit, and when I then ran the `run` script, it was working. Then, when I ran `run.sh` again on master, that was working too. Also, for all this testing, I tried to make sure to do a hard reload each time.

After that, I ran `exec:java` on the connectionless API repo, and that worked. I want to say that when I ran `exec:java` on the master repo right after, that that worked as well, but I'm not totally sure. It could be that I forgot to hard reload also. In any case, afterwards, I could not get `exec:java` on master to work; it always showed the token authentication page.

Then, I tried the commit right before connectionless and that worked for both `run.sh` and `exec:java`. I'm pretty sure I tried `run.sh` on the master repo right after and it still worked.

Then, I tried the commit right before master, and `run.sh` did not work. Then, interestingly, when I then tried `run.sh` on master, it became broken again. And for both, `exec:java` does not work.

Then, I tried the commit right before the `WebServer` refactor, and both `run.sh` and `exec:java` were working. Then, I tried `run.sh` on master and that was working again, but `exec:java` was still not.

Then, I tried the `WebServer` refactor commit, and everything stopped working again.

andrewazores:
The webserver refactor you mean is commit `bc069f526b25b33434604b25c9904ed9365bc8b7`? The CORS handling did somewhat change along with that refactor, although it went from being a hacked-together hand-written implementation to delegating to the Vert.x implementation, so if anything it should be *more* compliant now...

vic-ma:
Yes that one.

andrewazores:
With that commit, you see 403s in the container-jfr logs and CORS rejection messages in the browser console, right? And neither of those things, and the web-client working correctly, when you try the commit before it?

andrewazores:
What URL are you visiting to get to the web-client, and do you have `CONTAINER_JFR_ENABLE_CORS` set?

vic-ma:
Here is for the refactor commit.

CJFR Logs when I open `localhost:8181`:
```
[INFO] (0:0:0:0:0:0:0:1:56700): GET /api/v1/clienturl 200 101ms
[INFO] (0:0:0:0:0:0:0:1:56706): POST /api/v1/auth 403 0ms
[ERROR] io.vertx.ext.web.handler.impl.HttpStatusException: Internal Server Error

[INFO] (0:0:0:0:0:0:0:1:56702): GET /api/v1/grafana_datasource_url 500 23ms
[ERROR] io.vertx.ext.web.handler.impl.HttpStatusException: Internal Server Error

[INFO] (0:0:0:0:0:0:0:1:56704): GET /api/v1/grafana_dashboard_url 500 18ms
```

And  in the `auth` header: `403 CORS Rejected - Invalid origin`.

Maybe worth noting is that I don't get 403s for the font-related stuff anymore.

For the commit before, `auth` header is: `200 OK`. And I tested before and making a recording works.

vic-ma:
URL I use is `localhost:8080`. Right now `CONTAINER_JFR_ENABLE_CORS` is set to true. Though last week I tested with it off as well.

andrewazores:
Try `unset CONTAINER_JFR_ENABLE_CORS` and revisit your `localhost:8181`, and then try with/without the CORS env var set and visiting the URL printed out by container-jfr on startup (assuming you start it using `exec:java`). I think that matrix is the key to figuring out what exactly is broken about how we have CORS implemented.

vic-ma:
So, right now, only tested the `unset` part, but it fixed the issue on the `WebServer` commit when I run `exec:java`. Tested on master, and it's still broken.

vic-ma:
For `WebServer` commit, CORS unset and going to `http://192.168.0.107:8181` works. When CORS is set to true, it does not work.

andrewazores:
There are no differences between that commit and master which should be changing the CORS behaviour :-/

vic-ma:
Ah, my bad, maybe env var unset doesn't work across terminals or something? I unset it and it works on master now.

vic-ma:
Although that is really strange, since this would certainly have been the first thing I tested last week, since it's the most obvious change...

andrewazores:
Yea if you have the env var set in something like your `.bashrc` then unsetting it in one only unsets it for that process and its children, so a different terminal window wouldn't be affected.

andrewazores:
Okay so with master (or the webserver refactor commit), CONTAINER_JFR_ENABLE_CORS=true breaks it if you visit `localhost:8181` or the URL printed at startup. I think that's probably expected - if CORS is enabled then it's expecting and enforcing a different origin, even though you don't actually need it since your origin is actually the container-jfr webserver itself. So I'll need to check if that's up to spec, or if I need to also add those origins to the response headers.

What about the commit before the refactor? Did that work in your test scenario regardless of CONTAINER_JFR_ENABLE_CORS?

vic-ma:
Hmm even using `localhost:8181` is working properly on master now. I think I probably just didn't unset the env var properly last week when I was trying to figure out this issue. I'll test the commit before now.

vic-ma:
On master, if the env var is set, it does not work on either `localhost:8181` or `http://192.168.0.107:8181/`. This is the case even if I set `CONTAINER_JFR_ENABLE_CORS` to false (but not if I completely unset it).

vic-ma:
Before the refactor, it works regardless of if the env var is set or not.

andrewazores:
Cool, thanks for checking all that. I'll check what the spec should be and see if, when CORS is enabled, we should automatically include our own origin(s).

Setting the var to `false`, or any other string, should actually have the same effect as setting it to `true`. It's just looking for the env var to be set to enable the handler, not for it to have any particular value.

vic-ma:
But the problem with setting the variable causing this issue is specific to me, right? Like for you or Alex, having the variable set doesn't cause any problems?

andrewazores:
I don't know about @Alexjsenn , but `CONTAINER_JFR_ENABLE_CORS=true mvn exec:java` breaks due to CORS rejections while `mvn exec:java` works as expected. In the case where the env var is set, using a separate web-client dev instance with `npm run start:dev` does work though, since that does mean that the client origin is different from the container-jfr webserver, and as such is actually the expected request origin. Without the env var set, the separate web-client does not work.

I think all the confusion on my end previously came from me not cluing in that I had just recently suggested the CORS env var to you, not realizing that that breaks local development if you use the client as built by Maven and served by container-jfr. Usually I run a separate local backend process with Maven and a separate frontend with Webpack, so I do need the CORS configuration.

andrewazores:
Okay, so the CORS header `Access-Control-Allow-Origin` only allows a single origin to be specified, not a list, but there is also a `*` wildcard which allows any origin. I'll file an issue to make this configurable.

andrewazores:
Filed: #188

vic-ma:
Is there something wrong with this way of changing the `web-client` submodule's commit? Since when I run CJFR after, I still see the ping commands, but obviously this PR removes them.

1. I clone my fork of `container-jfr`
1. I change the URL in `.gitmodules` to https://github.com/andrewazores/container-jfr-web
1. I run `git submodule init && git submodule update`
1. I go into `web-client/` and run `git checkout remove-ping`
1. I go back into `container-jfr/` and build and run the project

andrewazores:
You shouldn't need to modify the `.gitmodules` just to check out a fork's branch - try adding my fork as a remote to your submodule repo. Then you can fetch and check out branches from my fork, as well as check out the original master branch, without having to manually edit any git state.

vic-ma:
Hmm did this and I still get pings in my logs. These are the pings that this PR removes right?

```
[INFO] (192.168.0.107:44152): CMD {"id":"4eoPU4ypc_MiYP6cHXh9Y","command":"ping","args":[]}
[INFO] (192.168.0.107:44152): CMD {"id":"GQGapWEibZk32lJeeYKNc","command":"ping","args":[]}
[INFO] (192.168.0.107:44152): CMD {"id":"0Z7668086Ro_lqalMW0X3","command":"ping","args":[]}
```

andrewazores:
Yes, those are the ones. I'm not sure where those are coming from, but there must again be some issue where the client build is stale or whatever client build is running in your browser is out of date, or maybe you have another instance of an outdated client still open somewhere and connecting to the same backend, because this PR definitely removes all of the pings from the frontend - try a `grep -R ping src` on the web-client before and after and you'll see that this PR is removing the only place that sends them.

vic-ma:
Ah, yup, looks like a hard refresh fixed it.

PR #96, opened 2020-06-29
Change tabs layout in event view to match the rest of the pages

Alexjsenn:


andrewazores:
Semi-related but maybe still worth including in this same PR - what do you think about removing the "Events" title text on the Events view card? The Recordings view card doesn't have this kind of title, and IMO it seems redundant to have the title for a card that also has tab titles summarizing what the content view is.

image

image

Alexjsenn:
Yes, that makes sense!

PR #97, opened 2020-06-29
enable sidebar to autohide on narrow screen

Alexjsenn:
When the window is resized to a vertical view (narrow width) the sidebar starts covering the page (overlay), and when a page on the menu is chosen the bar now closes automatically (given nothing can be done with the bar open).

### DIFF
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -54,7 +54,8 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
   };
 
   const Navigation = (
-    <Nav id="nav-primary-simple" theme="dark" variant="default">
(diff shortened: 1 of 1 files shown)


src/app/AppLayout/AppLayout.tsx
```diff
@@ -54,7 +54,8 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
   };
 
   const Navigation = (
-    <Nav id="nav-primary-simple" theme="dark" variant="default">
+    <Nav id="nav-primary-simple" theme="dark" variant="default" 
+        onSelect={(selected) => {if(isMobileView) setIsNavOpenMobile(false)}}>
```
andrewazores, line 58: That's kind of an awkward one-liner with the if-statement in there. Could you extract it out to a named function?

PR #98, opened 2020-07-02
Include target connection port number, if any

### LINKED ISSUE #89: Since connectionless API changes, target port numbers are not included in command args

andrewazores:
See #90. Reopening after default branch rename.

Fixes #89

### DIFF
--- a/src/app/TargetSelect/TargetSelect.tsx
+++ b/src/app/TargetSelect/TargetSelect.tsx
@@ -42,7 +42,11 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props) 
     if (isPlaceholder) {
       context.commandChannel.setTarget('');
     } else {
-      context.commandChannel.setTarget(selection.connectUrl);
+      let identifier = selection.connectUrl;
+      if (selection.port) {
+        identifier += `:${selection.port}`;
+      }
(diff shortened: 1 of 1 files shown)


Alexjsenn:
Im trying to test this using CRC, but when I run build-images in container-jfr it fails a test, I can't seem to figure out why, I have tried changing to different commits in both cjfr and web-client to no avail. Not sure how to troubleshoot this.

andrewazores:
Ah... good catch. I think that's valid. `build-images.sh` tries to build a minimal image without the web-client, and then a full-featured one afterward. For each it runs the full unit test and integration test suites. One of the newly merged integration tests has a check that the web-client document received has a `<script>` tag in it as a very basic smoketest - and this will definitely fail for a minimal build, which has no web-client assets at all.

Change the `verify` to a simple `test in `build-images` and you should be able to continue. We'll need to come up with some better way in the future to disable certain integration tests when testing against a minimal image (and some corresponding way to inform the integration test runner that it is testing a minimal image).

Otherwise you can skip using the `build-images.sh` script and just invoke `mvn package` yourself, directly.

PR #99, opened 2020-07-03
Call health endpoint

Alexjsenn:
Request health endpoint to determine if the Grafana configuration is set up, and only if it is then request URLs. Unified error message when configuration is not present.

### DIFF
--- a/src/app/Shared/Services/CommandChannel.service.tsx
+++ b/src/app/Shared/Services/CommandChannel.service.tsx
@@ -1,6 +1,6 @@
 import { Notifications } from '@app/Notifications/Notifications';
 import { nanoid } from 'nanoid';
-import { BehaviorSubject, combineLatest, from, Observable, ReplaySubject, Subject, forkJoin } from 'rxjs';
+import { BehaviorSubject, combineLatest, from, Observable, ReplaySubject, Subject, forkJoin, throwError } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
 import { webSocket, WebSocketSubject } from 'rxjs/webSocket';
 import { concatMap, filter, first, map } from 'rxjs/operators';
@@ -35,17 +35,18 @@ export class CommandChannel {
 
     fromFetch(`${this.apiSvc.authority}/health`)
       .pipe(
-        concatMap(resp => from(resp.json())))
-      .pipe(
-        concatMap(jsonResp => {
-            if((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) 
-              {return forkJoin([getDatasourceURL, getDashboardURL])}
-            else {this.logError('Grafana configuration', 'URLs unavailable');
-                  return jsonResp}}))
+        concatMap(resp => from(resp.json())), 
+        concatMap((jsonResp: any) => {
+          if ((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) {
+            return forkJoin([getDatasourceURL, getDashboardURL])
+          } else {
+              return throwError(new Error('ULRs unavailable'));
+          }}))
       .subscribe(
         (url: any) => 
           {this.grafanaDatasourceUrlSubject.next(url[0].grafanaDatasourceUrl);
-            this.grafanaDashboardUrlSubject.next(url[1].grafanaDashboardUrl);}
+            this.grafanaDashboardUrlSubject.next(url[1].grafanaDashboardUrl);},
+        err => this.logError('Grafana configuration', err)
       );
     
     this.onResponse('list-saved').pipe(
--- a/src/app/Shared/Services/CommandChannel.service.tsx
+++ b/src/app/Shared/Services/CommandChannel.service.tsx
@@ -46,7 +46,7 @@ export class CommandChannel {
         (url: any) => 
           {this.grafanaDatasourceUrlSubject.next(url[0].grafanaDatasourceUrl);
             this.grafanaDashboardUrlSubject.next(url[1].grafanaDashboardUrl);},
-        err => this.logError('Grafana configuration', err)
+        err => this.logError('Grafana configuration not found', err)
       );
     
     this.onResponse('list-saved').pipe(
--- a/src/app/Shared/Services/CommandChannel.service.tsx
+++ b/src/app/Shared/Services/CommandChannel.service.tsx
@@ -37,18 +37,18 @@ export class CommandChannel {
       .pipe(
         concatMap(resp => from(resp.json())), 
         concatMap((jsonResp: any) => {
-          if ((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) {
-            return forkJoin([getDatasourceURL, getDashboardURL])
+          if (jsonResp.dashboardAvailable && jsonResp.datasourceAvailable) {
+            return forkJoin([getDatasourceURL, getDashboardURL]);
           } else {
-            var errMessage = "";
-            if (jsonResp.dashboardAvailable == false) {
-              errMessage = "dashboard URL "
+            const missing: string[] = [];
+            if (!jsonResp.dashboardAvailable) {
+              missing.push('dashboard URL');
             }
-            if (jsonResp.datasourceAvailable == false) {
-              errMessage = errMessage.concat(
-                (errMessage == "")? "datasource URL " : ", datasource URL ")
+            if (!jsonResp.datasourceAvailable) {
+              missing.push('datasource URL');
             }
-            return throwError(errMessage.concat('missing'));
+            const message = missing.join(', ') + ' unavailable';
+            return throwError(message);
           }}))
       .subscribe(
         (url: any) => {
--- a/src/app/Shared/Services/CommandChannel.service.tsx
+++ b/src/app/Shared/Services/CommandChannel.service.tsx
@@ -1,6 +1,6 @@
 import { Notifications } from '@app/Notifications/Notifications';
 import { nanoid } from 'nanoid';
-import { BehaviorSubject, combineLatest, from, Observable, ReplaySubject, Subject } from 'rxjs';
+import { BehaviorSubject, combineLatest, from, Observable, ReplaySubject, Subject, forkJoin } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
 import { webSocket, WebSocketSubject } from 'rxjs/webSocket';
 import { concatMap, filter, first, map } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -27,21 +27,27 @@ export class CommandChannel {
...
-    fromFetch(`${this.apiSvc.authority}/api/v1/grafana_dashboard_url`)
-      .pipe(concatMap(resp => from(resp.json())))
+    fromFetch(`${this.apiSvc.authority}/health`)
+      .pipe(
+        concatMap(resp => from(resp.json())))
+      .pipe(
+        concatMap(jsonResp => {
+            if((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) 
+              {return forkJoin([getDatasourceURL, getDashboardURL])}
+            else {this.logError('Grafana configuration', 'URLs unavailable');
```
andrewazores, line 43: These curly braces are hard to read. Please reformat it similar to:

```
if (condition) {
  return something;
} else {
  doLog();
  return somethingElse;
}
```

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -27,21 +27,27 @@ export class CommandChannel {
...
-        (url: any) => this.grafanaDatasourceUrlSubject.next(url.grafanaDatasourceUrl),
-        (err: any) => this.logError('Grafana Datasource configuration', err)
-      );
-
-    fromFetch(`${this.apiSvc.authority}/api/v1/grafana_dashboard_url`)
-      .pipe(concatMap(resp => from(resp.json())))
+    fromFetch(`${this.apiSvc.authority}/health`)
+      .pipe(
+        concatMap(resp => from(resp.json())))
+      .pipe(
```
andrewazores, line 39: You can pass more than one operator into a `pipe`, and they are piped through in that order - so `.pipe(funcA).pipe(funcB)` should be equivalent to `.pipe(funcA, funcB)`. More idiomatic that way.

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -27,21 +27,27 @@ export class CommandChannel {
...
-      .subscribe(
-        (url: any) => this.grafanaDatasourceUrlSubject.next(url.grafanaDatasourceUrl),
-        (err: any) => this.logError('Grafana Datasource configuration', err)
-      );
-
-    fromFetch(`${this.apiSvc.authority}/api/v1/grafana_dashboard_url`)
-      .pipe(concatMap(resp => from(resp.json())))
+    fromFetch(`${this.apiSvc.authority}/health`)
+      .pipe(
+        concatMap(resp => from(resp.json())))
```
andrewazores, line 38: I think you could just `map(resp => resp.json())` here - no need to make the json into an Observable and then immediately concat it in this case, from what I can tell.

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -27,21 +27,27 @@ export class CommandChannel {
         (url: any) => this.clientUrlSubject.next(url.clientUrl),
         (err: any) => this.logError('Client URL configuration', err)
       );
+    
+    const getDatasourceURL = fromFetch(`${this.apiSvc.authority}/api/v1/grafana_datasource_url`)
+    .pipe(concatMap(resp => from(resp.json())));
+    const getDashboardURL = fromFetch(`${this.apiSvc.authority}/api/v1/grafana_dashboard_url`)
+    .pipe(concatMap(resp => from(resp.json())));
```
andrewazores, line 34: These `concatMap(resp => from(foo))`s look like they could also just be `map(resp -> foo)`

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -27,21 +27,27 @@ export class CommandChannel {
...
-      .pipe(concatMap(resp => from(resp.json())))
+    fromFetch(`${this.apiSvc.authority}/health`)
+      .pipe(
+        concatMap(resp => from(resp.json())))
+      .pipe(
+        concatMap(jsonResp => {
+            if((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) 
+              {return forkJoin([getDatasourceURL, getDashboardURL])}
+            else {this.logError('Grafana configuration', 'URLs unavailable');
+                  return jsonResp}}))
```
andrewazores, line 44: I'm a little confused on this - in the `if` branch we're returning an Observable with the `forkJoin` operator. But in the `else` branch, this looks like it's returning a simple JSON object from the HTTP response, isn't it?

I think the error handling can be cleaned up here by returning the same `forkJoin` in the `if` branch, and returning a `throw` operator in this `else` branch (and not logging the error, yet). Then, the subscriber below should gain an error handler, and this is what should call `logError`.

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -35,17 +35,18 @@ export class CommandChannel {
...
     fromFetch(`${this.apiSvc.authority}/health`)
       .pipe(
-        concatMap(resp => from(resp.json())))
-      .pipe(
-        concatMap(jsonResp => {
-            if((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) 
-              {return forkJoin([getDatasourceURL, getDashboardURL])}
-            else {this.logError('Grafana configuration', 'URLs unavailable');
-                  return jsonResp}}))
+        concatMap(resp => from(resp.json())),
```
andrewazores, line 38: Did the `concatMap(resp => from(resp.json()))` -> `map(resp => resp.json())` transform not apply here?

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -35,17 +35,18 @@ export class CommandChannel {
...
+          if ((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) {
+            return forkJoin([getDatasourceURL, getDashboardURL])
+          } else {
+              return throwError(new Error('ULRs unavailable'));
+          }}))
       .subscribe(
         (url: any) => 
           {this.grafanaDatasourceUrlSubject.next(url[0].grafanaDatasourceUrl);
-            this.grafanaDashboardUrlSubject.next(url[1].grafanaDashboardUrl);}
+            this.grafanaDashboardUrlSubject.next(url[1].grafanaDashboardUrl);},
```
andrewazores, line 48: Seems like this was pre-existing, but I think this handler should also be formatted with some newlines.

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -35,17 +35,18 @@ export class CommandChannel {
...
-            if((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) 
-              {return forkJoin([getDatasourceURL, getDashboardURL])}
-            else {this.logError('Grafana configuration', 'URLs unavailable');
-                  return jsonResp}}))
+        concatMap(resp => from(resp.json())), 
+        concatMap((jsonResp: any) => {
+          if ((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) {
+            return forkJoin([getDatasourceURL, getDashboardURL])
+          } else {
+              return throwError(new Error('ULRs unavailable'));
```
andrewazores, line 43: s/ULR/URL/

Could this error include details about which URL(s) in particular are not available?

src/app/Shared/Services/CommandChannel.service.tsx
Alexjsenn: So Ive been trying to figure out why it doesn't work there or above in the getDatasourceURL either, and I think it is because the concatMap waits for the resp.json() to finish before proceding, which the map doesn't. Im not 100% sure this is why, but when I tried changing it it would not parse the URLs

andrewazores: Strange. I'm not sure if it's because of the `.json()` "not finishing", but maybe something about how the `fromFetch` emits values and when it emits (or doesn't) its completion event. It's fine to leave it with the `concatMap` in that case, if that's the way that works.

src/app/Shared/Services/CommandChannel.service.tsx
Alexjsenn: I added specific messages depending on which is missing.

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -40,12 +40,21 @@ export class CommandChannel {
           if ((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) {
             return forkJoin([getDatasourceURL, getDashboardURL])
           } else {
-              return throwError(new Error('ULRs unavailable'));
+            var errMessage = "";
```
andrewazores, line 43: `var` should be replaced by `let`: https://www.typescriptlang.org/docs/handbook/variable-declarations.html

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -40,12 +40,21 @@ export class CommandChannel {
...
+            if (jsonResp.dashboardAvailable == false) {
+              errMessage = "dashboard URL "
+            }
+            if (jsonResp.datasourceAvailable == false) {
+              errMessage = errMessage.concat(
+                (errMessage == "")? "datasource URL " : ", datasource URL ")
+            }
+            return throwError(errMessage.concat('missing'));
```
andrewazores, line 51: This line uses a single-quote-delimited string, but above it's double-quotes. In TypeScript I tend to use single-quotes, other than when using backticks for string interpolation. Double quotes are used when the quoted text literal contains single-quotes/apostrophes, or in React TSX template fragments. Just a style thing.

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -40,12 +40,21 @@ export class CommandChannel {
           if ((jsonResp.dashboardAvailable == true) && (jsonResp.datasourceAvailable == true)) {
             return forkJoin([getDatasourceURL, getDashboardURL])
           } else {
-              return throwError(new Error('ULRs unavailable'));
+            var errMessage = "";
+            if (jsonResp.dashboardAvailable == false) {
```
andrewazores, line 44: `if (variable == false)` should be written as `if (!variable)`

src/app/Shared/Services/CommandChannel.service.tsx
```diff
@@ -40,12 +40,21 @@ export class CommandChannel {
...
+              errMessage = "dashboard URL "
+            }
+            if (jsonResp.datasourceAvailable == false) {
+              errMessage = errMessage.concat(
+                (errMessage == "")? "datasource URL " : ", datasource URL ")
```
andrewazores, line 49: This error message building is pretty awkward. Most of this ternary is the same, only the comma and additional space are different between the cases, and the `concat` is also a bit unusual.

What about something like this:

```
const missing = [];
if (!dashboardAvailable) {
  missing.push('Dashboard URL');
}
if (!datasourceAvailable) {
  missing.push('datasource URL');
}
const message = missing.join(', ') + ' unavailable.';
return throwError(message);
```

src/app/Shared/Services/CommandChannel.service.tsx
andrewazores: Also, `===` (triple-equals) should be used in nearly all cases.

src/app/Shared/Services/CommandChannel.service.tsx
Alexjsenn: yes, much better, should be fixed.

PR #101, opened 2020-07-09
Add per-file license headers

Alexjsenn:


andrewazores:
How were these headers added? Did you do this by hand?

Alexjsenn:
I used the npm package you had linked, https://www.npmjs.com/package/license-check-and-add.
Im not sure if I should leave the configuration file for it, and the package, or not.

andrewazores:
Yes, whatever changes you made to the project to include and use that package should definitely be included. We don't want to have to reopen an issue and redo the work you've done the next time a new file is added to the project.

Alexjsenn:
Ok, except the package does not do it automatically, a command has to be run. How would that be specified?

andrewazores:
Of course, it would be very weird if it somehow did it in the background.

Add the package to the project dependencies with `npm install --save-dev`, then edit `package.json` and add a `script` entry that runs whatever command provided by the package.

Alexjsenn:
It will now check and insert license headers when the format script is run.

package.json
```diff
@@ -14,7 +14,7 @@
     "test": "jest",
     "eslint": "eslint --ext .tsx,.js ./src/",
     "lint": "npm run eslint",
```
andrewazores, line 16: Is there a `license-check-and-add check` or something similar that could be added to the `lint` script as well so that it fails if any files are missing their headers?

andrewazores:
Looks good other than that comment about linting above. Thanks!

Alexjsenn:
It is giving a ton of errors when lint is run though, not sure if something should be done about that.

andrewazores:
That's currently normal. There is a bunch of cleanup that needs to be done, particularly with the type definitions used by the Command Channel. Those were ported over from the old Angular client and adjusted to work, but they don't meet the linter's standards now.

PR #104, opened 2020-07-15
Update README to reflect change of env var for configurable cors origin

Alexjsenn:


andrewazores:
I'll merge this when rh-jmc-team/container-jfr#204 is finished and approved, too, since they go together.

PR #108, opened 2020-07-24
Support Backend Changes to `upload-recording` that Assumes Datasource URL

### LINKED ISSUE #86: `upload-recording` command should not include datasource URL
https://github.com/rh-jmc-team/container-jfr/issues/166

When the backend command channel API for the `upload-recording` command is updated to assume that the datasource URL should be the one configured on the ContainerJFR deployment, the -web client must also be updated to not include the datasource URL as an argument when sending this command message.

vic-ma:
Fixes #86.

I guess something worth noting is that before, the web client would append `/load` to the datasource URL env var, but now, the command just uses the env var directly. So a previous env var of `http://localhost:8080` should be changed to `http://localhost:8080/load`.

### DIFF
--- a/src/app/RecordingList/ActiveRecordingsList.tsx
+++ b/src/app/RecordingList/ActiveRecordingsList.tsx
@@ -308,7 +308,7 @@ export const RecordingActions: React.FunctionComponent<RecordingActionsProps> = 
       notifications.info('Upload Started', `Recording "${props.recording.name}" uploading...`);
       const id = context.commandChannel.createMessageId();
       setUploadIds(ids => [...ids, id]);
-      context.commandChannel.sendMessage('upload-recording', [ props.recording.name, `${url}/load` ], id);
+      context.commandChannel.sendMessage('upload-recording', [ props.recording.name ], id);
     });
   };
 


andrewazores:
Looks good.

Since the required env var value has changed as well, then we also need to update the Operator to match. In `containerjfr_controller.go` there is a struct with a member that is assigned to the cluster service URL of the datasource, so the `/load` path just needs to be appended there.

andrewazores:
Actually, on second thought - I'm not sure I like that the environment variable needs to also specify the path on the datasource. There could well be a situation in the future where we want to be able to use different datasource paths, since those are how different capabilities of the datasource are exposed (ex. re-selecting a previously-uploaded recording). I've dismissed my review until we talk this over and figure out what course of action is best, but I think this PR itself is probably fine - it's the corresponding backend PR that should be tweaked slightly. IMO the environment variable should point to the root location of the datasource, and the upload command is what should carry the knowledge to append the `/load` path for its specific task. The web-client here as well as the Operator and anything else that might consume that API should still not need to care about the datasource URL or any paths at all.

vic-ma:
Yeah, now that you mention it, I agree. It's probably a lot easier on the user if they just need to point to the datasource itself and don't have to worry about its API. And it probably makes sense that `GRAFANA_DATASOURCE_URL` is just the URL to the datasource, not to the loading function. So I'll start working on this change now.

And one question: do we assume that the datasource the user uses will always be jfr-datasource, or could there potentially be other supported datasources in the future?

andrewazores:
I think we can assume it will always be jfr-datasource, or if the user decides to swap in their own implementation then they should be following the same API as it at least.

PR #110, opened 2020-07-31
Replace Command Channel usage with HTTP REST APIs

andrewazores:
See rh-jmc-team/container-jfr#222

This PR replaces all uses of the command channel with HTTP calls instead, in preparation for support for RJMX authentication (rh-jmc-team/container-jfr#225) to come later.

### DIFF
--- a/src/app/CreateRecording/CreateRecording.tsx
+++ b/src/app/CreateRecording/CreateRecording.tsx
@@ -66,40 +66,21 @@ const Comp: React.FunctionComponent< RouteComponentProps<{}, StaticContext, Crea
   const [activeTab, setActiveTab] = React.useState(0);
 
   const handleCreateRecording = (recordingName: string, events: string, duration?: number): void => {
-    const form = new FormData();
-    form.append('recordingName', recordingName);
-    form.append('events', events);
-    if (!!duration && duration > 0) {
-      form.append('duration', String(duration));
-    }
     context.commandChannel.target().pipe(
-      concatMap(target => context.api.sendRequest(`targets/${encodeURIComponent(target)}/recordings`, {
-        method: 'POST',
-        body: form,
-      })),
-      first(),
-    ).subscribe(resp => {
-      if (200 <= resp.status && resp.status < 300) {
-        notifications.success('Recording created');
+      concatMap(targetId => context.api.createRecording(targetId, { recordingName, events, duration }))
+    ).subscribe(success => {
+      if (success) {
         history.push('/recordings');
-      } else {
-        notifications.danger(`Request failed (Status ${resp.status})`, resp.statusText)
       }
     });
   };
 
   const handleCreateSnapshot = (): void => {
     context.commandChannel.target().pipe(
-      concatMap(target => context.api.sendRequest(`targets/${encodeURIComponent(target)}/snapshot`, {
-        method: 'POST',
-      })),
-      first(),
-    ).subscribe(resp => {
-      if (200 <= resp.status && resp.status < 300) {
-        notifications.success('Recording created');
+      concatMap(targetId => context.api.createSnapshot(targetId))
+    ).subscribe(success => {
+      if (success) {
         history.push('/recordings');
-      } else {
-        notifications.danger(`Request failed (Status ${resp.status})`, resp.statusText)
       }
     });
   };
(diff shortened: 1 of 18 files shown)


src/app/CreateRecording/CreateRecording.tsx
```diff
@@ -62,24 +63,32 @@ const Comp: React.FunctionComponent< RouteComponentProps<{}, StaticContext, Crea
...
-    context.commandChannel.onResponse(command).pipe(
-      filter(m => m.id === id),
-      first(),
-      )
-      .subscribe((msg) => {
-          if (msg.status === 0) {
-            notifications.success('Recording created');
-            history.push('/recordings');
-          } else {
-            notifications.danger(`Request failed (Status ${msg.status})`, msg.payload)
```
jiekang, line 79: Would it make sense to notify on failure in the new subscription below? I don't see handling of that at the moment.

andrewazores: There's a `tap` in the `context.api.createRecording` (Api.service.tsx line 14) that intercepts the response and displays the equivalent notifications that were removed here, so the handling is in the end unchanged, just done somewhere else.

jiekang:
When testing through c-jfr:

Using `MAVEN_OPTS="-Dcom.sun.management.jmxremote.port=9091 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.autodiscovery=true" mvn clean prepare-package exec:java` I no longer see c-jfr in targets to connect to.

When using `run.sh` with `CONTAINER_JFR_LISTEN_PORT=9191` (as cockpit takes 9090 for my system), I am not able to view reports, download reports or download jfr files. Creating recordings and archiving them shows success in the UI. I see some

`OPTIONS /api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcontainer-jfr:9091%2Fjmxrmi/reports/tmp 405 1ms`

Would that be CORS related?

and also a minor 'on the side':

`GET /api/v1//targets/service%3Ajmx%3Armi%3A%2F%2F%2Fjndi%2Frmi%3A%2F%2Fcontainer-jfr%3A9091%2Fjmxrmi/recordings 200 29ms`

(extra `/` between `v1` and `targets`)

andrewazores:
> When testing through c-jfr:
> 
> Using `MAVEN_OPTS="-Dcom.sun.management.jmxremote.port=9091 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.autodiscovery=true" mvn clean prepare-package exec:java` I no longer see c-jfr in targets to connect to.

There is no longer the "localhost loopback" selection item, but you should be able to see the JMX Service URL from its JDP self-discovery still. Is that not the case? If not, that would be a separate bug than from this PR, I think. Your next comment suggests that you do see container-jfr in the targets list.

> 
> When using `run.sh` with `CONTAINER_JFR_LISTEN_PORT=9191` (as cockpit takes 9090 for my system), I am not able to view reports, download reports or download jfr files. Creating recordings and archiving them shows success in the UI. I see some
> 
> `OPTIONS /api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcontainer-jfr:9091%2Fjmxrmi/reports/tmp 405 1ms`
> 
> Would that be CORS related?

Yes, I think anything with a `405`, especially with the `OPTIONS` verb, is a CORS failure. I'll test that out myself and see what's causing the failure, but it seems like it would be a backend bug.

`CONTAINER_JFR_LISTEN_PORT` is used for the TCP socket, which isn't used unless you're running container-jfr in that mode. If you're running it normally it's in WebSocket mode, which does not use that socket and so the env var is also unused. You're probably looking for `CONTAINER_JFR_WEB_PORT`, which controls the port number used for the HTTP/WebSocket server, but that defaults to `8181`.

> 
> and also a minor 'on the side':
> 
> `GET /api/v1//targets/service%3Ajmx%3Armi%3A%2F%2F%2Fjndi%2Frmi%3A%2F%2Fcontainer-jfr%3A9091%2Fjmxrmi/recordings 200 29ms`
> 
> (extra `/` between `v1` and `targets`)

Good catch, fixed.

jiekang:
@andrewazores looks fine to me. Can you resolve the conflict and double-check changed files for removal of unused imports and the like?

andrewazores:
There are more linting issues still reported, but they're either a) type definition issues in the CommandChannel (use of `any` primarily) which are out of scope of this PR, or b) React hook dependency issues, which need a closer look to ensure that any fixes are correct and don't break component update cycles. I'll come back around and fix those separately.

PR #114, opened 2020-08-07
remove download option for "ALL" template

### LINKED ISSUE #112: "ALL" meta-template should not be downloadable
All templates currently have a "Download" action, allowing the user to download the template definition as XML to their local machine. The "ALL" template is not a real XML template however - its name is a special token used by ContainerJFR to signify a meta-template computed at runtime. When an attempt is made to download the ALL "template", the response body is simply "Not Found", which is then written into the "ALL.xml" saved to the user's machine. Instead, the ALL template simply should not be downloadable - the action button should be removed or disabled for this template entry.

Alexjsenn:
Fixes #112

### DIFF
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -125,7 +125,7 @@ export const EventTemplates = () => {
     ] as IAction[];
 
     const template: EventTemplate = filteredTemplates[extraData.rowIndex];
-    if ((template.name !== 'ALL')||(template.provider !== 'ContainerJFR')) {
+    if ((template.name !== 'ALL')||(template.type !== 'TARGET')) {
       actions = actions.concat([
           {
             title: 'Download',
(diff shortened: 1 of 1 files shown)


src/app/Events/EventTemplates.tsx
```diff
@@ -122,13 +122,17 @@ export const EventTemplates = () => {
...
         onClick: (event, rowId, rowData) => history.push({ pathname: '/recordings/create', state: { template: rowData[0], templateType: String(rowData[3]).toUpperCase() } }),
       },
-      {
-        title: 'Download',
-        onClick: (event, rowId) => context.commandChannel.target().pipe(first()).subscribe(target => context.api.downloadTemplate(target, filteredTemplates[rowId])),
-      },
     ] as IAction[];
 
     const template: EventTemplate = filteredTemplates[extraData.rowIndex];
+    if (template.name != 'ALL') {
```
andrewazores, line 128: Should be `!==`

andrewazores:
This also prevents download of templates named `ALL` if the user uploads such a template themselves, which I don't think is desirable. This should only prevent downloading of the template named `ALL` that is defined by ContainerJFR itself.

Alexjsenn:
I don't actually know how the `provider` name is determined, but I assume that it would never be ContainerJFR except for this default one.

andrewazores:
The `provider` name is embedded in the template XML. Don't use the `provider`, use the template `.type`.

andrewazores:
Looks good. Please amend (`git commit --amend`) your last commit and add the "Fixes #something" text (replace with the issue number) so that the merge will auto-close the issue.

PR #118, opened 2020-08-12
Implement archive re-upload

### LINKED ISSUE #113: Users should be able to upload recordings into archives
This feature was present in the old Angular client and has not yet been ported to the React client.

andrewazores:
Fixes #113

### DIFF
--- a/src/app/RecordingList/ArchiveUploadModal.tsx
+++ b/src/app/RecordingList/ArchiveUploadModal.tsx
@@ -37,6 +37,7 @@
  */
 import * as React from 'react';
 import { ActionGroup, Button, FileUpload, Form, FormGroup, Modal, ModalVariant } from '@patternfly/react-core';
+import { first, tap } from 'rxjs/operators';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { NotificationsContext } from '@app/Notifications/Notifications';
 
@@ -53,6 +54,13 @@ export const ArchiveUploadModal: React.FunctionComponent<ArchiveUploadModalProps
   const [uploading, setUploading] = React.useState(false);
   const [rejected, setRejected] = React.useState(false);
 
+  const reset = () => {
+    setUploadFile(undefined);
+    setFilename('');
+    setUploading(false);
+    setRejected(true);
+  };
+
   const handleFileChange = (file, filename) => {
     setRejected(false);
     setUploadFile(file);
@@ -64,8 +72,7 @@ export const ArchiveUploadModal: React.FunctionComponent<ArchiveUploadModalProps
   };
 
   const handleClose = () => {
-    setUploadFile(undefined);
-    setFilename('');
+    reset();
     props.onClose();
   };
 
@@ -75,12 +82,15 @@ export const ArchiveUploadModal: React.FunctionComponent<ArchiveUploadModalProps
       return;
     }
     setUploading(true);
-    context.api.uploadRecording(uploadFile).subscribe(success => {
-      setUploading(false);
-      if (success) {
-        handleClose();
-      }
-    });
+    context.api.uploadRecording(uploadFile)
+      .pipe(
+        first(),
+        tap(() => setUploading(false)),
+      )
+      .subscribe(handleClose, err => {
+        notifications.warning('Upload Failed',  err);
+        reset();
+      });
   };
 
   return (
--- a/src/app/RecordingList/ArchiveUploadModal.tsx
+++ b/src/app/RecordingList/ArchiveUploadModal.tsx
@@ -38,6 +38,7 @@
 import * as React from 'react';
 import { ActionGroup, Button, FileUpload, Form, FormGroup, Modal, ModalVariant } from '@patternfly/react-core';
 import { ServiceContext } from '@app/Shared/Services/Services';
+import { NotificationsContext } from '@app/Notifications/Notifications';
 
 export interface ArchiveUploadModalProps {
   visible: boolean;
@@ -46,6 +47,7 @@ export interface ArchiveUploadModalProps {
 
 export const ArchiveUploadModal: React.FunctionComponent<ArchiveUploadModalProps> = props => {
   const context = React.useContext(ServiceContext);
+  const notifications = React.useContext(NotificationsContext);
   const [uploadFile, setUploadFile] = React.useState(undefined as File | undefined);
   const [filename, setFilename] = React.useState('' as string | undefined);
   const [uploading, setUploading] = React.useState(false);
@@ -69,7 +71,7 @@ export const ArchiveUploadModal: React.FunctionComponent<ArchiveUploadModalProps
 
   const handleSubmit = () => {
     if (!uploadFile) {
-      window.console.error('Attempted to submit JFR upload without a file selected');
+      notifications.warning('Attempted to submit JFR upload without a file selected');
       return;
     }
     setUploading(true);
--- a/src/app/RecordingList/ArchiveUploadModal.tsx
+++ b/src/app/RecordingList/ArchiveUploadModal.tsx
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2020 Red Hat, Inc.
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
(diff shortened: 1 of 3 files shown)


vic-ma:
When I run CJFR as a local process, I don't see CJFR in targets anymore. I'm not sure if my problem is the same as @jiekang's, but the first commit where it stopped working for me is rh-jmc-team/container-jfr#209. Is there a new flag or setting that needs to be set, to have the local process work with the JMX auth changes?

Also, I forget but are we supposed to be able to access a `run.sh` CJFR with the web-client? In this case, even if I use a really old commit, I don't see CJFR in the targets. If we are supposed to, then there's probably something wrong with my setup.

jiekang:
I don't see targets in local process mode either. I do see targets when using run.sh, though I still haven't figured out my CORS issue that happens after that for downloading recordings, etc.

andrewazores:
> When I run CJFR as a local process, I don't see CJFR in targets anymore. I'm not sure if my problem is the same as @jiekang's, but the first commit where it stopped working for me is [rh-jmc-team/container-jfr#209](https://github.com/rh-jmc-team/container-jfr/pull/209). Is there a new flag or setting that needs to be set, to have the local process work with the JMX auth changes?

You are probably seeing the results of this excerpt:

> The SelfDisoveryPlatformClient is also removed and JDP autodiscovery is enabled instead so that the ContainerJFR instance can use JDP to discover itself within its container in a more uniform manner, providing itself with its own JMXServiceURL and Main-Class name. In environments where JDP autodiscovery is not used, ex. Kubernetes/OpenShift, this also has the net effect of preventing the self-connection list item from appearing in the discovered targets list.

If you are running as a local process with `mvn exec:java`, you must also include the `-Dcom.sun.management.jmxremote.autodiscovery=true` JVM flag to turn on JDP, which will in turn allow ContainerJFR to discover itself. The "This ContainerJFR" entry was provided by the `SelfDiscoveryPlatformClient`. Now, with JDP enabled and running as a local process or in Podman (`run.sh`), you should instead see some JMX Service URL representing ContainerJFR itself instead. JDP is asynchronous, so if it doesn't appear give it a couple of seconds and press the refresh icon/retry your query.

FWIW I would recommend to normally run with something like `mvn clean package && sh run.sh`, since `run.sh` should be kept up-to-date with whatever configuration is needed to run correctly, so you don't need to worry about setting additional JVM flags or environment variables.

> 
> Also, I forget but are we supposed to be able to access a `run.sh` CJFR with the web-client? In this case, even if I use a really old commit, I don't see CJFR in the targets. If we are supposed to, then there's probably something wrong with my setup.

I'm assuming you mean connecting to a ContainerJFR backed run using `run.sh` from a web-client run using `npm run start:dev`? That should work perfectly fine, just be sure to export `CONTAINER_JFR_CORS_ORIGIN=http://localhost:9000` before you `run.sh`.

andrewazores:
> I don't see targets in local process mode either. I do see targets when using run.sh, though I still haven't figured out my CORS issue that happens after that for downloading recordings, etc.

This one is weird. I'm not sure why requests for targets would pass CORS and other requests might not. I know we talked about `CONTAINER_JFR_LISTEN_PORT` before and how you believed that to be conflicting with another service, although that port and its env var are unused when ContainerJFR is in websocket mode (as it is normally and how it must be if you're able to use the web-client at all). Do you have `CONTAINER_JFR_WEB_PORT`, `CONTAINER_JFR_EXT_WEB_PORT`, `CONTAINER_JFR_WEB_HOST`, or any other environment variables set?

jiekang:
> > When I run CJFR as a local process, I don't see CJFR in targets anymore. I'm not sure if my problem is the same as @jiekang's, but the first commit where it stopped working for me is [rh-jmc-team/container-jfr#209](https://github.com/rh-jmc-team/container-jfr/pull/209). Is there a new flag or setting that needs to be set, to have the local process work with the JMX auth changes?
> 
> You are probably seeing the results of this excerpt:
> 
> > The SelfDisoveryPlatformClient is also removed and JDP autodiscovery is enabled instead so that the ContainerJFR instance can use JDP to discover itself within its container in a more uniform manner, providing itself with its own JMXServiceURL and Main-Class name. In environments where JDP autodiscovery is not used, ex. Kubernetes/OpenShift, this also has the net effect of preventing the self-connection list item from appearing in the discovered targets list.
> 
> If you are running as a local process with `mvn exec:java`, you must also include the `-Dcom.sun.management.jmxremote.autodiscovery=true` JVM flag to turn on JDP, which will in turn allow ContainerJFR to discover itself. The "This ContainerJFR" entry was provided by the `SelfDiscoveryPlatformClient`. Now, with JDP enabled and running as a local process or in Podman (`run.sh`), you should instead see some JMX Service URL representing ContainerJFR itself instead. JDP is asynchronous, so if it doesn't appear give it a couple of seconds and press the refresh icon/retry your query.
> 

Ah; the README command has been updated by I've still been running off my old reverse-search version. Nice intuition to find this issue! The local mode now discovers itself, but I still get 405s :|

vic-ma:
>If you are running as a local process with mvn exec:java, you must also include the -Dcom.sun.management.jmxremote.autodiscovery=true

Ah, yeah that fixed it for me.

>I'm assuming you mean connecting to a ContainerJFR backend run using run.sh from a web-client run using npm run start:dev?

I've confused myself a bit going back and forth between commits, so I'll just focus on what happens on main.

I was just using `localhost:8181`, but even if I do `CONTAINER_JFR_CORS_ORIGIN=http://localhost:9000 sh run.sh -it` and start a separate web client and go to `localhost:9000`, it isn't working. I am getting these exceptions in logs when I open the web client (either `8181` or `9000`):
```
[INFO] Max concurrent WebSocket connections: 2
> [INFO] (127.0.0.1:38294): GET /api/v1/clienturl 200 106ms
[INFO] java.io.IOException: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:8080
	at com.redhat.rhjmc.containerjfr.net.web.handlers.HealthGetHandler.lambda$checkUri$0(HealthGetHandler.java:133)
	at io.vertx.ext.web.client.impl.HttpContext.handleFailure(HttpContext.java:309)
...
Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:3000
Caused by: java.net.ConnectException: Connection refused
	at java.base/sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)

...

[INFO] java.io.IOException: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:3000
	at com.redhat.rhjmc.containerjfr.net.web.handlers.HealthGetHandler.lambda$checkUri$0(HealthGetHandler.java:133)
	at io.vertx.ext.web.client.impl.HttpContext.handleFailure(HttpContext.java:309)
...
Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:3000
Caused by: java.net.ConnectException: Connection refused
	at java.base/sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)

...

[INFO] (127.0.0.1:38298): GET /health 200 305ms
[INFO] (127.0.0.1:38306): POST /api/v1/auth 200 1ms
[INFO] Connected remote client 127.0.0.1:38314
[INFO] (127.0.0.1:38314): CMD {"id":"ncjopWiRWfVUeBirjnfMX","command":"list-saved","args":[]}
[INFO] (127.0.0.1:38314): CMD {"id":"7etdhjRR4AYfDmBhb3G4P","command":"scan-targets","args":[]}
```
And in the webpage, I can't click on the "Select Target..." button. The errors mention the health handler, but Grafana and jfr-datsource are both running and should be working; I tested uploading with a local process and it worked. Although we said that the upload shouldn't work with the run script anyway since it's in a container and grafana isn't, right?

Here is the network monitor:

openingweb

andrewazores:
Why are you doing `sh run.sh -it`? That puts ContainerJFR into interactive mode, so it runs the interactive executor rather than the WebSocket executor. In this case I don't believe the server even starts listening for incoming WS connections, and if it does, there is nothing on the ContainerJFR side there to respond to any messages sent (including connection handshake messages). Try again without the `-it` argument.

vic-ma:
Yeah, removing `-it` fixed it. I haven't used the run script in a long time since I've been working on `upload-recording`, so I forgot :|

andrewazores:
Rebased.

vic-ma:
In main (as well as here since the rebase), I can't seem to upload an archived recording anymore, after #110. Regular recording uploads still work. Archived recording uploads work on the commit before.

```
Danger alert:Request failed (Status 404)
Not Found
```

```
[WARN] io.vertx.ext.web.handler.impl.HttpStatusException: Not Found
Caused by: com.redhat.rhjmc.containerjfr.net.internal.reports.ReportService$RecordingNotFoundException: Recording victor-work_foo1_20200817T145623Z.jfr not found in target service:jmx:rmi:///jndi/rmi://victor-work:9091/jmxrmi
	at com.redhat.rhjmc.containerjfr.net.web.handlers.TargetRecordingUploadPostHandler.lambda$doPost$0(TargetRecordingUploadPostHandler.java:135)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at com.redhat.rhjmc.containerjfr.net.web.handlers.TargetRecordingUploadPostHandler.lambda$doPost$1(TargetRecordingUploadPostHandler.java:133)
	at com.redhat.rhjmc.containerjfr.net.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:86)
...
```

andrewazores:
Ah, yea I see the cause of the bug. Good catch. I'll fix that separately.

vic-ma:
Could you rebase again, and I'll re-review with the fixed archive upload?

vic-ma:
On an unrelated note, I was thinking, when a download-recording fails (e.g., due to the recording being deleted from the backend), would it be better for the download not to go off at all? Since #120 added an error message popup for when it does fail, so the user doesn't have to rely on the file contents saying the file was deleted. Or is it good that it always returns a file so it's consistent?

andrewazores:
No, I agree that it would be better to not cause a "X not found" file download to occur. That should instead just result in an error notification being shown.

andrewazores:
@vic-ma would you like to take a look at that bug? It's a pretty easy one.

vic-ma:
Sure.

PR #120, opened 2020-08-17
Fix bug preventing archived recording uploads

### LINKED ISSUE #119: Archived recordings cannot be uploaded to Grafana
https://github.com/rh-jmc-team/container-jfr-web/pull/118#issuecomment-674942730

andrewazores:
A previous commit refactored the ApiService in a way that caused all
upload requests to use the HTTP API path as if they were all active
recordings. This re-adds a differentiation in the service implementation
between archived and active recordings to reflect the different HTTP API
paths for uploading either to a jfr-datasource.

Fixes #119

### DIFF
--- a/src/app/RecordingList/ActiveRecordingsList.tsx
+++ b/src/app/RecordingList/ActiveRecordingsList.tsx
@@ -211,7 +211,7 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
           <DataListItemCells
             dataListCells={listColumns}
           />
-          <RecordingActions index={props.index} recording={props.recording} />
+          <RecordingActions index={props.index} recording={props.recording} uploadFn={() => context.api.uploadActiveRecordingToGrafana(props.recording.name)} />
         </DataListItemRow>
         <DataListContent
           aria-label="Content Details"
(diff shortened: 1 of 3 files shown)


vic-ma:
I think there needs to be an error message for when the CJFR handler returns a failure response. For example, if I delete the archived recording from a terminal, or if I stop the jfr-datasource process, and then try to upload from the web client without refreshing first, then I will get a message saying the upload started, but no error message saying it didn't work.

andrewazores:
I did:

1. Create a recording using the web UI
2. Delete the recording using curl
3. Attempt to upload the recording using the web UI

and got a 404 error notification, which seems reasonable to me. Because of #119 I haven't tested it with archived recordings, but it should behave the same once #120 is merged. Do you not get the 404 error notification?

vic-ma:
I get the 404 when uploading active recordings, but not when uploading archived recordings. I was actually just manually deleting them with `rm`, but I tested now and I get the same thing when using the CJFR handler with curl.

It looks like `uploadActiveRecordingToGrafana` and `uploadArchivedRecordingToGrafana` are the same except [these lines](https://github.com/rh-jmc-team/container-jfr-web/blob/e54fe293f87a9c0225c210dbd06495596597a0f5/src/app/Shared/Services/Api.service.tsx#L218-L222) in the active one, which seems like is the error handling that is missing in the archived one?

PR #125, opened 2020-08-19
Interface for advanced recording options 

### LINKED ISSUE #103: View for customizing recording options
There should be a way for users to customize the recording options (max age, max size, to disk, etc. - see `RecordingOptionsCustomizer` in container-jfr). This is kept as internal state in container-jfr, but it may (or may not) make sense to implement this in a way that it appears to be associated with each individual new recording as it is created.

Alexjsenn:
interface
Fixes #103

### DIFF
--- a/src/app/CreateRecording/CreateRecording.tsx
+++ b/src/app/CreateRecording/CreateRecording.tsx
@@ -68,7 +68,7 @@ const Comp: React.FunctionComponent< RouteComponentProps<{}, StaticContext, Crea
 
   const handleCreateRecording = (recordingAttributes: RecordingAttributes): void => {
     addSubscription(
-      context.api.createRecording({ recordingAttributes })
+      context.api.createRecording(recordingAttributes)
       .pipe(first())
       .subscribe(success => {
         if (success) {
--- a/src/app/CreateRecording/CreateRecording.tsx
+++ b/src/app/CreateRecording/CreateRecording.tsx
@@ -65,10 +65,10 @@ const Comp: React.FunctionComponent< RouteComponentProps<{}, StaticContext, Crea
 
   const [activeTab, setActiveTab] = React.useState(0);
 
-  const handleCreateRecording = (recordingName: string, events: string, duration?: number, 
-          toDisk?: boolean, maxAge?: number, maxSize?: number): void => {
+  const handleCreateRecording = (recordingName: string, events: string, toDisk: boolean, 
+    maxAge: number, maxSize: number, duration?: number): void => {
     addSubscription(
-      context.api.createRecording({ recordingName, events, duration, toDisk, maxAge, maxSize })
+      context.api.createRecording({ recordingName, events, toDisk, maxAge, maxSize, duration })
       .pipe(first())
       .subscribe(success => {
         if (success) {
--- a/src/app/CreateRecording/CreateRecording.tsx
+++ b/src/app/CreateRecording/CreateRecording.tsx
@@ -65,9 +65,10 @@ const Comp: React.FunctionComponent< RouteComponentProps<{}, StaticContext, Crea
 
   const [activeTab, setActiveTab] = React.useState(0);
 
-  const handleCreateRecording = (recordingName: string, events: string, duration?: number): void => {
+  const handleCreateRecording = (recordingName: string, events: string, duration?: number, 
+          toDisk?: boolean, maxAge?: number, maxSize?: number): void => {
     addSubscription(
-      context.api.createRecording({ recordingName, events, duration })
+      context.api.createRecording({ recordingName, events, duration, toDisk, maxAge, maxSize })
       .pipe(first())
       .subscribe(success => {
         if (success) {
--- a/src/app/CreateRecording/CreateRecording.tsx
+++ b/src/app/CreateRecording/CreateRecording.tsx
@@ -36,7 +36,7 @@
  * SOFTWARE.
  */
 import * as React from 'react';
-import { RecordingOptions } from '@app/Shared/Services/Api.service';
+import { RecordingAttributes } from '@app/Shared/Services/Api.service';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { TargetView } from '@app/TargetView/TargetView';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
@@ -66,10 +66,9 @@ const Comp: React.FunctionComponent< RouteComponentProps<{}, StaticContext, Crea
 
   const [activeTab, setActiveTab] = React.useState(0);
 
-  const handleCreateRecording = (recordingName: string, events: string, options: RecordingOptions,
-     duration?: number): void => {
+  const handleCreateRecording = (recordingAttributes: RecordingAttributes): void => {
     addSubscription(
-      context.api.createRecording({ recordingName, events, options, duration })
+      context.api.createRecording({ recordingAttributes })
       .pipe(first())
       .subscribe(success => {
         if (success) {
--- a/src/app/CreateRecording/CreateRecording.tsx
+++ b/src/app/CreateRecording/CreateRecording.tsx
@@ -36,6 +36,7 @@
  * SOFTWARE.
  */
 import * as React from 'react';
+import { RecordingOptions } from '@app/Shared/Services/Api.service';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { TargetView } from '@app/TargetView/TargetView';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
@@ -65,10 +66,10 @@ const Comp: React.FunctionComponent< RouteComponentProps<{}, StaticContext, Crea
 
   const [activeTab, setActiveTab] = React.useState(0);
 
-  const handleCreateRecording = (recordingName: string, events: string, toDisk: boolean, 
-    maxAge: number, maxSize: number, duration?: number): void => {
+  const handleCreateRecording = (recordingName: string, events: string, options: RecordingOptions,
+     duration?: number): void => {
     addSubscription(
-      context.api.createRecording({ recordingName, events, toDisk, maxAge, maxSize, duration })
+      context.api.createRecording({ recordingName, events, options, duration })
       .pipe(first())
       .subscribe(success => {
         if (success) {
--- a/src/app/CreateRecording/CustomRecordingForm.tsx
+++ b/src/app/CreateRecording/CustomRecordingForm.tsx
@@ -127,7 +127,7 @@ export const CustomRecordingForm = (props) => {
     setToDisk(evt.target.checked);
   };
 
-  const setRecordingOptions = (options) => {
+  const setRecordingOptions = (options: RecordingOptions) => {
     // toDisk is not set, and defaults to true because of container-jfr issue #263
     setMaxAge(options.maxAge);
     setMaxAgeUnits(1);
@@ -310,7 +310,7 @@ export const CustomRecordingForm = (props) => {
                 >
                   <FormSelectOption key="1" value="1" label="Seconds" />
                   <FormSelectOption key="2" value={60} label="Minutes" />
-                  <FormSelectOption key="3" value={60*60}label="Hours" />
+                  <FormSelectOption key="3" value={60*60} label="Hours" />
                 </FormSelect>
               </SplitItem>
             </Split>
@@ -335,8 +335,8 @@ export const CustomRecordingForm = (props) => {
                     aria-label="Max size units input"
                   > 
                     <FormSelectOption key="1" value="1" label="B" />
-                    <FormSelectOption key="2" value={1000} label="KB" />
-                    <FormSelectOption key="3" value={1000*1000} label="MB" />
+                    <FormSelectOption key="2" value={1024} label="KB" />
+                    <FormSelectOption key="3" value={1024*1024} label="MB" />
                   </FormSelect>
                 </SplitItem>
             </Split>
--- a/src/app/CreateRecording/CustomRecordingForm.tsx
+++ b/src/app/CreateRecording/CustomRecordingForm.tsx
@@ -46,7 +46,7 @@ import { EventTemplate } from './CreateRecording';
 import { RecordingOptions } from '@app/Shared/Services/Api.service';
 
 export interface CustomRecordingFormProps {
-  onSubmit: (recordingName: string, events: string, duration?: number) => void;
+  onSubmit: (recordingName: string, events: string, options: RecordingOptions, duration?: number) => void;
 }
 
 export const RecordingNamePattern = /^[\w_]+$/;
@@ -148,7 +148,12 @@ export const CustomRecordingForm = (props) => {
       notifications.warning('Invalid form data', message);
       return;
     }
-    props.onSubmit(recordingName, getEventString(), toDisk, maxAge * maxAgeUnits, maxSize * maxSizeUnits, continuous ? undefined : duration * durationUnit);
+    const options: RecordingOptions = {
+      toDisk: toDisk,
+      maxAge: maxAge * maxAgeUnits,
+      maxSize: maxSize * maxSizeUnits 
+    }
+    props.onSubmit(recordingName, getEventString(), options, continuous ? undefined : duration * durationUnit);
   };
 
   React.useEffect(() => {
@@ -289,7 +294,11 @@ export const CustomRecordingForm = (props) => {
           <Text component={TextVariants.small}>
             A value of 0 for maximum age or size means unbounded.
           </Text>
-          <FormGroup label="Maximum age" fieldId="maxAge">
+          <FormGroup 
+            label="Maximum age" 
+            fieldId="maxAge"
+            helperText="The maximum age of recording data stored in the JFR repository buffer"
+          >
             <Split hasGutter={true}>
               <SplitItem isFilled>
                 <TextInput
@@ -315,7 +324,11 @@ export const CustomRecordingForm = (props) => {
               </SplitItem>
             </Split>
           </FormGroup>
-          <FormGroup label="Maximum size" fieldId="maxSize">
+          <FormGroup 
+            label="Maximum size" 
+            fieldId="maxSize"
+            helperText="The maximum size of recording data stored in the JFR repository buffer"
+          >
             <Split hasGutter={true}>
               <SplitItem isFilled>
                   <TextInput
--- a/src/app/CreateRecording/CustomRecordingForm.tsx
+++ b/src/app/CreateRecording/CustomRecordingForm.tsx
@@ -355,15 +355,18 @@ export const CustomRecordingForm = (props) => {
                     aria-label="Max size units input"
                   > 
                     <FormSelectOption key="1" value="1" label="B" />
-                    <FormSelectOption key="2" value={1024} label="KB" />
-                    <FormSelectOption key="3" value={1024*1024} label="MB" />
+                    <FormSelectOption key="2" value={1024} label="KiB" />
+                    <FormSelectOption key="3" value={1024*1024} label="MiB" />
                   </FormSelect>
                 </SplitItem>
             </Split>
           </FormGroup>
-          <FormGroup fieldId="toDisk">
+          <FormGroup 
+            fieldId="To Disk"
+            helperText="Write contents of buffer onto disk. If disabled, the buffer acts as circular buffer only keeping the most recent recording information"
+          >
             <Checkbox 
-              label="toDisk" 
+              label="To Disk" 
               id="toDisk-checkbox"
               isChecked={toDisk}
               onChange={handleToDiskChange} />
--- a/src/app/CreateRecording/CustomRecordingForm.tsx
+++ b/src/app/CreateRecording/CustomRecordingForm.tsx
@@ -43,10 +43,10 @@ import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import { useHistory } from 'react-router-dom';
 import { concatMap } from 'rxjs/operators';
 import { EventTemplate } from './CreateRecording';
-import { RecordingOptions } from '@app/Shared/Services/Api.service';
+import { RecordingOptions, RecordingAttributes, Recording } from '@app/Shared/Services/Api.service';
 
 export interface CustomRecordingFormProps {
-  onSubmit: (recordingName: string, events: string, options: RecordingOptions, duration?: number) => void;
+  onSubmit: (recordingAttributes: RecordingAttributes) => void;
 }
 
 export const RecordingNamePattern = /^[\w_]+$/;
@@ -153,7 +153,13 @@ export const CustomRecordingForm = (props) => {
       maxAge: maxAge * maxAgeUnits,
       maxSize: maxSize * maxSizeUnits 
     }
-    props.onSubmit(recordingName, getEventString(), options, continuous ? undefined : duration * durationUnit);
+    const recordingAttributes: RecordingAttributes = {
+      name: recordingName,
+      events: getEventString(),
+      duration: continuous ? undefined : duration * durationUnit,
+      options: options
+    }
+    props.onSubmit(recordingAttributes);
   };
 
   React.useEffect(() => {
@@ -208,6 +214,7 @@ export const CustomRecordingForm = (props) => {
         label="Duration"
         isRequired
         fieldId="recording-duration"
+        helperText="Time before the recording is automatically stopped"
       >
         <Checkbox
           label="Continuous"
--- a/src/app/CreateRecording/CustomRecordingForm.tsx
+++ b/src/app/CreateRecording/CustomRecordingForm.tsx
@@ -38,7 +38,7 @@
 import * as React from 'react';
 import { NotificationsContext } from '@app/Notifications/Notifications';
 import { ServiceContext } from '@app/Shared/Services/Services';
-import { ActionGroup, Button, Checkbox, Form, FormGroup, FormSelect, FormSelectOption, FormSelectOptionGroup, Split, SplitItem, Text, TextArea, TextInput, TextVariants, Tooltip, TooltipPosition, ValidatedOptions, ExpandableSection } from '@patternfly/react-core';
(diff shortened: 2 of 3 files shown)


andrewazores:
Cool, the interface prototyping looks great to me so far. Do you have any other significant changes planned, other than hooking it up to the backend?

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -38,7 +38,7 @@
 import * as React from 'react';
 import { NotificationsContext } from '@app/Notifications/Notifications';
 import { ServiceContext } from '@app/Shared/Services/Services';
-import { ActionGroup, Button, Checkbox, Form, FormGroup, FormSelect, FormSelectOption, FormSelectOptionGroup, Split, SplitItem, Text, TextArea, TextInput, TextVariants, Tooltip, TooltipPosition, ValidatedOptions } from '@patternfly/react-core';
+import { ActionGroup, Button, Checkbox, Form, FormGroup, FormSelect, FormSelectOption, FormSelectOptionGroup, Split, SplitItem, Text, TextArea, TextInput, TextVariants, Tooltip, TooltipPosition, ValidatedOptions, ExpandableSection } from '@patternfly/react-core';
```
andrewazores, line 41: Move the new import so that they're alphabetical please

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -67,6 +67,10 @@ export const CustomRecordingForm = (props) => {
   const [templateType] = React.useState(props.templateType || props?.location?.state?.templateType || '');
   const [eventSpecifiers, setEventSpecifiers] = React.useState(props?.eventSpecifiers?.join(' ') || '');
   const [eventsValid, setEventsValid] = React.useState((!!props.template || !!props?.location?.state?.template) ? ValidatedOptions.success : ValidatedOptions.default);
+  const [maxAge, setMaxAge] = React.useState(60);
+  const [maxAgeUnits, setMaxAgeUnits] = React.useState(1);
+  const [maxSize, setMaxSize] = React.useState(512);
+  const [maxSizeUnits, setMaxSizeUnits] = React.useState(1);
```
andrewazores, line 73: It doesn't matter yet, but there's a missing `useState` here for `toDisk`

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -101,6 +105,22 @@ export const CustomRecordingForm = (props) => {
...
+    setMaxAgeUnits(Number(evt));
+  }
+
+  const handleMaxSizeChange = (evt) => {
+    setMaxSize(Number(evt));
+  }
+
+  const handleMaxSizeUnitChange = (evt) => {
+    setMaxSizeUnits(Number(evt));
+  }
```
andrewazores, line 122: And there will need to be a handler here for when `toDisk` is (un)checked, too

Alexjsenn:
I was wondering if there should be a button to reset the values to the current default, in case someone changes them but then wants to know the defaults?

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -243,6 +263,65 @@ export const CustomRecordingForm = (props) => {
...
+        <Form>
+          <FormGroup label="Maximum age" fieldId="maxAge">
+            <Split hasGutter={true}>
+              <SplitItem isFilled>
+                <TextInput
+                  value={maxAge}
+                  isRequired
+                  type="number"
+                  id="maxAgeDuration"
+                  aria-describedby="maxAgeDuration-helper"
```
andrewazores, line 276: I don't recall off the top of my head what needs to be added for these `describedby` accessibility attributes, but there should be a text label or something with the id `maxAgeDuration-helper` I think that this "links" to? Same for all the other `aria-describedby` attributes on other elements.

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -243,6 +263,65 @@ export const CustomRecordingForm = (props) => {
...
+              </SplitItem>
+              <SplitItem>
+                <FormSelect
+                  value={maxAgeUnits}
+                  onChange={handleMaxAgeUnitChange}
+                  aria-label="maxAgeDuration Units Input"
+                >
+                  <FormSelectOption key="1" value="1" label="Seconds" />
+                  <FormSelectOption key="2" value={60} label="Minutes" />
+                  <FormSelectOption key="3" value={60*60}label="Hours" />
```
andrewazores, line 289: Missing a space between the `}` and `label`

andrewazores:
> I was wondering if there should be a button to reset the values to the current default, in case someone changes them but then wants to know the defaults?

That's a good question and idea. I don't think I ever implemented anything like that on the backend. If the user has not specified these options for ContainerJFR to track in its global state, then recordings are created with whatever value is default for the target JVM, which of course may vary with the targets. Once a client sets option values on the ContainerJFR backend then those values are used instead and override the targets' defaults.

So, this suggests that we should have backend HTTP APIs for:

1. Getting the values of these options (defaults) for a specific target
2. Getting the values of these options in ContainerJFR's global state
3. Setting option values in ContainerJFR's global state
4. Including option values in the existing `POST` request that creates recordings, so that recordings can be created with options that differ from the defaults either in ContainerJFR or the target, without modifying any state

src/app/CreateRecording/CustomRecordingForm.tsx
andrewazores: Also I'm sure these are missing pretty much everywhere else right now...

Alexjsenn: I changed the aria-describedby to aria-labels because it makes more sense. I think that other parts of the code have aria-describedby that don't link anywhere, so Im not sure If I should change them as well?

andrewazores: Nice. That sounds like a great enhancement, but certainly as a separate Issue/PR.

Alexjsenn:
Ok, so for now the defaults are read into the initial values for the recording options, then they are sent with the POST request when creating a new recording. For now there is no way to modify the defaults.

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -243,6 +284,72 @@ export const CustomRecordingForm = (props) => {
...
+              </SplitItem>
+              <SplitItem>
+                <FormSelect
+                  value={maxAgeUnits}
+                  onChange={handleMaxAgeUnitChange}
+                  aria-label="Max Age units Input"
+                >
+                  <FormSelectOption key="1" value="1" label="Seconds" />
+                  <FormSelectOption key="2" value={60} label="Minutes" />
+                  <FormSelectOption key="3" value={60*60}label="Hours" />
```
andrewazores, line 313: There's a missing space here between `}` and `label`

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -101,6 +107,34 @@ export const CustomRecordingForm = (props) => {
...
+
+  const handleMaxSizeUnitChange = (evt) => {
+    setMaxSizeUnits(Number(evt));
+  };
+
+  const handleToDiskChange = (checked, evt) => {
+    setToDisk(evt.target.checked);
+  };
+
+  const setRecordingOptions = (options) => {
```
andrewazores, line 130: I think adding a type annotation to `options` here would make sense and improve readability

src/app/CreateRecording/CreateRecording.tsx
```diff
@@ -65,9 +65,10 @@ const Comp: React.FunctionComponent< RouteComponentProps<{}, StaticContext, Crea
 
   const [activeTab, setActiveTab] = React.useState(0);
 
-  const handleCreateRecording = (recordingName: string, events: string, duration?: number): void => {
+  const handleCreateRecording = (recordingName: string, events: string, toDisk: boolean,
```
andrewazores, line 68: This function signature is getting pretty complicated with the number of arguments required in a specific order. Do you think it would make sense to refactor this to accept an object the same way that `api.createRecording` does, and simply pass that object through?

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -243,6 +284,72 @@ export const CustomRecordingForm = (props) => {
...
+                </SplitItem>
+                <SplitItem>
+                  <FormSelect
+                    value={maxSizeUnits}
+                    onChange={handleMaxSizeUnitChange}
+                    aria-label="Max size units input"
+                  > 
+                    <FormSelectOption key="1" value="1" label="B" />
+                    <FormSelectOption key="2" value={1000} label="KB" />
+                    <FormSelectOption key="3" value={1000*1000} label="MB" />
```
andrewazores, line 339: 1000-byte kilobytes...

Please make this `1024 / KiB` and `1024*1024 / MiB`.

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -243,6 +284,72 @@ export const CustomRecordingForm = (props) => {
           </SplitItem>
         </Split>
       </FormGroup>
+      <ExpandableSection toggleTextExpanded="Hide advanced options" toggleTextCollapsed="Show advanced options">
+        <Form>
+          <Text component={TextVariants.small}>
+            A value of 0 for maximum age or size means unbounded.
```
andrewazores, line 290: I definitely think we need some description/explanation for users about the difference between a recording's duration and the maximum age option - at face value they sound like they would do the same thing.

andrewazores:
Other than the couple of minor notes above the patch looks clean, and I tested it out and it works exactly as expected. Nice work overall.

src/app/CreateRecording/CreateRecording.tsx
Alexjsenn: Could I pass in a RecordingOptions object and keep the other values the same, or should I create a new object with everything needed?

Alexjsenn: Could I pass in a RecordingOptions object and keep the other values the same, or should I create a new object with everything needed?

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -38,11 +38,12 @@
...
-import { ActionGroup, Button, Checkbox, Form, FormGroup, FormSelect, FormSelectOption, FormSelectOptionGroup, Split, SplitItem, Text, TextArea, TextInput, TextVariants, Tooltip, TooltipPosition, ValidatedOptions } from '@patternfly/react-core';
+import { ActionGroup, Button, Checkbox, ExpandableSection, Form, FormGroup, FormSelect, FormSelectOption, FormSelectOptionGroup, Split, SplitItem, Text, TextArea, TextInput, TextVariants, Tooltip, TooltipPosition, ValidatedOptions } from '@patternfly/react-core';
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import { useHistory } from 'react-router-dom';
 import { concatMap } from 'rxjs/operators';
 import { EventTemplate } from './CreateRecording';
+import { RecordingOptions } from '@app/Shared/Services/Api.service';
 
 export interface CustomRecordingFormProps {
   onSubmit: (recordingName: string, events: string, duration?: number) => void;
```
Alexjsenn, line 49: I was also wondering if I should modify this interface, I don't know how its used anywhere, but its outdated nonetheless

src/app/CreateRecording/CreateRecording.tsx
andrewazores: IMO it makes slightly more sense to put all of the attributes into a single object, but I can see the reasoning behind putting the recording options into an object and keeping the name/duration/events split out. Up to you.

src/app/CreateRecording/CustomRecordingForm.tsx
andrewazores: Ah, it should be used to annotate the type of the props expected/received by the CustomRecordingForm component. The "parent" CreateRecording component is an example of this being used properly. Since you're already here, please do go ahead and correct the component definition to specify its prop types with this interface, and update the interface to reflect the actual requirements.

The SnapshotRecordingForm component also has the same problem.

src/app/CreateRecording/CustomRecordingForm.tsx
Alexjsenn: I added helper text, but I'm not sure if its explains it correctly. Let me know if there's a better way to phrase it.

src/app/CreateRecording/CustomRecordingForm.tsx
Alexjsenn: I fixed the interface with the proper requirements, Im still a bit confused as to how this props interface works though so I don't know if other things should be added?

src/app/CreateRecording/CreateRecording.tsx
Alexjsenn: I think for now its easier to use the recordingOptions object because we already have the interface for it, the problem is there is already an interface called Recording, we would have to create another one called newRecording or something to store everything together. I can change it to that if you want though!

src/app/CreateRecording/CustomRecordingForm.tsx
andrewazores: I think the explanation is phrased fine. Maybe now we just need some description of what exactly the "duration" of a recording is then - ie explain that maxAge is the maximum age of the data kept in the buffer, and duration specifies when the recording will be automatically stopped.

src/app/CreateRecording/CreateRecording.tsx
andrewazores: It could be named `RecordingAttributes`, or something. Don't name it `NewRecording`. Having another type with the same or similar name in a different context shouldn't be a strong consideration when deciding whether or how to structure or refactor things.

src/app/CreateRecording/CustomRecordingForm.tsx
andrewazores: The interface is used to specify the types that the component receives in its props. This is just used for compile-time type checking.

https://reactjs.org/docs/components-and-props.html

https://github.com/rh-jmc-team/container-jfr-web/blob/8cd7842af44ac78eba8fdf7b18e84c36e1ef98c6/src/app/CreateRecording/CreateRecording.tsx#L49

https://github.com/rh-jmc-team/container-jfr-web/blob/8cd7842af44ac78eba8fdf7b18e84c36e1ef98c6/src/app/CreateRecording/CreateRecording.tsx#L61

src/app/CreateRecording/CustomRecordingForm.tsx
Alexjsenn: Ok, I added helper text for duration as well.

src/app/CreateRecording/CreateRecording.tsx
Alexjsenn: Ok, yeah, its cleaner with a single object.

Alexjsenn:
I just noticed an issue when toDisk is turned off. When looking at the recording list, and clicking on the dropdown, it says `java.io.IOException: No recording data available`. The drop down should probably not appear for such recordings, or a more informative message could be displayed.

andrewazores:
If you download the recording, is it empty/0 bytes? That's what I would expect if you see `No recording data available`. If that's the case then I don't think that just removing the report dropdown section is an appropriate fix - this is a scenario where the recording is totally broken because it has no contents, so silently removing the expandable report section is just going to confuse users and make it very difficult to diagnose the root problem.

src/app/CreateRecording/CreateRecording.tsx
```diff
@@ -65,9 +66,9 @@ const Comp: React.FunctionComponent< RouteComponentProps<{}, StaticContext, Crea
 
   const [activeTab, setActiveTab] = React.useState(0);
 
-  const handleCreateRecording = (recordingName: string, events: string, duration?: number): void => {
+  const handleCreateRecording = (recordingAttributes: RecordingAttributes): void => {
     addSubscription(
-      context.api.createRecording({ recordingName, events, duration })
+      context.api.createRecording({ recordingAttributes })
```
andrewazores, line 71: `func({ key })` is syntactic sugar for `func({ key: key })` - that is, the function is being called with an object as its sole parameter, and that object has a single key. This can/should be simplified.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -90,15 +90,18 @@ export class ApiService {
     );
   }
 
-  createRecording(
-    { recordingName, events, duration  }: { recordingName: string; events: string; duration?: number }
+  createRecording({ recordingAttributes }: { recordingAttributes: RecordingAttributes }
```
andrewazores, line 93: This is the place where the function definition should be simplified. This function should accept simply a `RecordingAttributes` parameter, not an object with a single `recordingAttributes` key of type `RecordingAttributes`.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -411,3 +414,16 @@ export interface EventTemplate {
...
+  toDisk: boolean;
+  maxSize: number;
+  maxAge: number;
+}
+
+export interface RecordingAttributes {
+  name: string;
+  events: string;
+  duration?: number;
+  options: RecordingOptions;
```
andrewazores, line 428: If `duration` is marked as optional then I think it also makes sense to mark `options` as optional. And, within the options, each of its attributes should probably be optional as well.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -90,15 +90,18 @@ export class ApiService {
...
       const form = new window.FormData();
-      form.append('recordingName', recordingName);
-      form.append('events', events);
-      if (!!duration && duration > 0) {
-        form.append('duration', String(duration));
+      form.append('recordingName', recordingAttributes.name);
+      form.append('events', recordingAttributes.events);
+      form.append('toDisk', String(recordingAttributes.options.toDisk));
+      form.append('maxAge', String(recordingAttributes.options.maxAge));
+      form.append('maxSize', String(recordingAttributes.options.maxSize));
```
andrewazores, line 100: For validation purposes, `toDisk/maxAge/maxSize` should perform a check similar to what is used for `duration` below, ensuring that the values are defined before they are included in the form contents. Although the component that provides these values is currently written to first retrieve default values from the backend and then pass them through the attributes configuration object when the data flow reaches here, this function should still validate the incoming data for robustness.

Alexjsenn:
> If you download the recording, is it empty/0 bytes? That's what I would expect if you see `No recording data available`. If that's the case then I don't think that just removing the report dropdown section is an appropriate fix - this is a scenario where the recording is totally broken because it has no contents, so silently removing the expandable report section is just going to confuse users and make it very difficult to diagnose the root problem.

Yes, when you download the report it gives `<html> <head></head> <body>  <div>java.io.IOException: No recording data available  </div> </body></html>`. Im not sure how this should be dealt with, the dropdown could have a message explaing that because the recording isn't saved to disk the report cannot be shown?

src/app/Shared/Services/Api.service.tsx
Alexjsenn: I can mark them optional, however there will always be a value for them because unlike duration, there is no scenario where the recordingOptions isn't sent.

andrewazores:
> > If you download the recording, is it empty/0 bytes? That's what I would expect if you see `No recording data available`. If that's the case then I don't think that just removing the report dropdown section is an appropriate fix - this is a scenario where the recording is totally broken because it has no contents, so silently removing the expandable report section is just going to confuse users and make it very difficult to diagnose the root problem.
> 
> Yes, when you download the report it gives `<html> <head></head> <body> <div>java.io.IOException: No recording data available </div> </body></html>`. Im not sure how this should be dealt with, the dropdown could have a message explaing that because the recording isn't saved to disk the report cannot be shown?

The problem you describe isn't that the report cannot be shown - it's that the recording itself is broken and contains no data. That's why I'm saying the fix is not to remove or modify the display of the report.

src/app/Shared/Services/Api.service.tsx
andrewazores: That may be the case *right now*, but you can't make that claim as an incontrovertible fact for all potential future modifications to the code...

The type definition should model not only how the type is currently used by the implementation, but model the general specifications and requirements. The options are not required to be sent when performing requests to the backend, so the type should mark the options as, well, optional.

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -243,6 +296,80 @@ export const CustomRecordingForm = (props) => {
...
+                </SplitItem>
+                <SplitItem>
+                  <FormSelect
+                    value={maxSizeUnits}
+                    onChange={handleMaxSizeUnitChange}
+                    aria-label="Max size units input"
+                  > 
+                    <FormSelectOption key="1" value="1" label="B" />
+                    <FormSelectOption key="2" value={1024} label="KB" />
+                    <FormSelectOption key="3" value={1024*1024} label="MB" />
```
andrewazores, line 359: `KiB`/`MiB` please

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -243,6 +296,80 @@ export const CustomRecordingForm = (props) => {
...
+                  </FormSelect>
+                </SplitItem>
+            </Split>
+          </FormGroup>
+          <FormGroup fieldId="toDisk">
+            <Checkbox
```
andrewazores, line 365: I think `toDisk` should be written as "To Disk" and have a snippet of descriptive text as well, like the numeric options above it have.

andrewazores:
I think the `toDisk` form handling is somehow broken. I set up a recording as shown in the screenshot, but when I check the list of recordings I see that the recording's `maxAge/maxSize` attributes are as expected but the `toDisk` is still `true`.

Screenshot_2020-09-16 Create Recording

```
$ curl -vk -H "X-JMX-Authorization: Basic $(echo -n admin:adminpass123 | base64)" "https://0.0.0.0:8181/api/v1/targets/$(echo -n localhost:9093 | jq -sRr @uri)/recordings"
*   Trying 0.0.0.0:8181...
* TCP_NODELAY set
* Connected to 0.0.0.0 (127.0.0.1) port 8181 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep 16 18:22:10 2020 GMT
*  expire date: Mar 15 18:22:10 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> GET /api/v1/targets/localhost%3A9093/recordings HTTP/1.1
> Host: 0.0.0.0:8181
> User-Agent: curl/7.66.0
> Accept: */*
> X-JMX-Authorization: Basic YWRtaW46YWRtaW5wYXNzMTIz
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-length: 410
< 
* Connection #0 to host 0.0.0.0 left intact
[{"downloadUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9093%2Fjmxrmi/recordings/test","reportUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9093%2Fjmxrmi/reports/test","id":5,"name":"test","state":"RUNNING","startTime":1600280876999,"duration":0,"continuous":true,"toDisk":true,"maxSize":12582912,"maxAge":60000}]
```

andrewazores:
I also tested it with various other settings combinations for duration/maxAge/maxSize, but the recording always seems to have `toDisk=true` when I have the checkbox unticked.

Alexjsenn:
> I also tested it with various other settings combinations for duration/maxAge/maxSize, but the recording always seems to have `toDisk=true` when I have the checkbox unticked.

Yeah, I made a mistake with the validation in the createRecording function, it now works.

andrewazores:
I just stumbled across these documentation notes while researching the empty recording problem above:

JDK15: https://docs.oracle.com/en/java/javase/15/docs/specs/man/java.html
JDK11: https://docs.oracle.com/en/java/javase/11/tools/java.html#GUID-3B1CE181-CD30-4178-9602-230B800D4FAE
JDK8: https://docs.oracle.com/javase/8/docs/technotes/tools/unix/java.html#CBBFHAJA

JDK8 has some different defaults and a few other options that have since been deprecated, but I think the behaviour should be the same for all JFR-enabled JDKs when the settings are explicitly set.

These imply that the `maxAge/maxSize` UI elements should be disabled and their values ignored when `toDisk=false`.

andrewazores:
Also, some of the form attribute description texts now end in periods and others do not. I don't personally mind either way, but they should be consistent. Please check if there is any Patternfly guidance on which way is preferred.

Alexjsenn:
> Also, some of the form attribute description texts now end in periods and others do not. I don't personally mind either way, but they should be consistent. Please check if there is any Patternfly guidance on which way is preferred.

Looking at the examples from PatternFly's website the helper text does not seem to end in periods, so I removed them.

andrewazores (approved):
The empty recording issue when `toDisk=false` is a separate problem - either a backend problem or simply us as users misunderstanding the JFR options. The UI design here and its implementation look good.

PR #131, opened 2020-09-18
Display recording options for current recording list

### LINKED ISSUE #127: [design] Recording's options (max age, max size, toDisk) should be discoverable
The API responses from `/api/v1/targets/:targetId/recordings` includes details on each recording about their maxAge/maxSize/toDisk option settings, but there is no UI element that exposes this information to the user. Since #103 adds a way for users to customize these settings and change their values from the defaults when a recording is created, there should also be a corresponding way for the user to see the values of these settings for each recording.

Alexjsenn:
Fixes #127

### DIFF
--- a/src/app/RecordingList/ActiveRecordingsList.tsx
+++ b/src/app/RecordingList/ActiveRecordingsList.tsx
@@ -219,6 +219,16 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
           id={`active-ex-expand-${props.index}`}
           isHidden={!isExpanded}
         >
+          <Text>Recording Options:</Text>
+          <Text>
+            toDisk = { String(props.recording.toDisk) } &emsp; 
+            maxAge = {props.recording.maxAge / 1000}s &emsp; 
+            maxSize = { props.recording.maxSize }B
+          </Text>
+          <br></br>
+          <hr></hr>
+          <br></br>
+          <Text>Automated Analysis:</Text>
           <ReportFrame recording={props.recording} width="100%" height="640" />
         </DataListContent>
       </DataListItem>


Alexjsenn:
@andrewazores Do you think it makes sense to change the recording report's appearance? Or is it unecessary?

andrewazores:
Change the appearance in what sense or for what purpose?

Alexjsenn:
Making the titles (at least) a bit smaller, so that they aren't so much larger than the rest of the text.

andrewazores:
I don't think it's necessary at this point, but if you don't find any other good work items to take up then that could be something to look at. That report transformation is done on the ContainerJFR backend, however, so it wouldn't be part of this PR.

Alexjsenn:
Ok, makes sense!

src/app/RecordingList/ActiveRecordingsList.tsx
```diff
@@ -219,6 +219,16 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
...
         >
+          <Text>Recording Options:</Text>
+          <Text>
+            toDisk = { String(props.recording.toDisk) } &emsp; 
+            maxAge = {props.recording.maxAge / 1000}s &emsp; 
+            maxSize = { props.recording.maxSize }B
+          </Text>
+          <br></br>
+          <hr></hr>
+          <br></br>
```
andrewazores, line 230: `hr` and `br` are empty elements in HTML, so these should simply be `<br>` and `<hr>` with no `</br>` closing tags.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br

Alexjsenn: Im trying to understand how the Text component works, and it seemes like it should work but for some reason it gives an error when I don't close hr and br.

andrewazores: Hmm. Maybe it's a quirk of TSX syntax, or the way that it's compiled. No problem then.

andrewazores (approved):

PR #136, opened 2020-09-28
Create loading state in recording list views

### LINKED ISSUE #133: Selecting a target that fails to connect leaves previous content in recording list view
When in recording list view, if a target is selected and it fails to connect (for example a 404 error is thrown), the recording list maintains its previous state. If for example a list of recordings of a different target were being viewed, after clicking on the new unreachable target, the target selector shows the unreachable target is selected but the list continues to display the first target's recordings. There should be a loading state that clears the view in preparation for a new incoming list (and remains empty if the target fails to connect).

Alexjsenn:
Fixes #133. Currently the loading spinner is visible while waiting for the user to input the JMX auth credentials. However, in the case of a connection failure (for example when cjfr does not have the cert file for the target in smoketest) the loading spinner just stays there forever. I think ideally an error message or icon should replace it, but doing so might require more changes to the Api.service.

### DIFF
--- a/src/app/RecordingList/RecordingsDataTable.tsx
+++ b/src/app/RecordingList/RecordingsDataTable.tsx
@@ -45,12 +45,12 @@ export interface RecordingsDataTableProps {
   listTitle: string;
   isHeaderChecked: boolean;
   isLoading: boolean;
-  errorMessage: string;
+  isConnectionError: boolean;
   onHeaderCheck: (checked: boolean) => void;
 }
 
 export const RecordingsDataTable: React.FunctionComponent<RecordingsDataTableProps> = (props) => {
-  if (props.errorMessage != '') {
+  if (props.isConnectionError) {
     return (<>
       <br/>
       <Bullseye>
--- a/src/app/RecordingList/RecordingsDataTable.tsx
+++ b/src/app/RecordingList/RecordingsDataTable.tsx
@@ -45,12 +45,12 @@ export interface RecordingsDataTableProps {
   listTitle: string;
   isHeaderChecked: boolean;
   isLoading: boolean;
-  isConnectionError: boolean;
+  errorMessage: String;
   onHeaderCheck: (checked: boolean) => void;
 }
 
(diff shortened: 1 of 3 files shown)


Alexjsenn:
Screenshot from 2020-09-28 13-58-24
This error message appears (following the PatternFly empty state guidelines) when the connection to a target fails (this example is running smoketest without the cert file for the demo app).

Alexjsenn:
This kind of empty state might also make sense in the Events page.

src/app/RecordingList/RecordingsDataTable.tsx
```diff
@@ -36,33 +36,55 @@
...
-  </>);
+  if (props.isConnectionError) {
+    return (<>
+      <br/>
+      <Bullseye>
+        <ExclamationCircleIcon size='md' color='Red'/>
+      </Bullseye>
+      <Bullseye>
+        <Text>
+          Unable to connect
```
andrewazores, line 61: I don't think such a specific error message should be used here - there could be plenty of other reasons why an error occurs, too. This should either be a generic message, or an exact message should be passed to the component.

src/app/RecordingList/RecordingsDataTable.tsx
```diff
@@ -36,33 +36,55 @@
...
+        </Text>
+      </Bullseye>
+    </>)
+  } else {
+    return props.isLoading? (<>
```
andrewazores, line 66: Since we already have this whole thing wrapped in an outer `if/else`, could you refactor this to remove the ternary statement?

ie we have

```
if (A) {
  return (<TemplateA/>);
} else {
  return B ? (<TemplateB/>) : (<TemplateC/>);
}
```

This seems like it would be easier to read like:

```
if (A) {
  return (<TemplateA/>);
} else if (B) {
  return (<TemplateB/>);
} else {
 return (<TemplateC/>);
}
```

src/app/RecordingList/RecordingsDataTable.tsx
Alexjsenn: Ok, I changed it to display the error from the failed request

andrewazores (approved):

PR #138, opened 2020-09-29
Prevent refresh when same target is reselected

### LINKED ISSUE #126: Re-selecting a target causes view refresh
Clicking the target selection dropdown and selecting a target, then repeating this for the same target, causes a refresh of the main view. This is noticeable when the main view is Recordings and a report frame is open. Re-selecting the same target should not cause views to be redrawn or network requests to be re-issued. If the main view needs a way to refresh its content then this should be implemented within the view itself.

Alexjsenn:
Fixes #126

### DIFF
--- a/src/app/TargetSelect/TargetSelect.tsx
+++ b/src/app/TargetSelect/TargetSelect.tsx
@@ -87,7 +87,10 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props) 
     if (isPlaceholder) {
       context.target.setTarget('');
     } else {
-      context.target.setTarget(selection.connectUrl);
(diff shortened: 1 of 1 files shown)


andrewazores:
This looks like it probably works, but I don't think it's quite the right solution. I don't think the subscription should be created and then immediately unsubscribed within that `onSelect` handler - this only works if the target value happens to be emitted synchronously, which is an implementation detail that currently holds but is not guaranteed and shouldn't be assumed.

The component already has a subscription to the `context.target.target()` (line 82), so the `selected` state of the component should be usable as a proxy for the current value, rather than re-subscribing.

Alexjsenn:
Yeah, you are absolutely right!

andrewazores (approved):

PR #139, opened 2020-10-02
Display alias for selected target in widget (refactor TargetService)

### LINKED ISSUE #134: Target selection widget should primarily display target alias, not URL
Screenshot_2020-09-23_17-57-30

When selected, the displayed label for the current target is its connection URL. This should probably be the target's alias instead, since that is intended to be more human-readable and useful at a glance.

Alexjsenn:
Fixes #134

### DIFF
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -370,7 +370,7 @@ export class ApiService {
       concatMap(parts => {
         const headers = parts[0];
         const target = parts[1];
-        if (!!target.connectUrl && this.target.hasCredentials(target.connectUrl)) {
+        if (!!target && !!target.connectUrl && this.target.hasCredentials(target.connectUrl)) {
           const credentials = this.target.getCredentials(target.connectUrl);
           if (credentials) {
             headers.set('X-JMX-Authorization', `Basic ${this.target.getCredentials(target.connectUrl)}`);
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -125,8 +125,8 @@ export class ApiService {
         }
       }
 
-      return this.target.target().pipe(concatMap(targetId =>
-        this.sendRequest(`targets/${encodeURIComponent(targetId)}/recordings`, {
+      return this.target.target().pipe(concatMap(target =>
+        this.sendRequest(`targets/${encodeURIComponent(target.connectUrl)}/recordings`, {
           method: 'POST',
           body: form,
         }).pipe(
@@ -141,8 +141,8 @@ export class ApiService {
   }
 
   createSnapshot(): Observable<boolean> {
-    return this.target.target().pipe(concatMap(targetId =>
-      this.sendRequest(`targets/${encodeURIComponent(targetId)}/snapshot`, {
+    return this.target.target().pipe(concatMap(target =>
+      this.sendRequest(`targets/${encodeURIComponent(target.connectUrl)}/snapshot`, {
         method: 'POST',
       }).pipe(
         tap(resp => {
@@ -157,9 +157,9 @@ export class ApiService {
   }
 
   archiveRecording(recordingName: string): Observable<boolean> {
-    return this.target.target().pipe(concatMap(targetId =>
+    return this.target.target().pipe(concatMap(target =>
       this.sendRequest(
-        `targets/${encodeURIComponent(targetId)}/recordings/${encodeURIComponent(recordingName)}`,
+        `targets/${encodeURIComponent(target.connectUrl)}/recordings/${encodeURIComponent(recordingName)}`,
         {
           method: 'PATCH',
           body: 'SAVE',
@@ -172,9 +172,9 @@ export class ApiService {
   }
 
   stopRecording(recordingName: string): Observable<boolean> {
-    return this.target.target().pipe(concatMap(targetId =>
+    return this.target.target().pipe(concatMap(target =>
       this.sendRequest(
-        `targets/${encodeURIComponent(targetId)}/recordings/${encodeURIComponent(recordingName)}`,
+        `targets/${encodeURIComponent(target.connectUrl)}/recordings/${encodeURIComponent(recordingName)}`,
         {
           method: 'PATCH',
           body: 'STOP',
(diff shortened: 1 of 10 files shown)


src/app/Shared/Services/Api.service.tsx
```diff
@@ -370,10 +370,10 @@ export class ApiService {
       concatMap(parts => {
         const headers = parts[0];
         const target = parts[1];
-        if (!!target && this.target.hasCredentials(target)) {
-          const credentials = this.target.getCredentials(target);
+        if (!!target.connectUrl && this.target.hasCredentials(target.connectUrl)) {
```
andrewazores, line 373: I think, to be safe, that this condition should still start with `!!target`, in addition to the new `!!target.connectUrl`. This will guard against a TypeError in case the `target` object itself is undefined/null.

src/app/Shared/Services/Services.tsx
```diff
@@ -42,6 +42,11 @@ import { ApiService } from './Api.service';
 import { CommandChannel } from './CommandChannel.service';
 import { ReportService } from './Report.service';
 
+export interface Target {
```
andrewazores, line 45: Should this be moved into the `Target.service.tsx`? This would be more akin to how the types are organized around `Api.service.tsx`.

src/app/Shared/Services/Target.service.tsx
```diff
@@ -36,19 +36,19 @@
  * SOFTWARE.
  */
 import { Observable, Subject, BehaviorSubject } from 'rxjs';
+import { Target } from './Services';
 
 class TargetService {
-
-  private readonly _target: Subject<string> = new BehaviorSubject('');
+  private readonly _target: Subject<Target> = new BehaviorSubject({} as Target);
```
andrewazores, line 42: I think it's worth defining some specific publicly-readable constant value as a `NO_TARGET`, which could be defined as the empty object `{}` or something else. But this way we have a consistent value to use to represent this case.

src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -46,14 +46,9 @@ export interface TargetSelectProps {
...
 
-interface Target {
-  connectUrl: string;
-  alias: string;
-}
-
 export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props) => {
   const context = React.useContext(ServiceContext);
-  const [selected, setSelected] = React.useState('');
+  const [selected, setSelected] = React.useState({connectUrl:'', alias:''} as Target);
```
andrewazores, line 51: And here is the case in point for a `NO_TARGET` constant - the service initializes with `{}` as its default state, but this component uses `{ connectUrl: '', alias: ''}` as its default state - these seem to be intended to mean the same thing, so they should be represented with the same (or at least equal in comparison) value.

src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -85,9 +80,9 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props)
 
   const onSelect = (evt, selection, isPlaceholder) => {
     if (isPlaceholder) {
-      context.target.setTarget('');
+      context.target.setTarget({connectUrl:'', alias:''} as Target);
```
andrewazores, line 83: This should also be refactored to use `NO_TARGET`, or at least some constant value defined within this class that is also used when initializing the state above.

src/app/TargetView/TargetView.tsx
```diff
@@ -53,7 +53,9 @@ export const TargetView: React.FunctionComponent<TargetViewProps> = (props) => {
   const [connected, setConnected] = React.useState(false);
 
   React.useLayoutEffect(() => {
-    const sub = context.target.target().pipe(map(t => !!t)).subscribe(setConnected);
+    const sub = context.target.target().subscribe(target => {
+      setConnected(!!target.connectUrl);
```
andrewazores, line 57: Same comment here re. `!!target` and field access safety.

andrewazores (changes requested):
Just needs a little bit of polishing, but it looks generally pretty good and works exactly as expected.

src/app/Shared/Services/Target.service.tsx
Alexjsenn: Would this be in Target.service.tsx and then have to imported anywhere its needed?

andrewazores: Yes, that sounds about right. It could also be a member of the `context.target` instance itself, but that doesn't seem quite right.

Alexjsenn:
I added the NO_TARGET, should it also be used in places where currently it checks for !!target?

andrewazores:
We can add some validation in the `Target.service` to ensure that it will only hold valid state, ie its `setTarget` should ensure that the parameter passed to it is non-null/undefined, and that it is either `NO_TARGET` or has both required fields and that both fields are non-null. If any of these conditions is not met then throw an error. This way anyone consuming `context.target.target()` knows that any future values received will either be `NO_TARGET` or a valid object, and we don't need to add complex validations on every site that consumes the API.

Alexjsenn:
When setting a target, right now I catch the error and simply set the target to NO_TARGET, Im not sure if it makes sense to do anything else?

src/app/Shared/Services/Target.service.tsx
```diff
@@ -51,7 +51,11 @@ class TargetService {
   private readonly _credentials: Map<string, string> = new window.Map();
 
   setTarget(target: Target): void {
-    this._target.next(target);
+    if (target === NO_TARGET || (target.alias != '' && target.connectUrl != '')) {
```
andrewazores, line 54: If I do `setTarget({ alias: 'something', connectUrl: undefined} as Target)`, does this throw an error or accept the state change? What should it do?

Alexjsenn: You are right, I changed it to !!target.alias, it seems that checks for null, undefined and empty string

andrewazores: Pretty much.

https://developer.mozilla.org/en-US/docs/Glossary/Falsy

https://medium.com/better-programming/javascript-bang-bang-i-shot-you-down-use-of-double-bangs-in-javascript-7c9d94446054

src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -85,9 +81,15 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props)
...
     if (isPlaceholder) {
-      context.target.setTarget('');
+      context.target.setTarget(NO_TARGET);
     } else {
-      if (selection.connectUrl != selected) context.target.setTarget(selection.connectUrl);  
+      if (selection != selected) {
+        try {
+          context.target.setTarget(selection);  
+        } catch (error) {
+          context.target.setTarget(NO_TARGET);
```
andrewazores, line 90: I guess displaying a notification with the error message might also make sense, so that the user can tell what went wrong? Otherwise, if somehow this case is ever tripped, the user is going to have no idea what went wrong and won't be able to file helpful bug reports for us.

Alexjsenn: Yeah, I was thinking on the one hand the user probably cannot do anything about it, but it makes sense to indicate what went wrong.

andrewazores (approved):

PR #152, opened 2020-10-26
UI to upload SSL Certificates

### LINKED ISSUE #135: UI for uploading SSL certs
See https://github.com/rh-jmc-team/container-jfr/issues/273

Alexjsenn:
Fixes #135 
Would the message regarding the restart required need to be different depending on the platform? I am not sure what would be the best way to indicate this.
Also, I am not sure if the way I check if the file type is compatible is exhaustive or the best way to do so.

### DIFF
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -128,8 +128,7 @@ export class ApiService {
       }
 
       return this.target.target().pipe(concatMap(target =>
-        this.sendRequest(`targets/${encodeURIComponent(target.connectUrl)}/recordings`, 
-        "v1", {
+        this.sendRequest('v1', `targets/${encodeURIComponent(target.connectUrl)}/recordings`, {
           method: 'POST',
           body: form,
         }).pipe(
@@ -145,8 +144,7 @@ export class ApiService {
 
   createSnapshot(): Observable<boolean> {
     return this.target.target().pipe(concatMap(target =>
-      this.sendRequest(`targets/${encodeURIComponent(target.connectUrl)}/snapshot`, 
-      "v1", {
+      this.sendRequest('v1', `targets/${encodeURIComponent(target.connectUrl)}/snapshot`, {
         method: 'POST',
       }).pipe(
         tap(resp => {
@@ -163,8 +161,8 @@ export class ApiService {
   archiveRecording(recordingName: string): Observable<boolean> {
     return this.target.target().pipe(concatMap(target =>
       this.sendRequest(
-        `targets/${encodeURIComponent(target.connectUrl)}/recordings/${encodeURIComponent(recordingName)}`,
-        "v1", {
+        'v1', `targets/${encodeURIComponent(target.connectUrl)}/recordings/${encodeURIComponent(recordingName)}`,
+        {
           method: 'PATCH',
           body: 'SAVE',
         }
@@ -178,8 +176,8 @@ export class ApiService {
   stopRecording(recordingName: string): Observable<boolean> {
     return this.target.target().pipe(concatMap(target =>
       this.sendRequest(
-        `targets/${encodeURIComponent(target.connectUrl)}/recordings/${encodeURIComponent(recordingName)}`,
-        "v1", {
+        'v1', `targets/${encodeURIComponent(target.connectUrl)}/recordings/${encodeURIComponent(recordingName)}`,
+        {
           method: 'PATCH',
           body: 'STOP',
         }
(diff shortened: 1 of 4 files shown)


andrewazores:
> Would the message regarding the restart required need to be different depending on the platform?

I think a generic "restart is needed to apply changes" would be sufficient. The user/administrator on whatever platform should hopefully be familiar enough with that platform to be able to restart a container.

andrewazores:
> Also, I am not sure if the way I check if the file type is compatible is exhaustive or the best way to do so.

Do you mean the `uploadFile.type` value? I'm not sure where this comes from - I suppose it's a file type detected by the browser automatically, in which case that's really about all we should be doing from within the web-client. The container-jfr backend is where the real validation is performed. The front-end should just be checking that the selected filetype looks like a plausible certificate to help the user understand what it is that they should be uploading.

On that note, the `dropzoneProps` property of the `FileUpload` component should probably also be set (see `ArchiveUploadModal` which does this for `.jfr` files) to hint to the user/browser that only certain files should be selectable from the file picker dialog.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -381,10 +405,10 @@ export class ApiService {
     );
   }
 
-  private sendRequest(path: string, config?: RequestInit): Observable<Response> {
+  private sendRequest(path: string, apiVersion: ApiVersion, config?: RequestInit): Observable<Response> {
```
andrewazores, line 408: Doesn't really matter, but I think having `apiVersion` before `path` might be a bit more readable? The full API paths have the form `/api/v1/foo`, so specifying this as `'/foo', 'v1'` seems maybe backward.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -361,6 +367,24 @@ export class ApiService {
       );
   }
 
+  uploadSSLCertificate(file: File): Observable<string> {
+    const body = new window.FormData();
+    body.append('cert', file);
+    return this.sendRequest('certificates', "v2", {
```
andrewazores, line 373: For code style consistency the string literal `"v2"` here (and other `"v1"`s above) should use single-quotes, not double-quotes.

src/app/Shared/Services/Api.service.tsx
Alexjsenn: > On that note, the `dropzoneProps` property of the `FileUpload` component should probably also be set (see `ArchiveUploadModal` which does this for `.jfr` files) to hint to the user/browser that only certain files should be selectable from the file picker dialog.
I tried getting it to work with multiple extensions but couldn't figure it out, according to the patterfly page and its link to dropzoneProps, the accept could be a string array, but it apparently is only allowed to be a string. Therefore, I didn't want to restrict it to a single extension.

Alexjsenn:
> On that note, the `dropzoneProps` property of the `FileUpload` component should probably also be set (see `ArchiveUploadModal` which does this for `.jfr` files) to hint to the user/browser that only certain files should be selectable from the file picker dialog.

I tried getting it to work with multiple extensions but couldn't figure it out, according to the patterfly page and its link to dropzoneProps, the accept could be a string array, but it apparently is only allowed to be a string. Therefore, I didn't want to restrict it to a single extension.

andrewazores:
Hmm, the PF docs do say "string array", but the docs for dropzone itself suggest otherwise:

https://react-dropzone.js.org/

So, `accept: '.cer,.der'` should probably work, or maybe this can/should mirror the `fileUpload.type` you've already defined by `accept: 'application/x-x509-ca-cert, application/pkix-cert'`.

Alexjsenn:
Ok yes, that seems to work!

andrewazores (approved):
Looks good. I think a good enhancement for the new Security nav item would be some kind of view that provides a basic listing of the certificates that ContainerJFR already knows about - at least the name of the certificate and the file containing it, but maybe also some other properties of the certificate like the expiry date, issuer, subject, and the date created/modified on disk in the backend.

Alexjsenn:
> Looks good. I think a good enhancement for the new Security nav item would be some kind of view that provides a basic listing of the certificates that ContainerJFR already knows about - at least the name of the certificate and the file containing it, but maybe also some other properties of the certificate like the expiry date, issuer, subject, and the date created/modified on disk in the backend.

Yes, I was thinking about that, I am guessing we would only list the certificates that users may have added in the truststore directory, and not the default ones already in the truststore (there are a lot of them).

andrewazores:
Yea, just the user-added ones in the truststore directory. The other ones in the truststore come from the base image and there really are a lot of them, so I don't think we need to show them in our application.

PR #154, opened 2020-10-28
When connection fails due to SSL, show link to upload cert

### LINKED ISSUE #151: Target connection failure due to SSL trust should display informative message
Currently, when a target connection attempt fails due to an SSL trust issue on the backend, a simple error notification is displayed in the web-client. This does not inform the user of what the root problem was or how to fix it. With rh-jmc-team/container-jfr#307 fixed the response from the server will be more informative. The web-client should implement a message dialog for this scenario similar to the JMX auth prompt that already exists. Initially the dialog may only inform the user of the problem and how to fix it by manually adding the target's SSL cert to the server's truststore. After #135 the dialog should also somehow link to or summon the UI for uploading certificates to the server directly from the web-client.

Alexjsenn:
Fixes #151

### DIFF
--- a/src/app/AppLayout/SslErrorModal.tsx
+++ b/src/app/AppLayout/SslErrorModal.tsx
@@ -35,76 +35,39 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-import { Observable, Subject, BehaviorSubject } from 'rxjs';
+import * as React from 'react';
+import { Button, Modal, ModalVariant, Text } from '@patternfly/react-core';
+import { useHistory, useRouteMatch } from 'react-router-dom';
 
-export const NO_TARGET = {} as Target;
-
-export interface Target {
-  connectUrl: string;
-  alias: string;
+export interface SslErrorModalProps {
+    visible: boolean;
+    onDismiss: () => void;
 }
 
-class TargetService {
-  private readonly _target: Subject<Target> = new BehaviorSubject(NO_TARGET);
-  private readonly _authFailure: Subject<void> = new Subject();
-  private readonly _authRetry: Subject<void> = new Subject();
-  private readonly _credentials: Map<string, string> = new window.Map();
-  private readonly _sslFailure: Subject<void> = new Subject();
+export const SslErrorModal: React.FunctionComponent<SslErrorModalProps> = (props) => {
+    const routerHistory = useHistory();
+    const { url } = useRouteMatch();
 
-  setTarget(target: Target): void {
-    if (target === NO_TARGET || (!!target.alias && !!target.connectUrl)) {
-      this._target.next(target);
-    } else {
-      throw new Error("Malformed target");
+    const handleClick = () => {
+        routerHistory.push('/security');
+        props.onDismiss();
     }
-  }
-
-  target(): Observable<Target> {
-    return this._target.asObservable();
-  }
-
-  authFailure(): Observable<void> {
-    return this._authFailure.asObservable();
-  }
-
-  setAuthFailure(): void {
-    this._authFailure.next();
-  }
-
-  authRetry(): Observable<void> {
-    return this._authRetry.asObservable();
-  }
-
-  setAuthRetry(): void {
-    this._authRetry.next();
-  }
-
-  hasCredentials(target: string): boolean {
-    return this._credentials.has(target);
-  }
-
-  deleteCredentials(target: string): void {
-    this._credentials.delete(target);
-  }
-
-  setCredentials(target: string, credentials: string): void {
-    this._credentials.set(target, credentials);
-  }
-
-  getCredentials(target: string): string | undefined {
-    return this._credentials.get(target);
-  }
-
-  sslFailure(): Observable<void> {
-    return this._sslFailure.asObservable();
-  }
-
-  setSslFailure(): void {
-    this._sslFailure.next();
-  }
-
+    
+    return (
+        <Modal
+            isOpen={props.visible}
+            variant={ModalVariant.medium}
+            showClose={true}
+            onClose={props.onDismiss}
+            title="SSL Error"
+            description="The connection failed because the SSL Certificate for the target is not trusted"
+        >
+            <div>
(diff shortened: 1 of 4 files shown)


src/app/ErrorView/ErrorView.tsx
```diff
@@ -44,15 +44,31 @@ export interface ErrorViewProps {
...
+            <br/>
+            <Bullseye>
+                <ExclamationCircleIcon size='md' color='Red'/>
+            </Bullseye>
+            <Bullseye>
+                <Text>Error:&nbsp;{props.message}</Text>
+            </Bullseye>
+            <Bullseye>
+                <Text>
+                    To add the SSL certificate for this target, click <a href='/Security'>here</a> or go to the Security tab
```
andrewazores, line 59: Using an `<a>` anchor like this causes a page refresh. I think the better mechanism to use here is what the "Create" button on the Active Recording List does, which gives in-application navigation with URL updates without refreshing the client.

src/app/ErrorView/ErrorView.tsx
```diff
@@ -44,15 +44,31 @@ export interface ErrorViewProps {
...
 export const ErrorView: React.FunctionComponent<ErrorViewProps> = (props) => {
-    return (<>
-        <br/>
-        <Bullseye>
-            <ExclamationCircleIcon size='md' color='Red'/>
-        </Bullseye>
-        <Bullseye>
-            <Text>
+
+    if (props.message === 'Target SSL Untrusted'){
```
andrewazores, line 48: The message string shouldn't be used like this, since it's easy for that string to get reworded or otherwise modified (even returned in a different language/localization, for example). The HTTP status code from the response should be used to determine when this interaction is presented to the user, not the HTTP status message.

src/app/ErrorView/ErrorView.tsx
```diff
@@ -44,15 +44,31 @@ export interface ErrorViewProps {
 }
 
 export const ErrorView: React.FunctionComponent<ErrorViewProps> = (props) => {
```
andrewazores, line 46: Related to the previous comment, I don't think I like the idea of implementing this as a specialization of the `ErrorView` - this precludes us from presenting this information to the user unless the failure occurs specifically when they are looking at a view where replacing the content with an `ErrorView` makes sense. I think a message dialog as called out in the original issue would be more versatile, since it can be presented in any scenario, and I think it also provides a better hint to the user that something is wrong but their action is required to (hopefully) fix it.

src/app/ErrorView/ErrorView.tsx
andrewazores: Also, I think associating the page movement with a button rather than an anchor link would fit in better with the style of the application. It could simply be a label like `To add the SSL certificate for this target, go to [Security]`, where `[]` denotes a clickable button.

src/app/ErrorView/ErrorView.tsx
```diff
@@ -44,15 +44,31 @@ export interface ErrorViewProps {
...
+            <Bullseye>
+                <Text>Error:&nbsp;{props.message}</Text>
+            </Bullseye>
+            <Bullseye>
+                <Text>
+                    To add the SSL certificate for this target, click <a href='/Security'>here</a> or go to the Security tab
+                </Text>
+            </Bullseye>
+        </>)
+    } else {
```
andrewazores, line 63: Wherever this specialized case of an error message is implemented, whether it's here in the `ErrorView` or as a dialog, it shouldn't be done with a simple `if/else` check on some property. Otherwise, the view/dialog will balloon in size as more cases are added. This would be a good time to exercise some React view composition.

andrewazores (changes requested):

src/app/ErrorView/ErrorView.tsx
Alexjsenn: So it would be similar to the JMX authentication modal? And I am guessing it would just present the error and the button to go to the security tab?

andrewazores: Yes, pretty much.

src/app/ErrorView/ErrorView.tsx
Alexjsenn: I am trying to catch the SSL error, but I am wondering how I should handle the case of two handlers that return 502 codes related to Grafana. How can I destinguish between those and all of the rest?

andrewazores: For that one handler that overloads the `502` code for both SSL trust failures and Grafana failures there is not (yet) a reliable way to distinguish those cases. Part of our API V2 work can/should be to come up with a better JSON API response format so that we can embed more detailed information within response bodies rather than relying solely upon HTTP status codes.

Since that's a much larger project, maybe the workaround for now is to remove the `502` overloading by changing the Grafana failure code to some other `5xx` code, and leave `502` for target SSL trust failures. This would be a much smaller PR on the backend that can go along with this one.

Alexjsenn:
I switched to a modal. I think the button should be to the right of the text, not below it, but I couldn't figure out how to arrange it like that.

src/app/AppLayout/SslErrorModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+    }
+    
+    return (
+        <Modal
+            isOpen={props.visible}
+            variant={ModalVariant.medium}
+            showClose={true}
+            onClose={props.onDismiss}
+            title="SSL Error"
+            description="The connection failed because the SSL Certificate for the target is not trusted"
```
andrewazores, line 63: I think there should be a period at the end of this sentence - the description looks more like a piece of the body text than a subheading or something.

andrewazores:
The error notifications corresponding to the `502` still appear in the top right while the modal dialog is also summoned. I don't think the notifications are really needed if a more informative dialog with actions is also presented to the user at the same time.

andrewazores:
> I switched to a modal. I think the button should be to the right of the text, not below it, but I couldn't figure out how to arrange it like that.

Put the `<Button>` inside the `<Text>`

Alexjsenn:
Yes, makes sense, I removed the notification when the modal triggers.

andrewazores (approved):

PR #155, opened 2020-10-29
When auth failure occurs display errorView

### LINKED ISSUE #150: Dismissing the JMX credentials prompt should produce view error state
If the user dismisses the JMX credentials prompt (by clicking the X in the corner, or pressing `Esc`, or `Cancel`), the prompt dialog is closed and the connection attempt is aborted. The loading state in the view "behind" the prompt, ex. the active recordings list, remains in a loading state. This should instead be an error state indicating that the attempt failed because the user refused to authenticate.

Alexjsenn:
Fixes #150 
I don't know if this is the optimal way to solve this, I could not think of any other way because handleError in Api.service.tsx doesn't throw an error, it retries. Therefore, the api call never gets back to the caller with an error, and was showing the loading screen.

### DIFF
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -121,6 +121,13 @@ export const EventTemplates = () => {
     refreshTemplates();
   }, []);
 
+  React.useEffect(() => {
+    const sub = context.target.authFailure().subscribe(() => {
+      setErrorMessage("Auth failure");
+    });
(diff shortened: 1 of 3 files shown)


andrewazores (approved):
This is probably the simplest approach and good enough for now. Eventually we'll have to go through some of this and do some refactoring to reduce the amount of duplicated code between views for this kind of error handling, though.

PR #156, opened 2020-10-30
Improve client-side report caching and error handling

andrewazores:
See also: rh-jmc-team/container-jfr#318


Alexjsenn:
Right after creating a recording (running smoketest) I get the following error:
`java.util.concurrent.ExecutionException: com.redhat.rhjmc.containerjfr.net.internal.reports.SubprocessReportGenerator$ReportGenerationException: [1] Connection to target JVM failed.`
Does this have to do with the JMX authentication?

andrewazores:
Did you try to connect back to container-jfr itself for that? That's a known bug with the new subprocess report generation - something about the Podman setup in run.sh/smoketest.sh doesn't work with the subprocess trying to connect back to the parent process. If you try to connect to the other sample application in the smoketest then it should successfully connect.

Alexjsenn (approved):

PR #159, opened 2020-11-05
Configurable auto-refresh

### LINKED ISSUE #158: Auto-refresh should be configurable
Various views have an auto-refresh polling period. This is generally just used to ensure that if some client performs some action (ex. starting/stopping/deleting/archiving recordings) that other clients will eventually see these changes too. Currently these periods are generally set to 30 seconds, but this should be configurable. Once the command channel is replaced with a notification channel for pushed notifications of events like recording started/stopped/deleted/archived the need for polling at all will be greatly reduced and this can be either disabled by default or set to a much longer period.

andrewazores:
Fixes #158


Alexjsenn (approved):
Seems to work perfectly!

PR #160, opened 2020-11-06
Use error response body as error detail message

andrewazores:
Do not expect response status message to contain additional details, rather these additional details are written in the response body

See also rh-jmc-team/container-jfr#322


andrewazores:
Testing steps are the same as here: https://github.com/rh-jmc-team/container-jfr/pull/322#issuecomment-723291526

The error notification displayed in most instances should have the response status code and default short message in the notification title. and the custom exception detail message in the notification body.

vic-ma (approved):

Alexjsenn (approved):

PR #161, opened 2020-11-09
Configure webpack bundle splitting

### LINKED ISSUE #149: app.bundle.js is huge
`app.bundle.js` produced by Webpack in `npm run build` is currently `3.45MiB`. This is gigantic and must be reduced.

https://github.com/webpack-contrib/webpack-bundle-analyzer can be used to figure out why the bundle is so large.

Bundle-splitting is probably the answer. There are various tools and tutorials for Webpack code/bundle splitting online.

andrewazores:
Adds configuration to split app.bundle.js so that the main app.bundle.js
contains generally only container-jfr-web assets, and dependencies
(node_modules) are split out into separate bundle files. The
app.bundle.js for production builds also gains a content hash part of
its filename (as do all vendored dependency bundles) to help ensure that
assets are not incorrectly cached by browsers or webservers.

This allows the client web browser to only download the portions of the
web-client that have changed when (re)visiting the application, reducing
load times. This also allows for better cache handling since
infrequently-updated dependencies do not need to be re-downloaded every
time the application is rebuilt.

Fixes #149

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -13,6 +13,28 @@ module.exports = {
       template: path.resolve(__dirname, 'src', 'index.html')
     }),
   ],
+  // https://medium.com/hackernoon/the-100-correct-way-to-split-your-chunks-with-webpack-f8a9df5b7758
+  optimization: {
+    runtimeChunk: 'single',
+    splitChunks: {
+      chunks: 'all',
+      maxInitialRequests: Infinity,
+      minSize: 0,
+      cacheGroups: {
+        vendor: {
+          test: /[\\/]node_modules[\\/]/,
+          name(module) {
+            // get the name. E.g. node_modules/packageName/not/this/part.js
+            // or node_modules/packageName
+            const packageName = module.context.match(/[\\/]node_modules[\\/](.*?)([\\/]|$)/)[1];
+
+            // npm package names are URL-safe, but some servers don't like @ symbols
+            return `npm.${packageName.replace('@', '')}`;
+          },
+        },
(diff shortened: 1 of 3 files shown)


andrewazores:
To test:

1. Production mode:
Do a standard `mvn clean verify` and then `sh smoketest.sh` to run the resulting image. Open a new tab in your web browser, then open the developers' tools (ex. F12), switch to the Network/Requests tab, and visit the web-client. Take note of how many requests are made for web-client assets (`*.bundle.js` in particular) and the response codes. Then repeat this process while on this branch. With this patch applied, many more bundles should be downloaded, but each should be relatively quite small. After the first load, the response code received should be 304, indicating that the asset has not changed and the web server tells the browser to reuse cached assets.

2. Development mode:
Similar to above, but set the CORS origin when running `smoketest.sh` and then do an `npm run start:dev` to run the web-client using the Webpack Dev Server. The same behaviour with `304`s and cached assets should also be observed.

Alexjsenn:
I see the smaller bundles, but I don't see the 304 code. It shows a 200 greyed out and then says (memory cache). Is that fine?

andrewazores:
That might be how your browser represents it, I'm not sure. As long as your browser somehow indicates that it has reused a cached version of the bundle then that should mean it's working as expected. I think you would also see the `304` response code in the ContainerJFR backend logs.

Alexjsenn:
Ok, makes sense!

Alexjsenn (approved):

PR #164, opened 2020-12-04
Support Notifications Channel

### LINKED ISSUE #162: Support Notifications Channel
The CommandChannel implementation should be renamed/repurposed to support the new Notifications Channel (rh-jmc-team/container-jfr#330) instead. The API would be largely similar, however the data flow is now one-way, server->client.

andrewazores:
Fixes #162

This heavily refactors, simplifies, and renames the `CommandChannel` to the `NotificationsChannel`, with some of the old functionality moved to the `ApiService`.

The new `NotificationsChannel` allows consumer components to subscribe to notifications sent by the ContainerJFR backend and respond in various ways. Two such consumers are included: the first is the `NotificationsChannel` itself, which displays a notification message box when `WS_CLIENT_ACTIVITY` notification messages are received (when a WebSocket client (dis) connects). The second is the `TargetSelect` component, which now listens for `TargetJvmDiscovery` notifications, which are emitted by the backend when async target discovery observes a new target appearing or an old target disappearing. The `TargetSelect` component uses this notification to update its own state, rather than relying on polling or the user manually requesting a refresh of the data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "@patternfly/react-icons": "^4.3.2",
     "@patternfly/react-styles": "^4.3.2",
     "@patternfly/react-table": "^4.3.2",
+    "@types/lodash": "^4.14.165",
     "express": "^4.17.1",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
--- a/src/app/Events/EventTypes.tsx
+++ b/src/app/Events/EventTypes.tsx
@@ -172,7 +172,7 @@ export const EventTypes = () => {
     });
 
     setDisplayedTypes(rows);
-  }, [context.commandChannel, currentPage, perPage, filterTypesByText, openRow]);
+  }, [currentPage, perPage, filterTypesByText, openRow]);
 
   const onCurrentPage = (evt, currentPage) => {
     setOpenRow(-1);
@@ -203,7 +203,7 @@ export const EventTypes = () => {
   if (errorMessage != '') {
     return (<ErrorView message={errorMessage}/>)
   } else if (isLoading) {
-    return (<LoadingView/>) 
+    return (<LoadingView/>)
   } else {
     return (<>
       <Toolbar id="event-types-toolbar">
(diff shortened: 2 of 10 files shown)


andrewazores:
To test the `WS_CLIENT_ACTIVITY` handling, simply connect and disconnect with a second WebSocket client. This can be a second web-client instance, or ex. `websocat ws://0.0.0.0:8181/api/v1/command`.

To test the `TargetJvmDiscovery` handling, use `smoketest.sh` in the ContainerJFR repo. Then do `podman kill vertx-fib-demo` to cause a target to disappear, and `podman run --name vertx-fib-demo -pod container-jfr --rm -d quay.io/andrewazores/vertx-fib-demo:0.4.0` to cause it to reappear.

Alexjsenn:
Should a notification show up everytime a target appears or disappears? While testing it showed up once, the first time I disconnected it, and now its not showing anymore when it finds or loses a target. The target list does get updated (I turned off the auto refresh) so I guess its just an issue with the notifications?

andrewazores:
No, the intent is not for a notification pop-up to appear any time that a target (dis)appears. The message on the notification channel should only be consumed to update the state of the `TargetSelect` widget.

A notification pop-up will appear to inform the user when a new WebSocket client connects to the server however, which should also include the current web-client session you're looking at yourself.

Alexjsenn:
Ah perfect, reading through the changes I realized it only shows a LOST notification if you had selected the target and then it was lost, which makes sense!

Alexjsenn (approved):

andrewazores:
I just realized I forgot to handle the `CHANGED` event. This is never emitted by the upcoming OpenShift watch-based async discovery, and I'm not sure under what circumstances it is emitted by the JDP discovery client which would be active with `run.sh/smoketest.sh`, so testing this is a bit unclear. I have tried to implement it so that the target state is updated by removing targets with service URLs matching the `CHANGED` one, then re-adding the new one to the state, and finally re-selecting the new one by assuming that it has the same `connectUrl`. The idea here being that the `connectUrl` (JMXServiceURL) should be consistent - if something has changed about a target, then it would be the alias, or in the future some labels or other attributes.

andrewazores:
@Alexjsenn could you take another look at this? There is only one new commit since the last time you reviewed it, although that was a month ago now so I'm sure this isn't fresh in your memory anymore.

andrewazores:
https://github.com/rh-jmc-team/container-jfr/issues/388 tracks the `CHANGED` event discussion.

Issue #212, opened 2021-04-27
Manually-added targets


andrewazores:
UI for https://github.com/cryostatio/cryostat/issues/468

andrewazores:
Closed by #213

PR #213, opened 2021-04-29
Custom targets

andrewazores:
#212

### DIFF
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -150,6 +150,24 @@ export class ApiService {
     );
   }
 
+  createTarget(target: Target): Observable<boolean> {
+    const form = new window.FormData();
+    form.append('connectUrl', target.connectUrl);
+    if (!!target.alias && !!target.alias.trim()) {
+      form.append('alias', target.alias);
+    }
+    return this.sendRequest(
+      'v2', `targets`,
+      {
+        method: 'POST',
+        body: form,
+      }
+    ).pipe(
+      map(resp => resp.ok),
+      first(),
+    );
+  }
+
   deleteTarget(target: Target): Observable<boolean> {
     return this.sendRequest(
       'v2', `targets/${encodeURIComponent(target.connectUrl)}`,
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -1,8 +1,8 @@
 /*
  * Copyright The Cryostat Authors
- * 
+ *
  * The Universal Permissive License (UPL), Version 1.0
- * 
+ *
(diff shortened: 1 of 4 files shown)


Alexjsenn:
Hey, so I was testing this out and I created a fake target similar to what you had suggested when testing out the api implementation, but when I select this new target and then go to the recordings tab, it crashes. I got the same result in both chrome and firefox. Do you get the same behaviour or is something wrong on my end?

andrewazores:
What happens when it "crashes" - is this a frontend error message or an actual backend Cryostat API error?

Did the target you defined point to a target that actually existed/was reachable?

Alexjsenn:
No, it did't point to anything that existed. I ran it in smoketest and I added: service:jmx:rmi:///jndi/rmi://cryostat:9099/jmxrmi
The browser tab just stops responding.

andrewazores:
Interesting, thanks for that note - I'll take a look at it. Sounds like I made some kind of React mistake.

andrewazores:
> No, it did't point to anything that existed. I ran it in smoketest and I added: service:jmx:rmi:///jndi/rmi://cryostat:9099/jmxrmi
> The browser tab just stops responding.

Actually, I see this behaviour on `main`, not just on this PR's branch. Do you see that too? I'll investigate the root cause further and open a separate PR.

andrewazores:
After a git bisect, the first bad revision appears to be `441b63dbbe695a5ff36fe08c3a196d9808078343` - that's #214 , which touched the `ActiveRecordingList`, so that looks plausible.

@Josh-Matsuoka if you have some free cycles, could you take another look at that? If not then I'll take it on.

andrewazores:
@Alexjsenn for the purposes of testing the functionality in this PR, I'm going to rebase it onto `main` as of a couple of commits back and force-push it here.

Try adding `service:jmx:rmi:///jndi/rmi://localhost:0/jmxrmi` as a custom target definition as well - this should instruct the Cryostat JVM to connect back to itself, so you can easily test out a target definition that does actually point to something.

Alexjsenn:
> @Alexjsenn for the purposes of testing the functionality in this PR, I'm going to rebase it onto `main` as of a couple of commits back and force-push it here.
> 
> Try adding `service:jmx:rmi:///jndi/rmi://localhost:0/jmxrmi` as a custom target definition as well - this should instruct the Cryostat JVM to connect back to itself, so you can easily test out a target definition that does actually point to something.

Ok, it works perfectly now!

Alexjsenn (approved):

Josh-Matsuoka:
@andrewazores hey, I'll take a look at that bug this week, sorry about that.

andrewazores:
> @andrewazores hey, I'll take a look at that bug this week, sorry about that.

No worries, I should have caught it in review as well.

Issue #165, opened 2020-12-10
Enhanced Notifications handling


andrewazores:
After #162 is resolved, further enhancements to Notifications handling should be made, matching backend enhancements in rh-jmc-team/container-jfr#354 .

jan-law:
This issue seems related to #284 and #267. Were there any other improvements you were thinking of?

andrewazores:
I think that more or less covers it, really. I'll close this issue as superceded by those two.

PR #214, opened 2021-05-10
Add Handlers to refresh recording list when notifications are received

Josh-Matsuoka:
Hi,

This PR addresses https://github.com/cryostatio/cryostat-web/issues/165 , adding notification handlers for each of the new notifications emitted by the backend on the Notifications channel. Whenever the web client receives one of these it will trigger a refresh of the recordings list to reflect new recordings being created/deleted/archived in the backend.

### DIFF
--- a/src/app/RecordingList/ActiveRecordingsList.tsx
+++ b/src/app/RecordingList/ActiveRecordingsList.tsx
@@ -67,17 +67,19 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
   const [errorMessage, setErrorMessage] = React.useState('');
   const { url } = useRouteMatch();
 
-  const RECORDING_CREATED = "RecordingCreated";
-  const RECORDING_DELETED = "RecordingDeleted";
-  const RECORDING_SAVED = "RecordingSaved";
-  const RECORDING_ARCHIVED = "RecordingArchived";
-
   const tableColumns: string[] = [
     'Name',
     'Start Time',
     'Duration',
     'State',
   ];
+
+  enum NotificationCategory {
+    RecordingCreated = 'RecordingCreated',
+    RecordingDeleted = 'RecordingDeleted',
+    RecordingSaved = 'RecordingSaved',
+    RecordingArchived = 'RecordingArchived'
+  };
 
   const addSubscription = useSubscriptions();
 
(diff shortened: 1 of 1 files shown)


src/app/RecordingList/ActiveRecordingsList.tsx
```diff
@@ -66,6 +67,11 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
   const [errorMessage, setErrorMessage] = React.useState('');
   const { url } = useRouteMatch();
 
+  const RECORDING_CREATED = "RecordingCreated";
+  const RECORDING_DELETED = "RecordingDeleted";
+  const RECORDING_SAVED = "RecordingSaved";
+  const RECORDING_ARCHIVED = "RecordingArchived";
```
andrewazores, line 73: Could these be made into an enum? TypeScript has syntax like this:

```
enum NotificationCategory {
  RecordingCreated = 'RecordingCreated',
  ...
}
```

which I think would work nicely here.

src/app/RecordingList/ActiveRecordingsList.tsx
```diff
@@ -122,6 +128,34 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
...
+      .subscribe(v => {
+        const recordingName = Object.values(v.message)[0];
+        const targetId = Object.values(v.message)[1];
+        notifications.info('WebSocket Client Activity', `Recording ${recordingName} created in target: ${targetId}`);
+        refreshRecordingList();
+      });
+      const archived_sub = context.notificationChannel.messages(RECORDING_ARCHIVED)
+      .subscribe(v => {
+         const recordingName = Object.values(v.message)[0];
+         notifications.info('WebSocket Client Activity', `Recording ${recordingName} archived`);
```
andrewazores, line 142: And on the topic of the notification message itself, the title `WebSocket Client Activity` was meant to indicate that a WebSocket client had (dis)connected, so that a user would know if another user had connected another remote session, or to give an individual user some notification if their connection to the server died - it wasn't intended as a catch-all title for all server-side notifications.

I think I would prefer to see a title for this event like `Recording Archived` and the message body say `${recordingName} was archived`. What do you think?

src/app/RecordingList/ActiveRecordingsList.tsx
Josh-Matsuoka: Fixed with the most recent push, thanks

src/app/RecordingList/ActiveRecordingsList.tsx
```diff
@@ -122,6 +128,34 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
     );
   }, []);
 
+  React.useEffect(() => {
+    const created_sub = context.notificationChannel.messages(RECORDING_CREATED)
```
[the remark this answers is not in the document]
Josh-Matsuoka: Changed to use addSubscription, thanks

src/app/RecordingList/ActiveRecordingsList.tsx
Josh-Matsuoka: Makes sense to me, I've changed this now

src/app/RecordingList/ActiveRecordingsList.tsx
```diff
@@ -122,6 +128,34 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
...
+      .subscribe(v => {
+        const recordingName = Object.values(v.message)[0];
+        const targetId = Object.values(v.message)[1];
+        notifications.info('WebSocket Client Activity', `Recording ${recordingName} created in target: ${targetId}`);
```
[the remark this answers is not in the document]
Josh-Matsuoka: I've changed this now to handle the message as a proper data type similar to how TargetSelect does it

src/app/RecordingList/ActiveRecordingsList.tsx
```diff
@@ -73,6 +74,13 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
     'State',
   ];
 
+  enum NotificationCategory {
+    RecordingCreated = 'RecordingCreated',
+    RecordingDeleted = 'RecordingDeleted',
+    RecordingSaved = 'RecordingSaved',
+    RecordingArchived = 'RecordingArchived'
+  };
+
```
andrewazores, line 83: Last round of minor changes - could you pull up this enum and the `RecordingNotificationEvent` interface so that they're declared outside of the component, but still within this same file?

src/app/RecordingList/ActiveRecordingsList.tsx
```diff
@@ -122,6 +130,47 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
...
   }, []);
 
+  React.useEffect(() => {
+    addSubscription(context.notificationChannel.messages(NotificationCategory.RecordingCreated)
+      .subscribe(v => {
+        const event: RecordingNotificationEvent = v.message;
+        notifications.info('Recording Created', `${event.recording} created in target: ${event.target}`);
+        refreshRecordingList();
+      }));
+  }, []);
```
andrewazores, line 145: I think the effect dependencies list (`[]` parameter) needs to include the various dependencies the lambda function has here - so that would be `context.notificationChannel`, `notifications`, and `refreshRecordingList`. Although in practice this is unlikely to mean much since all of these are references to things that won't change for the duration of the client lifecycle, it's still technically correct for code cleanliness.

src/app/RecordingList/ActiveRecordingsList.tsx
Josh-Matsuoka: Done, thanks

src/app/RecordingList/ActiveRecordingsList.tsx
Josh-Matsuoka: Ah I see, that's good to know. I've fixed this now

andrewazores (approved):
Looks good, thank you!

Issue #216, opened 2021-05-19
Tab hangs when visiting Recordings


andrewazores:
https://github.com/cryostatio/cryostat-web/pull/213#issuecomment-843545395

PR #228, opened 2021-06-22
Tab hangs when visiting Recordings

Josh-Matsuoka:
This PR resolves https://github.com/cryostatio/cryostat-web/issues/216 . The hang was caused by including the refreshRecordingsList function in the dependency array for the notifcation handlers. Since React does shallow comparison when determining if the dependencies have changed, this will cause problems as the function object will have a different reference on each render despite essentially being the same. 

This patch changes the way the function dependency is handled, wrapping it inside of a React.useCallback hook so that the notification handlers will be called when the dependencies of refreshRecordingList change (effectively what we want from including it in the dependency array), resolving the hang.

### DIFF
--- a/src/app/RecordingList/ActiveRecordingsList.tsx
+++ b/src/app/RecordingList/ActiveRecordingsList.tsx
@@ -144,7 +144,7 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
           refreshRecordingList();
         }, [refreshRecordingList]);
       }));
-  }, [context.notificationChannel, notifications, refreshRecordingList]); 
+  }, [context.notificationChannel, notifications]); 
 
   React.useEffect(() => {
     addSubscription(context.notificationChannel.messages(NotificationCategory.RecordingSaved)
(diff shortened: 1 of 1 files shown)


andrewazores:
Hmm. `refreshRecordingList` itself is already wrapped in a `useCallback` call, so I would expect that since that reference is already to a memoized wrapper of the function that the shallow comparison should still behave as expected. Are the dependencies of `refreshRecordingList` itself perhaps changing on each render and breaking the memoization?

src/app/RecordingList/ActiveRecordingsList.tsx
```diff
@@ -140,36 +140,44 @@ export const ActiveRecordingsList: React.FunctionComponent<ActiveRecordingsListP
       .subscribe(v => {
         const event: RecordingNotificationEvent = v.message;
         notifications.info('Recording Created', `${event.recording} created in target: ${event.target}`);
-        refreshRecordingList();
+        React.useCallback(() => {
+          refreshRecordingList();
+        }, [refreshRecordingList]);
```
andrewazores, line 145: Does `refreshRecordingList` actually get called when this is done? From my understanding of the `useCallback` hook, this returns a memoized wrapper of the provided function (which is in this case actually a lambda/fat-arrow-function), but I'm not sure that it will actually cause the function (memoized version or original) to be invoked.

Josh-Matsuoka:
closing in favor of https://github.com/cryostatio/cryostat-web/pull/229

PR #243, opened 2021-08-19
fix(settings): move from vert nav to masthead tool

### LINKED ISSUE #242: Settings page should be accesed via masthead, not vertical nav
https://www.patternfly.org/v4/components/page/design-guidelines/#page-with-vertical-navigation

> 3. Utility items: Place any items in the utility area that you want to have accessible from all pages in the application. Utility items are right aligned and placed to the left of the user menu. The following are common items ordered from left to right. All items are optional depending on the needs of your application.
> - Settings: The settings menu contains ways to configure and customize an application. Interacting with settings either takes users to a settings area or reveals a dropdown menu with more options. The settings menu is represented by the fa-cog icon.

andrewazores:
Fixes #242

image

image


ebaron (approved):
Works as intended

PR #247, opened 2021-08-24
fix(targets): hold targets state in new service

### LINKED ISSUE #246: In-app navigation causes repeated GET /api/v1/targets queries

andrewazores:
Fixes #246

Query `/api/v1/targets` at web-client startup to find initial list of known
targets state. This list is updated incrementally via notification channel
updates asynchronously, which the `TargetSelect` component receives via RxJS
Observable emissions to update the graphical state accordingly.

This removes the need for a `GET /api/v1/targets` on each in-app navigation
(when the TargetSelect component is re-rendered), increasing page
responsiveness and fluidity. Update handling and state logic is also decoupled
from presentational React component and moved into the TargetsService, making
the component smaller and easier to understand/more maintainable.

To test, open your browser's dev tools panel and go to the Network tab (or
simply watch Cryostat backend logs). Click between Recordings and Events or
any other in-app navigation between pages that contain the TargetSelect
component at the top. On the current upstream main branch, observe that the
browser/logs record a `GET /api/v1/targets` (or even two, actually) on each
in-app navigation. On this PR branch, perform the same test and observe that
only one `GET` is performed at the application startup. The + and trashcan
icons can be used to test the custom target creation/deletion, which should
now update the UI without performing any `GET`. `podman kill/podman run`
(if running the backend using `smoketest.sh`) can be used to trigger async
discovery `FOUND/LOST` events as well, which should also cause the UI to
update without a `GET` query. Clicking the refresh circular-arrow icon on
the TargetSelect should prompt a single `GET /api/v1/targets` query. Enabling
view auto-refresh (gear icon on top nav bar) should cause singular `GET`s to
occur at the specified refresh interval.

### DIFF
--- a/src/app/Shared/Services/Services.tsx
+++ b/src/app/Shared/Services/Services.tsx
@@ -57,7 +57,7 @@ const api = new ApiService(TargetInstance, NotificationsInstance);
 const notificationChannel = new NotificationChannel(api, NotificationsInstance);
 const reports = new ReportService(api, NotificationsInstance);
 const settings = new SettingsService();
-const targets = new TargetsService(api, notificationChannel, NotificationsInstance);
+const targets = new TargetsService(api, NotificationsInstance, notificationChannel);
 
 const defaultServices: Services = { target: TargetInstance, targets, api, notificationChannel, reports, settings };
 
--- a/src/app/Shared/Services/Services.tsx
+++ b/src/app/Shared/Services/Services.tsx
@@ -57,7 +57,7 @@ const api = new ApiService(TargetInstance, NotificationsInstance);
 const notificationChannel = new NotificationChannel(api, NotificationsInstance);
 const reports = new ReportService(api, NotificationsInstance);
 const settings = new SettingsService();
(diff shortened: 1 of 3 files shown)


jan-law:
How do I start the web client on your branch? 

The web client's API requests fail, eg `http://localhost:8181/health` returns 'failed to load response data`. I tried the following in this order:
```
~/cryostat (main)> git pull upstream main
~/cryostat (main)> mvn clean package && CRYOSTAT_CORS_ORIGIN=http://localhost:9000 sh run.sh

(new terminal tab)
~/cryostat-web (api-query-service) [1]> npm run start:dev
```

andrewazores:
You'll need to disable SSL as well.
I do:

```bash
CRYOSTAT_CORS_ORIGIN=http://localhost:9000 CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true sh smoketest.sh
```

to start the backend, and

```bash
yarn yarn:frzinstall && yarn start:dev
```

to start the web-client frontend.

PR #251, opened 2021-08-26
feat(login): auto-retry login with existing credentials

### LINKED ISSUE #250: UI should attempt to reconnect after WebSocket connection terminations

andrewazores:
Fixes #250

This PR enhances the login flow.

To test, start the backend as follows:

```bash
$ mkdir conf
$ echo "user:d74ff0ee8da3b9806b18c877dbf29bbde50b5bd8e4dad7a3a725000feb82e8f1" > conf/cryostat-users.properties
$ # the line above creates a Basic auth user with name "user" and password "pass"
$ CRYOSTAT_IMAGE=quay.io/andrewazores/cryostat:periodic-disconnect-ws CRYOSTAT_AUTH_MANAGER=io.cryostat.net.BasicAuthManager CRYOSTAT_CORS_ORIGIN=http://localhost:9000 CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true sh smoketest.sh
```

Start the frontend as usual:

```bash
$ yarn yarn:frzinstall && yarn start:dev
```

You should be presented with the Basic auth login form. Enter `user/pass` credentials and log in. Pay attention to the graphical notifications - there should be one that the connection was accepted here. After ~10s pass, the backend will terminate the WebSocket connection. A graphical notification should be displayed here as well, and you should be returned to the graphical login page momentarily. The login page should have the username and password fields pre-filled. After a very short delay, the login page should disappear and you should be returned back to the in-app page you were on previously. ~10s more later and you should again be disconnected. This will loop indefinitely. The web client should continuously reconnect itself after the connection is lost.

Next, attempt the same procedure as above. However, some time after the first successful web client connection is established, you should forcibly kill the Cryostat backend (Ctrl-C). This should cause the web client to display a disconnection notification and the login page again. Wait some time (and observe in your browser's dev tools network tab that the client is attempting to reconnect via `POST /api/v1/auth`), then restart the Cryostat backend using the same command as the first time. Within a short time (Cryostat backend startup + ~5s or less), the web client should automatically reconnect to the newly restarted backend instance.

Next, retry the above two procedures, but without setting the `CRYOSTAT_AUTH_MANAGER` env var when starting the backend. This will default it to the `NoopAuthManager`, which does not require any authentication credentials. This test case should behave very similarly to the first case with Basic authentication enabled, however the delay between disconnect and successful reconnect is likely to be much shorter.

Finally, the above scenarios can also be tested in CRC to exercise the Bearer authentication login flow. This should behave identically to the Basic authentication flow, albeit with only a single form field to fill, and perhaps a slower reconnection time due to a longer request latency to validate your supplied credentials.

### DIFF
--- a/src/app/Login/Login.tsx
+++ b/src/app/Login/Login.tsx
@@ -39,8 +39,8 @@ import * as React from 'react';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
 import { Card, CardBody, CardFooter, CardHeader, PageSection, Title } from '@patternfly/react-core';
-import { timer } from 'rxjs';
-import { first } from 'rxjs/operators';
+import { combineLatest, timer } from 'rxjs';
+import { debounceTime, first } from 'rxjs/operators';
 import { Base64 } from 'js-base64';
 import { BasicAuthDescriptionText, BasicAuthForm } from './BasicAuthForm';
 import { BearerAuthDescriptionText, BearerAuthForm } from './BearerAuthForm';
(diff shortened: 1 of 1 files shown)


ebaron:
This worked as expected for me, but is the disconnecting every 10s something temporary to illustrate this behaviour?

andrewazores:
Oh yes, the backend is disconnecting all clients every 10s on purpose to illustrate the behaviour. That's part of the custom image I made, not something upstream. I just took the background pruning task that rejects connections from unauthenticated clients every 10s and made it close all connections every 10s instead. ---- On Fri, 27 Aug 2021 16:49:54 -0700  ***@***.***  wrote ----
This worked as expected for me, but is the disconnecting every 10s something temporary to illustrate this behaviour?

ebaron (approved):
Seems good then!

PR #259, opened 2021-08-30
fix(reports): cache to sessionStorage

### LINKED ISSUE #258: In-memory reports caching is unbounded
Currently, the collapsible report frames in the recording lists will cache the report contents in-memory if the report was generated for a stopped or archived recording. This cache is currently unbounded in size however, which is not desirable, and the cache object is also tied to the React application lifecycle, so it is cleared on page refresh.

The cache should have a maximum size so that the application does not continuously grow and consume browser resources, and it would be nice if it persistent a little longer than the React lifecycle.

sessionStorage would meet both of these goals by providing 5-10MB or so of browser-defined cache space, as well as allowing the cache to survive page reloads, automatically clearing when the window/tab is closed.

andrewazores:
Fixes #258

### DIFF
--- a/src/app/Shared/Services/Report.service.tsx
+++ b/src/app/Shared/Services/Report.service.tsx
@@ -1,8 +1,8 @@
 /*
  * Copyright The Cryostat Authors
- * 
+ *
  * The Universal Permissive License (UPL), Version 1.0
- * 
+ *
  * Subject to the condition set forth below, permission is hereby granted to any
  * person obtaining a copy of this software, associated documentation and/or data
  * (collectively the "Software"), free of charge and under any and all copyright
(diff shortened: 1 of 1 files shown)


src/app/Shared/Services/Report.service.tsx
```diff
@@ -80,21 +80,30 @@ export class ReportService {
         const isArchived = !isActiveRecording(recording);
         const isActivedStopped = isActiveRecording(recording) && recording.state === RecordingState.STOPPED;
         if (isArchived || isActivedStopped) {
-          this.reports.set(recording.reportUrl, report);
+          try {
+            sessionStorage.setItem(this.key(recording), report);
+          } catch (error) {
+            this.notifications.warning('Report Caching Failed', error.message);
+            sessionStorage.clear();
```
jan-law, line 87: If the session storage is full and throws an error, is the purpose of `sessionStorage.clear()` to make space for the report URL the next time the report is requested? 

https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem#exceptions

andrewazores: Yes, I figure if the store is full then it's time to clear it out to make room for more cached items. It isn't very elegant, but I don't really want to implement a whole client-side LRU cache on top of sessionStorage. Since we don't know which other items in cache are oldest or least recently accessed, just that the cache is full and we couldn't fit the most recently requested thing, I just clear it out so that the next access can be cached again. It would be even better if I could clear the cache and re-`setItem` the new report into the cache, but this:

```
Particularly, in Mobile Safari (since iOS 5) it always throws when the user enters private mode. (Safari sets the quota to 0 bytes in private mode, unlike other browsers, which allow storage in private mode using separate data containers.)
```

makes that sound infeasible for at least some possible scenarios.

The Cryostat backend does some caching as well anyway, so this client-side cache is primarily just to avoid the network round trip latency of closing and reopening the UI element.

jan-law (approved):
LGTM

Issue #257, opened 2021-08-30
Store user credentials in cookie/localstorage


andrewazores:
The user ~~basic auth (`base64(username:password)`) or~~ bearer (token) credentials should be persisted, such that upon Cryostat web-client reload or new tab opening the user is not required to re-enter their credentials. These credentials must also be persisted in a way that they are unique per Cryostat instance (distinguished by domain) and in a way that they are private to the web-client instance for a given domain (no cross-origin access).

There should also be a corresponding user menu added to the page masthead to allow users to "log out" of the web-client by clearing the stored credentials.

https://www.patternfly.org/v4/components/page/design-guidelines/#masthead

EDIT: actually, in the case that we detect the server is expecting Basic authorization, we should not offer the option to persist these credentials until we have backend support for creating a user session using these credentials. Otherwise we're writing their password to plaintext, which we don't want to do.

Saving the Bearer (JWT) tokens is acceptable, I think, but there should still be a checkbox or something to allow the user to opt-out of having these persisted.

jan-law:
> The UI could save the user's credentials into a cookie or localStorage and have a user menu with logout button, but this just hasn't been implemented.

After looking into cookies vs localStorage vs sessionStorage, I was wondering if you had a preference as to which we should use.
If we use a cookie would we need to notify users?
I know sessionStorage expires when the browser session ends, while localStorage data won't expire. Would it make sense to use sessionStorage after the session from a security perspective?

andrewazores:
sessionStorage is probably the best bet for now, since I think it hardly raises any security concern assuming the user's browser implements it correctly, but it does smooth out the user's experience a fair amount.

Eventually we will want some way to do actual user login sessions with expiring tokens that we can store, rather than directly storing the user's underlying platform credentials, but that's a longer term plan that includes several changes on the backend as well.

andrewazores:
> If we use a cookie would we need to notify users?

I believe this is only needed for things like *tracking* cookies, for analytics and advertising, but not required for functionality of simply keeping the user signed in to the application. It would apply for cookies as well as local/session storage, too.

jan-law:
> EDIT: actually, in the case that we detect the server is expecting Basic authorization, we should not offer the option to persist these credentials until we have backend support for creating a user session using these credentials. Otherwise we're writing their password to plaintext, which we don't want to do.

[...]

PR #261, opened 2021-08-30
feat(auth): Store credentials in sessionStorage

jan-law:
Related #257 

Creates a user session that will remember credentials after the browser tab is refreshed and will only forget the credentials when the user logs out or closes the tab. If Cryostat was configured with Basic authentication, the user menu will display the username and display a user icon otherwise.

Since this PR only changes the frontend, backend support would likely be required for further security improvements.


src/app/AppLayout/AppLayout.tsx
```diff
@@ -152,6 +154,26 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
...
+  }, [serviceContext.api]);
+
+  const handleUserInfoToggle = () => {
+    setShowUserInfoDropdown(!showUserInfoDropdown);
+  }
+
+  const handleLogout = () => {
+    sessionStorage.removeItem('token');
+    setShowUserInfo(false);
+  };
```
jan-law, line 173: When "logout" gets clicked, the token is removed from session storage and the page does not re-render. When I manually refresh the page, another `useEffect` hook will recognize there are no stored credentials and bring me to the login page. How could we tell react to check the authorization  / re-render the page?

andrewazores: The `routes.tsx` component is the one that "knows" if the user is authenticated and displays the Login component if not. `AppLayout` is the right place for the user menu stuff you've added here.

src/app/index.tsx
```diff
@@ -43,14 +43,28 @@ import { AppRoutes } from '@app/routes';
...
+    setAuthenticated(false);
+  };
+
+return(
   <ServiceContext.Provider value={defaultServices}>
     <Router>
-      <AppLayout>
-        <AppRoutes />
+      <AppLayout handleLogout={handleLogout}>
+        <AppRoutes authenticated={authenticated} setAuthenticated={setAuthenticated} handleAuthenticated={handleAuthenticated}/>
```
jan-law, line 63: Since the `useState` hook `authenticated` was declared in `AppRoutes`, how could I elegantly share that prop with my logout function for`AppLayout`?

src/app/Login/Login.tsx
```diff
@@ -74,11 +85,10 @@ export const Login = (props) => {
...
   const handleSubmit = React.useCallback((evt, token, authMethod) => {
-    setToken(token);
     setAuthMethod(authMethod);
     checkAuth(token, authMethod, true);
     evt.preventDefault();
-  }, [setToken, setAuthMethod, checkAuth]);
+  }, [setAuthMethod, checkAuth]);
 
   React.useEffect(() => {
     const sub = serviceContext.api.getAuthMethod().subscribe(setAuthMethod);
```
jan-law, line 94: How does the service api remember token values? When I click `logout`, the sessionStorage token is removed but `checkAuth()` gets called after `api.service.getToken()`, which then automatically logs the user back into Cryostat.

I found `this.token.next(token);` in `Api.service.tsx` and am not sure how that gets saved to `private readonly token = new ReplaySubject<string>(1);`

jan-law: It looks like `ReplaySubject` will remember values for as long as the timeout (default is infinity).
https://rxjs-dev.firebaseapp.com/api/index/class/ReplaySubject 

Maybe we could create some sort of reset function?
https://stackoverflow.com/questions/51145664/resetting-replaysubject-in-rxjs-6/51147023#51147023

jan-law: Why does the `ReplaySubject` token have no expiration timeout?Instead of a reset function, I tried adding a 1s timeout to the `ReplaySubject` `token`, which logs out a user successfully.

src/app/index.tsx
andrewazores: We would probably want to extract the shared state into something like a `Shared/Services/Login.service.tsx`. Parts of the `Api.service` would be moved into there, too, most likely. This way `AppRoutes` and `AppLayout` can both observe and modify the same pieces of application state, and manage their own small pieces of state for presenting things in their views.

The other alternative is to keep all of this state within a component, and pass down the state pieces (user info, logout function, etc.) as props to both the `AppRoutes` and `AppLayout` component. Both of those are in `index.tsx`, so the top-level `App` component could do this pretty easily.

I think extracting out a service is much cleaner, though.

src/app/Login/Login.tsx
andrewazores: The subject has no expiration because there was just never a case for it to have one previously.

Rather than timing it out, you could probably do `this.token.next(null)` or `this.token.next('')` when you call `reset()`. You might need to change the type of the subject to `ReplaySubject<string | undefined>` if you want to null it out. This way, any component that calls `context.api.getToken().subscribe()` after a reset/logout will immediately receive a null/empty string value.

You can prevent components from getting the null/empty string and have them instead receive nothing at all until a valid token is available pretty easily:

```typescript
context.api.getToken()
  .pipe(filter(t => !!t)) // !! in JavaScript gives the "truthiness" of a value - null is falsy and empty string are falsy. Nonempty strings are truthy
  .subscribe( ... )
```

src/app/Login/Login.tsx
```diff
@@ -46,49 +46,48 @@ import { Base64 } from 'js-base64';
...
+    const sub = serviceContext.login.getAuthMethod().subscribe(setAuthMethod);
     checkAuth('', 'Basic'); // check auth once at component load to query the server's auth method
     return () => sub.unsubscribe();
-  }, [serviceContext, serviceContext.api, setAuthMethod, checkAuth]);
+  }, [serviceContext, serviceContext.login, setAuthMethod, checkAuth]);
 
   React.useEffect(() => {
     const sub =
-      combineLatest(serviceContext.api.getToken(), serviceContext.api.getAuthMethod(), timer(0, 5000))
+      combineLatest(serviceContext.login.getToken(), serviceContext.login.getAuthMethod(), timer(0, 5000))
```
jan-law, line 90: When I go to localhost:9000, I can see that four requests, `/auth`, '`v1/recordings', `v1/targets`, and '/auth`, fail with a `401`. The console log tells me that the `Alert` component was trying to render a JSON object in the `{message}` with the text `Error: Unauthorized`, which prevents the page from rendering.

If I comment out the `{message}`, the page will endlessly call `/auth` in sets of two. I think at least one of the `useEffect` hooks is triggering repeatedly.

Why are there two `/auth` calls at the start? Why are both `serviceContext` and `serviceContext.login` dependencies to the `useEffect` hooks?

image

andrewazores: There is one `/auth` call performed unconditionally at application start-up, which it uses to probe the server for the expected auth method (Bearer or Basic). The server responds with a 401 and a header indicating the expected auth method. Once that's known the application can attempt to actually auth the user with the credentials they supply, using the now-known method, so that actual attempt would be the second `/auth` that you see.

The `useEffect` dependency arrays contain references to everything that would "invalidate" the closure of the lambda function passed in if the reference changed, so that React can re-render and recompute the lambda with the updated references when they change, basically. `ServiceContext` and the `LoginService` are both actually "static" in the sense that there is only ever 1 instance for the lifetime of the application, so these should never cause a re-render to occur in practice. Not including them in the array leads to linter warnings, however, and there is no harm in including them.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -71,7 +70,8 @@ export class ApiService {
 
   constructor(
```
jan-law, line 71: I'm trying to find the reason why my branch makes a call to `v1/targets` and `v1/recordings` on the login page and why the HTTP responses appear as toast notifications. On the main branch login page, there are two `401` responses that don't appear as toast notifications. 

I see that `queryForTargets()` gets called in the `TargetsService` constructor and `<TargetSelect/>`.  `doGet<Recording[]>()` is also called in the `ApiService` constructor and a few other components. Since`<TargetSelect />` isn't rendered on the login page, why would I see a call to `/v1/targets` on my branch?

I also noticed the `notifications` request results in a `101 protocol switched` - which protocol is this referring to? Is this relevant to this problem?
image
image

andrewazores: `101 protocol switched` looks like it's just this: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/101
It's a normal part of the WebSocket connection procedure.

HTTP error responses are automatically turned into notifications by the `Api.service` through certain call paths. It looks like you're seeing it coming from here: https://github.com/cryostatio/cryostat-web/blob/main/src/app/Shared/Services/Api.service.tsx#L548 - many of the public methods in that class pipe their results like `.pipe(handleError(err => this.handleError<Response>(err, req))`, so the original component that subscribed to this request Observable will observe the error event, and a notification will be displayed automatically.

Other HTTP calls that don't go through this call path, like the `checkAuth` function (https://github.com/cryostatio/cryostat-web/blob/a29e59b1cf7caed9416722bdcccfd6814fb3a3d9/src/app/Shared/Services/Api.service.tsx#L127), won't automatically display the toast notification. In this particular case it's mostly because of that startup "probing" `/auth` request that I described earlier - that happens every time the React app is opened, and in any real deployment it will fail with a 401, so it would be annoying/confusing to show a notification here. Notifications are instead only shown when the user explicitly tries to authenticate and *that* fails (https://github.com/cryostatio/cryostat-web/blob/main/src/app/Login/Login.tsx#L70).


> I see that queryForTargets() gets called in the TargetsService constructor and <TargetSelect/>. doGet<Recording[]>() is also called in the ApiService constructor and a few other components. Since<TargetSelect /> isn't rendered on the login page, why would I see a call to /v1/targets on my branch?

Without analyzing your code too deeply (yet), I would suspect that the application is briefly considering the user as authenticated and renders the Dashboard view, even just for a single frame, which does contain a TargetSelect component and would trigger the `GET /api/v1/targets` request. Then, state is properly restored and the app considers the user not-authenticated, kicking them back to the Login page and removing the TargetSelect/Dashboard from the rendered view, but not before the request is fired.

jan-law: I tried setting the initial value of  `ReplaySubject authenticated` and the state variable in `routes.tsx` based on the contents of sessionStorage. The `useEffect` hook is also called every time the browser refreshes. The web page still flickers between the login page and the dashboard when I try to login or refresh the page. What else could cause the application to briefly read `authenticated` as `true`?

`routes.tsx`:
```
const [authenticated, setAuthenticated] = React.useState(context.login.isAuthenticated());

  React.useEffect(() => {
    const sub = combineLatest(context.login.getAuthenticated().pipe(distinctUntilChanged()),
      context.notificationChannel.isReady().pipe(distinctUntilChanged()))
    .subscribe(
      ([isAuthenticated, notificationChannelReady]) => {
        setAuthenticated(isAuthenticated && notificationChannelReady);
      },
      err => console.error(err),
    );
    return () => sub.unsubscribe();
}, [context.login, context.notificationChannel, addSubscription, setAuthenticated]);
```
`LoginService`:
```
 constructor() {
    ...
    this.token.next(this.replaceWithCachedToken(''));
    this.authenticated.next(this.isAuthenticated());
  }
...
  private replaceWithCachedToken(defaultToken: string) {
    const cachedToken = this.getCachedToken();
    return (!!cachedToken) ? cachedToken : defaultToken;
  }
...
  isAuthenticated() : boolean {
    return !!this.getCachedToken();
  }
...
  private getCachedToken() : string {
    const token = sessionStorage.getItem('token');
    return (!!token) ? token : '';
  }
```

jan-law: Update: I'm going to try to change the way `LoginService` emits changes to `authenticated`. `loginService` can emit one event when the user logs in and one event when the user logs out. Then `routes` and `AppLayout` can subscribe to those messages instead of relying on the value of the replay subject `authenticated` at that moment.

jan-law:
When the `Login` component renders, it sends an auth request to check for the auth method, the messaging server connects to the client and then disconnects when it notices an auth failure.
```
Sep 08, 2021 2:50:39 PM io.cryostat.core.log.Logger info
INFO: Outgoing WS message: {"meta":{"category":"WS_CLIENT_ACTIVITY","type":{"type":"application","subType":"json"},"serverTime":1631112639},"message":{"10.0.2.100:46278":"connected"}}
Sep 08, 2021 2:50:39 PM io.cryostat.core.log.Logger info
INFO: Disconnected remote client 10.0.2.100:46278 due to authentication failure
Sep 08, 2021 2:50:39 PM io.cryostat.core.log.Logger info
INFO: Outgoing WS message: {"meta":{"category":"WS_CLIENT_ACTIVITY","type":{"type":"application","subType":"json"},"serverTime":1631112639},"message":{"10.0.2.100:46278":"auth failure"}}
Sep 08, 2021 2:50:39 PM io.cryostat.core.log.Logger info
INFO: Disconnected remote client 10.0.2.100:46278
Sep 08, 2021 2:50:39 PM io.cryostat.core.log.Logger info
INFO: Outgoing WS message: {"meta":{"category":"WS_CLIENT_ACTIVITY","type":{"type":"application","subType":"json"},"serverTime":1631112639},"message":{"10.0.2.100:46278":"disconnected"}}
Sep 08, 2021 2:50:40 PM io.cryostat.core.log.Logger warn
WARNING: Exception thrown
io.vertx.ext.web.handler.impl.HttpStatusException: Unauthorized
```

If I quickly login, then logout, then login again, then logout again, the Cryostat web server will send a web socket error saying `INFO: Dropping remote client 10.0.2.100:46294 due to too many concurrent connections`.  Logging in one more time will trigger the server to repeatedly drop the new connection, causing the web client to repeatedly auto-retry the login `/auth` request.

Since there may be a slight delay in between processing the auth request to disconnecting clients, and `Max concurrent WebSocket connections: 2`, is there anything I can do on the client side to properly disconnect the web client before logging in again?

```
Sep 08, 2021 2:50:46 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:46216): POST /api/v1/auth 401 1ms
Sep 08, 2021 2:50:54 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:46218): OPTIONS /api/v1/auth 200 0ms
Sep 08, 2021 2:50:54 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:46216): POST /api/v1/auth 200 0ms
Sep 08, 2021 2:50:54 PM io.cryostat.core.log.Logger info
INFO: Dropping remote client 10.0.2.100:46286 due to too many concurrent connections
Sep 08, 2021 2:50:54 PM io.cryostat.core.log.Logger info
INFO: Outgoing WS message: {"meta":{"category":"WS_CLIENT_ACTIVITY","type":{"type":"application","subType":"json"},"serverTime":1631112654},"message":{"10.0.2.100:46286":"dropped"}}
Sep 08, 2021 2:50:55 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:46216): POST /api/v1/auth 200 0ms
Sep 08, 2021 2:50:55 PM io.cryostat.core.log.Logger info
INFO: Dropping remote client 10.0.2.100:46288 due to too many concurrent connections
Sep 08, 2021 2:50:55 PM io.cryostat.core.log.Logger info
INFO: Outgoing WS message: {"meta":{"category":"WS_CLIENT_ACTIVITY","type":{"type":"application","subType":"json"},"serverTime":1631112655},"message":{"10.0.2.100:46288":"dropped"}}
Sep 08, 2021 2:50:57 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:46216): POST /api/v1/auth 200 0ms
Sep 08, 2021 2:50:57 PM io.cryostat.core.log.Logger info
INFO: Dropping remote client 10.0.2.100:46290 due to too many concurrent connections
Sep 08, 2021 2:50:57 PM io.cryostat.core.log.Logger info
INFO: Outgoing WS message: {"meta":{"category":"WS_CLIENT_ACTIVITY","type":{"type":"application","subType":"json"},"serverTime":1631112657},"message":{"10.0.2.100:46290":"dropped"}}
```

andrewazores:
It sounds like you might need to add some logic to the `NotificationChannel.service` so that it can handle disconnect/reconnect nicely. Right now it attempts to open a WebSocket connection with `combineLatest(notificationsUrl, this.login.getToken(), this.login.getAuthMethod())`, so if the token and/or authMethod change it will cause a new connection attempt to be made. With this new user session flow these things can change much more often over the lifecycle of the application, so some better websocket handling is needed.

Specifically, you'll probably want to close the websocket (`this.ws.complete()` in the service code) on connection failures and probably on logout as well, and open a new connection again once there is a new login request from the user and the authMethod/token are non-empty, etc.

jan-law:
@andrewazores Thanks for your help so far, this PR is ready for review

andrewazores:
Seems to behave as expected when using the BasicAuthManager. With NoopAuthManager, the login screen appears but there is also a blank user avatar icon in the masthead. This seems okay for now, since the backend doesn't really emulate an actual auth challenge login flow when using the NoopAuthManager. Once the backend does properly support user sessions we can do something better here.

I did twice see websocket repeated websocket connection failures with error notifications appearing after a page refresh, but I'm not sure how to reproduce that. I'll see if I can figure it out.

src/app/AppLayout/AppLayout.tsx
```diff
@@ -165,6 +167,42 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
   const handleAboutModalToggle = () => {
     setAboutModalOpen(!aboutModalOpen);
   };
+
+  React.useEffect(() => {
+    const sub = serviceContext.login.loggedIn().subscribe(
+      () => setShowUserIcon(true)
+    );
+    return () => sub.unsubscribe();
+  }, [serviceContext, serviceContext.login]);
```
andrewazores, line 176: I think `setShowUserIcon` should also be in the dependencies array here.

src/app/AppLayout/AppLayout.tsx
```diff
@@ -165,6 +167,42 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
...
+
+  React.useEffect(() => {
+    const sub = serviceContext.login.loggedOut().subscribe(
+      () => setShowUserIcon(false)
+    );
+    return () => sub.unsubscribe();
+  }, [serviceContext, serviceContext.login]);
```
andrewazores, line 183: I think `setShowUserIcon` should also be in the dependencies array here.

src/app/AppLayout/AppLayout.tsx
```diff
@@ -165,6 +167,42 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
...
+
+  const handleLogout = () => serviceContext.login.setLoggedOut();
+
+  const handleUserInfoToggle = () => setShowUserInfoDropdown(!showUserInfoDropdown);
```
andrewazores, line 187: Please wrap these two functions with `React.useCallback`

src/app/AppLayout/AppLayout.tsx
```diff
@@ -165,6 +167,42 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
...
+
+  const userInfoItems = [
+    <DropdownGroup key={0}>
+      <DropdownItem onClick={handleLogout}>Logout</DropdownItem>
+    </DropdownGroup>
+  ];
+
+  const UserInfoToggle = (
+    <DropdownToggle onToggle={handleUserInfoToggle} toggleIndicator={CaretDownIcon}>
+      {username ? username : <UserIcon color="white" size="sm"/>}
```
andrewazores, line 202: Just a style thing, feel free to ignore it, but since the empty string is falsy, this could be shortened to `{ username || <UserIcon /> }`

src/app/Login/Login.tsx
```diff
@@ -46,60 +46,60 @@ import { Base64 } from 'js-base64';
...
         if (v) {
-          onLoginSuccess();
+          serviceContext.login.setLoggedIn();
         } else if (userSubmission) {
           notifications.danger('Authentication Failure', `${authMethod} authentication failed`);
         }
       })
     );
-  }, [serviceContext, serviceContext.api, addSubscription, onLoginSuccess, notifications, authMethod]);
+  }, [serviceContext, addSubscription, notifications]);
```
andrewazores, line 72: Add `serviceContext.login`

src/app/Shared/Services/Login.service.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  getAuthMethod(): Observable<string> {
+    return this.authMethod.asObservable();
+  }
+
+  getUsername(): Observable<string> {
+    return this.username.asObservable();
+  }
+
+  isAuthenticated(): boolean {
```
andrewazores, line 122: It might be better here to return an `Observable<boolean>`. You can just use `of` for now.

src/app/Shared/Services/NotificationChannel.service.tsx
```diff
@@ -105,8 +109,17 @@ export class NotificationChannel {
         },
         (err: any) => this.logError('Notifications URL configuration', err)
       );
+
+    combineLatest(this.login.loggedOut())
```
andrewazores, line 113: This can be unwrapped/the `combineLatest` removed, since there is only one inner observable.

src/app/AppLayout/AppLayout.tsx
```diff
@@ -173,18 +173,22 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
...
+  }, [serviceContext, serviceContext.login, setShowUserIcon]);
 
-  const handleLogout = () => serviceContext.login.setLoggedOut();
+  const handleLogout = React.useCallback(() =>
+    serviceContext.login.setLoggedOut(),
+    [serviceContext.login]);
 
-  const handleUserInfoToggle = () => setShowUserInfoDropdown(!showUserInfoDropdown);
+  const handleUserInfoToggle = React.useCallback(() =>
+    setShowUserInfoDropdown(!showUserInfoDropdown),
```
andrewazores, line 190: I think in this case, when toggling state using a React hook, it's better to do: `setShowUserInfoDropdown(v => !v)` - ie pass a mutation function to the state setter, rather than read the actual state "externally" from the setter and modify it from there.

andrewazores:
When I use the BasicAuthManager and first log in to the web-client, then go to Recordings and select a target, only the Active Recording List is shown (Archives are disabled). If I refresh the page, the Archived recordings tab appears. I can reproduce this by logging out, closing the tab, re-opening the tab, and then signing back in and going to Recordings.

src/app/Shared/Services/Login.service.tsx
jan-law: If I use `of()` to convert `isAuthenticated` to an Observable, this useState hook would require an initial boolean value. 
AppLayout.tsx, line 76
```
const [showUserIcon, setShowUserIcon] = React.useState(serviceContext.login.isAuthenticated());
```
I ended up adding a custom hook to update the useState hook whenever the value of `isAuthenticated` is updated. Please let me know what you think.
```
const showUserIcon = useObservable(serviceContext.login.isAuthenticated());
```

src/app/utils/useObservable.ts
```diff
@@ -0,0 +1,53 @@
...
+ * SOFTWARE.
+ */
+
+import * as React from 'react';
+
+export function useObservable(observable) {
+  const [value, setValue] = React.useState(observable && observable.value);
+
+  React.useEffect(() => {
+    if (!observable || !observable.subscribe) {
```
andrewazores, line 45: I think this check wouldn't be necessary if the `observable` was type-checked - the compiler would be able to do this assertion at compile-time, rather than delegating it to runtime here.

```
function useObservable(observable: Observable<any>): Function {
...
}
```

Unfortunately though, `observable.value` will fail to compile, since not every Observable actually has a `value` property. I think this would be broken in many cases anyway, however - what if I pass in an Observable that has no `.value` property, and which emits `string` or some other non-boolean type? `observable && observable.value` will evaluate to `false`, but the Observable I passed in emits `string`s, so the local state in this hook has some messy type semantics.

src/app/Shared/Services/Login.service.tsx
andrewazores: I think defaulting it to `false` at first is fine, and then updating it as needed once the actual value is available. Since that's currently just coming from sessionStorage it's local and synchronous and so the update can be done very quickly, but in the future when we have server-side sessions, it will have to be asynchronous and defaulting to `false` and waiting for a network request to complete would be required anyway.

src/app/utils/useObservable.ts
jan-law: Resolving this comment since I removed the `useObservable` hook

andrewazores (dismissed):

jan-law:
Changing this PR to a draft while I add a checkbox to optionally save either basic credentials or bearer token

jan-law:
How can I test this branch on Openshift by creating a custom cryostat image? I checked out the `jan-law/user-session` branch in `cryostat/web-client` and ran `mvn clean package`. Not sure where to go from here.

andrewazores:
After that, you should have your custom-built image in your local podman registry (on your local machine). `podman images` should reflect this, and the `:latest` on your machine is what `smoketest.sh` will pick up and run by default. Do that, and once you've verified that the image contains the changes you expected, do ex. `podman tag quay.io/cryostat/cryostat:latest quay.io/jan-law/cryostat:my-custom-image` to give a new and different tag to that same image. Then `podman push quay.io/jan-law/cryostat:my-custom-image` to push that tagged image to quay. Go to quay.io and ensure your cryostat image repo is public. Then, start CRC or whatever other OpenShift instance you will use. Do `cd cryostat-operator ; DEPLOY_NAMESPACE=$(oc project -q) make cert_manager deploy` to get the Operator running in your namespace, then do `oc edit deployment cryostat-operator-controller-manager && sleep 30 && make create_cryostat_cr`. After `oc edit` you will be presented with your `$EDITOR` (or `$VISUAL` maybe?) and the YAML definition of the operator deployment - go to line ~41 and change the `RELATED_IMAGE_CORE` value to `quay.io/jan-law/cryostat:my-custom-image`, then save and exit the editor. The operator will then deploy this tagged image when it sees the Cryostat CR exists and spins up a Cryostat deployment in your namespace.

ebaron:
You can skip the step where you need to edit the Cryostat custom resource by running `make deploy CORE_IMG=quay.io/jan-law/cryostat:my-custom-image` instead of just `make deploy`. Might be a little easier/faster.

andrewazores:
Rebase please

andrewazores:
Just tested it with the typical smoketest/webpack-dev-server setup and it looks and works great for both Basic auth and no auth. I'll try it out soon in OpenShift with Bearer auth.

One note I have is that the new Login.service holds state for both a username and a token, but only one of these will ever be in use at a time depending on the auth configuration of the server. Could you find a way to refactor the service and the components that access it so that there is just one piece of stored state that is interpreted as either a username or a token depending on the detected auth method? I'm fine with this being done as a follow-up PR, in fact I would prefer that since it will make the review a bit easier.

jan-law:
Sounds good. I was also wondering if there's a way to display a username from the OpenShift auth implementation. We can continue this discussion in the follow up PR

andrewazores:
I do think we can add something in the backend to query the OpenShift auth server and get the username back from the Bearer token:

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#tokenreview-request-0

We already use TokenReviews to check if the user's supplied token is valid in cases where the user does not require any additional resource permissions (ex. create recordings) - so we could create ex. `POST /api/v2/auth` that returns information about the requesting user when given their token, with just a couple of small hooks into the auth manager implementations to supply this info.

PR #273, opened 2021-09-03
fix(events): reset event template search filter when switching targets

### LINKED ISSUE #264: Event template search/filter breaks when switching between targets
Similar to #263 

To reproduce:
- Select the `9093` application
- Navigate to Events/Event Templates
- Click the Filter text box for Templates and enter "prof"
- Observe that the list is updated to show only templates matching this string (ex. Profiling)
- Select the `quarkus-run.jar` application

Current behaviour:
The templates list is broken. The filter box contains no text, but the templates table/list still contains only the Profiling item. Clicking in the filter box and entering any text updates the filter/resets the view, ex. pressing 'space' resets the view to show all items.

Expected behaviour:
If the filter box is empty, there should be no filtering applied, and all of the templates should be present in the list.
Switching between targets should reset both the filter box and the table state.

hareetd:
Fixes #264

### DIFF
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -115,8 +115,10 @@ export const EventTemplates = () => {
 
   React.useEffect(() => {
     addSubscription(
-      context.target.target().subscribe(refreshTemplates)
-    );
+      context.target.target().subscribe(() => {
+        setFilterText(filterText => filterText = '');
+        refreshTemplates();
+      }));
   }, [context, context.target, addSubscription, refreshTemplates]);
 
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/app/Events/EventTemplates.tsx
```diff
@@ -115,8 +115,10 @@ export const EventTemplates = () => {
 
   React.useEffect(() => {
     addSubscription(
-      context.target.target().subscribe(refreshTemplates)
-    );
+      context.target.target().subscribe(() => {
+        setFilterText(filterText => filterText = '');
```
andrewazores, line 119: Would simply `setFilterText('')` also work here?

hareetd: It does, thanks

andrewazores (approved):

PR #283, opened 2021-09-14
feat(notifications): Sort notifications by category

### LINKED ISSUE #268: Notification Categories
Since #245 / #255, notifications go into a drawer with the most recent entry at the top, organized purely chronologically. It might be more useful for users if notifications were grouped by either notification category (as in WebSocket notifications channel "category" key), or by graphical notification level, and sorted chronologically within those categories.

jan-law:
Fixes #268

Groups notifications into three categories: "Completed Actions", "Network Info", and "Problems". 
Notifications confirming a user's action has succeeded will appear in "Completed Actions". "Network Info" contains any messages related to JVM target discovery and WebSocket disconnection. All warning and error notifications appear in the "Problems" category.

When opening the drawer, "Completed Actions" is expanded by default. If there are unread errors, the "Problems" tab is expanded instead. Only one category can be expanded at once.

### DIFF
--- a/src/app/Notifications/NotificationCenter.tsx
+++ b/src/app/Notifications/NotificationCenter.tsx
@@ -42,6 +42,7 @@ import { AlertVariant, Dropdown, DropdownItem, DropdownPosition, KebabToggle,
   NotificationDrawerListItemBody, NotificationDrawerListItemHeader, Text,
   TextVariants } from '@patternfly/react-core';
 import { Notification, NotificationsContext } from './Notifications';
+import { combineLatest } from 'rxjs';
 
 export interface NotificationCenterProps {
   onClose: () => void;
@@ -62,8 +63,16 @@ export const NotificationCenter: React.FunctionComponent<NotificationCenterProps
   ]);
 
   React.useEffect(() => {
-    const sub = context.notifications().subscribe(notifications => {
-      sortNotificationsByCategory(notifications);
+    const sub = combineLatest(context.actionsNotifications(), context.networkInfoNotifications(), context.problemsNotifications())
+    .subscribe(categories => {
+      let updatedDrawerCategories = [...drawerCategories];
+
+      categories.forEach((notifications, idx) => {
+        updatedDrawerCategories[idx].notifications = notifications;
+        updatedDrawerCategories[idx].unreadCount = countUnreadNotifications(notifications);
+      });
+
+      setDrawerCategories(updatedDrawerCategories);
     });
     return () => sub.unsubscribe();
   },[context, context.notifications]);
@@ -73,41 +82,6 @@ export const NotificationCenter: React.FunctionComponent<NotificationCenterProps
       setUnreadNotificationsCount(s.length)});
     return () => sub.unsubscribe();
   }, [context, context.unreadNotifications, setUnreadNotificationsCount]);
-
-  const sortNotificationsByCategory = notifications => {
-    let updatedActionsNotifications = [] as Notification[];
-    let updatedNetworkInfoNotifications = [] as Notification[];
-    let updatedProblemsNotifications = [] as Notification[];
-
-    notifications.forEach((msg) => {
-      if(isProblemNotification(msg)) {
-        updatedProblemsNotifications = [...updatedProblemsNotifications, msg];
-      } else if(isNetworkInfoNotification(msg)) {
-        updatedNetworkInfoNotifications = [...updatedNetworkInfoNotifications, msg];
-      } else {
-        updatedActionsNotifications = [...updatedActionsNotifications, msg];
-      }
-    });
-
-    const sortedNotifications = [updatedActionsNotifications, updatedNetworkInfoNotifications, updatedProblemsNotifications];
-    let updatedDrawerCategories = [...drawerCategories];
-
-    sortedNotifications.forEach((notifications, idx) => {
-      updatedDrawerCategories[idx].notifications = notifications;
-      updatedDrawerCategories[idx].unreadCount = countUnreadNotifications(notifications);
-    });
-
-    setDrawerCategories(updatedDrawerCategories);
-  };
-
-  const isProblemNotification = (msg) => (
-    (msg.variant === AlertVariant.warning) || (msg.variant === AlertVariant.danger)
-  );
-
-  const isNetworkInfoNotification = (msg) => (
-    (msg.category === 'WsClientActivity' || msg.category === 'TargetJvmDiscovery')
-      && (msg.variant === AlertVariant.info)
-  );
 
   const countUnreadNotifications = (notifications: Notification[]) => {
     return notifications.filter(n => !n.read).length;
--- a/src/app/Notifications/NotificationCenter.tsx
+++ b/src/app/Notifications/NotificationCenter.tsx
@@ -67,14 +67,14 @@ export const NotificationCenter: React.FunctionComponent<NotificationCenterProps
   React.useEffect(() => {
     const sub = combineLatest([context.actionsNotifications(), context.networkInfoNotifications(), context.problemsNotifications()])
     .subscribe(notificationLists => {
-      const updatedDrawerCategories = [...drawerCategories];
+      setDrawerCategories(drawerCategories => {
+        drawerCategories.map((category: NotificationDrawerCategory, idx) => {
 
-      updatedDrawerCategories.map((category: NotificationDrawerCategory, idx) => {
-        category.notifications = notificationLists[idx];
-        category.unreadCount = countUnreadNotifications(notificationLists[idx]);
+          category.notifications = notificationLists[idx];
+          category.unreadCount = countUnreadNotifications(notificationLists[idx]);
+        });
+        return drawerCategories;
       });
-
-      setDrawerCategories(updatedDrawerCategories);
     });
     return () => sub.unsubscribe();
   },[context, context.notifications, setDrawerCategories]);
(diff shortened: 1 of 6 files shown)


src/app/Notifications/NotificationCenter.tsx
```diff
@@ -37,35 +37,79 @@
...
-  const [unreadNotificationsCount, setUnreadNotificationsCount] = React.useState(0);
+  const [totalUnreadNotificationsCount, setTotalUnreadNotificationsCount] = React.useState(0);
   const [isHeaderDropdownOpen, setHeaderDropdownOpen] = React.useState(false);
+  const PROBLEMS_CATEGORY_IDX = 2;
+  const [drawerCategories, setDrawerCategories] = React.useState([
+    {title: "Completed Actions", isExpanded: true, notifications: [] as Notification[], unreadCount: 0},
+    {title: "Network Info", isExpanded: false, notifications: [] as Notification[], unreadCount: 0},
+    {title: "Problems", isExpanded: false, notifications: [] as Notification[], unreadCount: 0}
+  ] as NotificationDrawerCategory[]);
+  const unreadProblemsCount = drawerCategories[PROBLEMS_CATEGORY_IDX].unreadCount;
```
andrewazores, line 61: Won't this always initialize to 0?

src/app/Notifications/NotificationCenter.tsx
```diff
@@ -37,35 +37,79 @@
...
+    const sub = combineLatest([context.actionsNotifications(), context.networkInfoNotifications(), context.problemsNotifications()])
+    .subscribe(notificationLists => {
+      const updatedDrawerCategories = [...drawerCategories];
+
+      updatedDrawerCategories.map((category: NotificationDrawerCategory, idx) => {
+        category.notifications = notificationLists[idx];
+        category.unreadCount = countUnreadNotifications(notificationLists[idx]);
+      });
+
+      setDrawerCategories(updatedDrawerCategories);
```
andrewazores, line 77: Since this hook updates the `drawerCategories` state, but also depends on the previous state, I think it should use the `setDrawerCategories(prevDrawerCategories => { doSomething; return newDrawerCategories; });` form of the state setter function. It'll look a little unwieldy, but it should better ensure proper consistent state.

src/app/Notifications/NotificationCenter.tsx
```diff
@@ -37,35 +37,79 @@
...
+    setHeaderDropdownOpen(v => !v);
+  }, [setHeaderDropdownOpen]);
+
+  const handleToggleExpandCategory = React.useCallback((categoryIdx) => {
+    const updatedDrawerCategories = [...drawerCategories];
+    updatedDrawerCategories.map((category: NotificationDrawerCategory, idx) => {
+        category.isExpanded = (idx === categoryIdx) ? !category.isExpanded : false
+      });
+
+    setDrawerCategories(updatedDrawerCategories);
```
andrewazores, line 98: Same here regarding depending on the old state to update the new state.

src/app/Notifications/NotificationCenter.tsx
```diff
@@ -37,35 +37,79 @@
...
 
-  const handleToggleDropdown = () => {
-    setHeaderDropdownOpen(!isHeaderDropdownOpen);
-  };
+    const updatedDrawerCategories = [...drawerCategories];
+    updatedDrawerCategories.map((category: NotificationDrawerCategory, idx) => {
+      category.isExpanded = (idx === PROBLEMS_CATEGORY_IDX) ? true : false;
+    });
+    setDrawerCategories(updatedDrawerCategories);
+  }, [setDrawerCategories, unreadProblemsCount]);
```
andrewazores, line 112: Here too. Since there is also the `count === 0` guard at the top, should this hook dependency array include the `drawerCategories` state item?

src/app/Notifications/Notifications.tsx
```diff
@@ -46,10 +46,18 @@ export interface Notification {
   key?: string;
   title: string;
   message?: string;
+  category?: string;
   variant: AlertVariant;
   timestamp?: number;
 }
 
+export interface NotificationDrawerCategory {
```
andrewazores, line 54: I would probably move this into the same file as the component where it's used - it doesn't seem quite related to this service, which is more about the pure information side of a notification and not so much about the presentation of them.

src/app/Notifications/Notifications.tsx
```diff
@@ -117,6 +146,20 @@ export class Notifications {
     }
     this._notifications$.next(this._notifications);
   }
+
+  private isActionNotification(n: Notification): boolean {
+    return (n.category !== 'WsClientActivity' && n.category !== 'TargetJvmDiscovery')
+      && (n.variant === AlertVariant.success || n.variant === AlertVariant.info);
```
andrewazores, line 152: This ends up being basically any notification than isn't a networkInfo or a problem, right? Could it be re-written in those terms?

andrewazores:
Looks and works great, despite my nitpicking comments :-)

As a side note, while testing this I notice that "Recording Created" appears twice - once as a success and once as an info with more details. "Recording Archived" appears once, as an info. It's probably a good thing that this notification grouping makes this kind of detail about the notifications more obvious/visible. I'll file a separate issue for that.

src/app/Notifications/NotificationCenter.tsx
jan-law: Originally I was using `unreadProblemsCount = drawerCategories[PROBLEMS_CATEGORY_IDX].unreadCount;`. I replaced this with `drawerCategories` to simplify this array.

src/app/Notifications/NotificationCenter.tsx
jan-law: Removed this line after replacing the useEffect dependency on line 110 from 'unreadProblemsCount` to `drawerCategories`

src/app/Notifications/Notifications.tsx
jan-law: The following edit prevents the notification drawer from displaying any notifications, including errors. Will keep looking into it.
```
private isActionNotification(n: Notification): boolean {
    return !this.isNetworkInfoNotification(n) && !this.isProblemNotification(n);
}
```

src/app/Notifications/NotificationCenter.tsx
```diff
@@ -37,35 +37,88 @@
...
-    const sub = context.notifications().subscribe(setNotifications);
+    const sub = combineLatest([context.actionsNotifications(), context.networkInfoNotifications(), context.problemsNotifications()])
+    .subscribe(notificationLists => {
+      setDrawerCategories(drawerCategories => {
+        drawerCategories.map((category: NotificationDrawerCategory, idx) => {
+
+          category.notifications = notificationLists[idx];
+          category.unreadCount = countUnreadNotifications(notificationLists[idx]);
+        });
+        return drawerCategories;
```
andrewazores, line 82:
```javascript
var a = [1,2,3];
a.map(v => v+1); // [2,3,4]
return a; // still [1,2,3]
```

I think this should be `return drawerCategories.map(...)`

src/app/Notifications/NotificationCenter.tsx
```diff
@@ -37,35 +37,88 @@
...
+  }, [setHeaderDropdownOpen]);
+
+  const handleToggleExpandCategory = React.useCallback((categoryIdx) => {
+    setDrawerCategories(drawerCategories => {
+      let newDrawerCategories = [...drawerCategories];
+      newDrawerCategories.map((category: NotificationDrawerCategory, idx) => {
+
+        category.isExpanded = (idx === categoryIdx) ? !category.isExpanded : false
+      });
+      return newDrawerCategories;
```
andrewazores, line 105: same

src/app/Notifications/NotificationCenter.tsx
```diff
@@ -37,35 +37,88 @@
...
+    }
+    setDrawerCategories(drawerCategories => {
+      drawerCategories.map((category: NotificationDrawerCategory, idx) => {
 
-  const handleToggleDropdown = () => {
-    setHeaderDropdownOpen(!isHeaderDropdownOpen);
-  };
+        category.isExpanded = (idx === PROBLEMS_CATEGORY_IDX) ? true : false;
+      });
+      return drawerCategories;
```
andrewazores, line 119: same

src/app/Notifications/Notifications.tsx
jan-law: `a.filter(this.isActionNotification)` was causing an `Uncaught TypeError: Cannot read properties of undefined (reading 'isNetworkInfoNotification')`. Replacing `a.filter(this.isActionNotification)`  with `a.filter(n => this.isActionNotification(n))` resolves the type error.

src/app/Notifications/NotificationCenter.tsx
```diff
@@ -37,35 +37,87 @@
...
-    const sub = context.notifications().subscribe(setNotifications);
+    const sub = combineLatest([context.actionsNotifications(), context.networkInfoNotifications(), context.problemsNotifications()])
+    .subscribe(notificationLists => {
+        setDrawerCategories(drawerCategories => {
+
+          return drawerCategories.map((category: NotificationDrawerCategory, idx) => {
+            category.notifications = notificationLists[idx];
+            category.unreadCount = countUnreadNotifications(notificationLists[idx]);
+            return category;
+        }) as unknown as NotificationDrawerCategory[];
```
andrewazores, line 82: What's up with this `as unknown as NotificationDrawerCategory[]`? That looks awkward.

jan-law: Whoops, that was leftover from when I was trying to fix a return type error. It turns out I was missing the `return category` line. I'll remove this.

src/app/Notifications/NotificationCenter.tsx
```diff
@@ -37,35 +37,87 @@
...
+    if(drawerCategories[PROBLEMS_CATEGORY_IDX].unreadCount === 0) {
+      return;
+    }
 
-  const handleToggleDropdown = () => {
-    setHeaderDropdownOpen(!isHeaderDropdownOpen);
-  };
+    setDrawerCategories(drawerCategories => {
+      return drawerCategories.map((category: NotificationDrawerCategory, idx) => {
+        category.isExpanded = (idx === PROBLEMS_CATEGORY_IDX) ? true : false;
```
andrewazores, line 116: The `? true : false` here is redundant

andrewazores (approved):

PR #288, opened 2021-09-16
fix(notifications): Fix inconsistent action notification categories

### LINKED ISSUE #284: Notification Categories are inconsistent between actions
As a side note, while testing this I notice that "Recording Created" appears twice - once as a success and once as an info with more details. "Recording Archived" appears once, as an info. It's probably a good thing that this notification grouping makes this kind of detail about the notifications more obvious/visible. I'll file a separate issue for that.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-web/issues/283#issuecomment-919535403_

jan-law:
Fixes #284 

Only one notification should appear when recordings are created. All notifications indicating that a user's action was successfully processed now display the green "success" message style.

>  the subscription to this notification category should be somewhere long-lived

The NotificationChannel now subscribes to all WebSocket messages related to updating recording state (ie recording created/archived/saved/deleted) so that these notifications are displayed even if users navigate away from the Recordings page.

### DIFF
--- a/src/app/Notifications/Notifications.tsx
+++ b/src/app/Notifications/Notifications.tsx
@@ -101,10 +101,12 @@ export class Notifications {
     );
   }
 
-  networkInfoNotifications(): Observable<Notification[]> {
+  backendStatusNotifications(): Observable<Notification[]> {
     return this.notifications()
     .pipe(
-      map(a => a.filter(this.isNetworkInfoNotification))
+      map(a => a.filter(n =>
+        (this.isWsClientActivity(n) || this.isJvmDiscovery(n))
+      ))
     );
   }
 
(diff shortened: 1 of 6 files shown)


src/app/Notifications/Notifications.tsx
```diff
@@ -145,7 +146,8 @@ export class Notifications {
   }
 
   private isNetworkInfoNotification(n: Notification): boolean {
-    return (n.category === 'WsClientActivity' || n.category === 'TargetJvmDiscovery')
+    return (n.category === NotificationCategory.WsClientActivity
+      || n.category === NotificationCategory.JvmDiscovery)
```
andrewazores, line 150: I think this function should at least be renamed, since JvmDiscovery doesn't seem like a "network info" or very related to WsClientActivity - but it would probably be better to split this into two functions.

andrewazores:
Rather than "backend" status, could we say "Cryostat Status"?

andrewazores (approved):

PR #289, opened 2021-09-16
feat(notifications): Show toast on user action success

### LINKED ISSUE #267: Notification Toasts
Since #245 / #255 , notifications are kept in a drawer and possibly displayed as toasts (danger/warning level). However, some other notifications at "lower" levels, like "info" or "success", may still be useful for the user to be displayed as toast notifications as well, such as the notification that appears when a recording has been successfully archived. This is because there is no other visual feedback/indication to the user that their action has succeeded, or even that the request was made or accepted.

jan-law:
Fixes #267 

Since the purpose of adding toast notifications is to confirm a user’s action was processed successfully and Patternfly recommends that all of those notifications should have the “success” message variant, I decided to change all “user action confirmed” notifications to use the “success” variant in #288. 

Then the alert group would just need to filter and display messages that have either error, warning, or success variants. 
Adding a default 8-second timeout also seemed appropriate since toasts will appear after completing user actions. Any unread errors will still appear in the notifications drawer.

### DIFF
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -48,7 +48,7 @@ import { AboutModal, Alert, AlertGroup, AlertVariant, AlertActionCloseButton,
 import { BellIcon, CogIcon, HelpIcon } from '@patternfly/react-icons';
 import { map } from 'rxjs/operators';
 import { matchPath, NavLink, useHistory, useLocation } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


andrewazores (dismissed):

andrewazores:
Rebase please

jan-law:
Decided to leave all the helper functions as private except for `isProblemsNotification`, which is referenced in `AppLayout`

andrewazores (approved):

Issue #296, opened 2021-09-24
Cryostat logo opens new tab when clicked


jan-law:
Clicking on the Cryostat logo in the top left corner will open a new tab presenting the login screen. Repeatedly clicking the logo generates more tabs.

aptmac:
Fixed by: https://github.com/cryostatio/cryostat-web/commit/9dd3871a5d1c83c2a586dfdbee38f6879cd16f59

PR #297, opened 2021-09-24
fix(AppLayout): remove target from logoProps to prevent opening a new tab

aptmac:
This one-liner address https://github.com/cryostatio/cryostat-web/issues/296

The target '_blank' opens a new tab when the logo in the top left is clicked, which is probably not desirable. Now it'll just reload the page in place.


jan-law (approved):
Looks good to me!

andrewazores (approved):
LGTM

PR #301, opened 2021-09-28
fix(login): do not re-attempt ws connection on auth fail

### LINKED ISSUE #300: Bad credentials can cause login loop
It looks like the looping behaviour comes from the web-client attempting to reconnect to the backend when the connection fails, even if that failure was due to "auth failure". The web-client tries again and the backend denies it again, and this simply repeats. The scenario I had in mind when implementing the websocket auto-reconnect was simple transient network failures, not auth denials, so this case of a persistent failure was just not considered. From all I am seeing, the backend is behaving as intended here (other than checking the wrong permissions, but that should have been taken care of by your #689).

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/687#issuecomment-929268067_

andrewazores:
Fixes #300

Detect WebSocket connection rejection reason status code. If the code indicates that the connection was rejected/closed because the authenticating user has insufficient permissions, do not attempt automatic reconnections.

This should fix the looping problem described in cryostatio/cryostat#687 , where a full OpenShift reduced-permissions user account is set up. A possibly easier test case is to run this web-client using the typical devserver setup against a Cryostat smoketest.sh, but where the Cryostat image has been modified to reject WebSocket subprotocol authentication. I have prepared one such image that can be used like:

```
CRYOSTAT_IMAGE=quay.io/andrewazores/cryostat:fail-ws-subproto CRYOSTAT_AUTH_MANAGER=io.cryostat.net.BasicAuthManager CRYOSTAT_CORS_ORIGIN=http://localhost:9000 CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true sh smoketest.sh
```

This image will immediately fail all websocket connection attempts, reportedly due to a failed permissions check, but pass basic auth checks otherwise according to normal Cryostat basic auth.

### DIFF
--- a/src/app/Login/Login.tsx
+++ b/src/app/Login/Login.tsx
@@ -38,6 +38,7 @@
 import * as React from 'react';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { NotificationsContext } from '../Notifications/Notifications';
+import { CloseStatus } from '@app/Shared/Services/NotificationChannel.service';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
 import { Card, CardBody, CardFooter, CardHeader, PageSection, Title } from '@patternfly/react-core';
 import { combineLatest, timer } from 'rxjs';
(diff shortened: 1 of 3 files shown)


ebaron (approved):
Fixes the looping issue for me on OpenShift

PR #305, opened 2021-09-28
fix(notifications): handle Error type messages

### LINKED ISSUE #304: Error notifications can cause React child component render failure
I wrote a fix to return a 401 error instead of 500 when the user has inadequate permissions, but it didn't affect the white screen issue. Looking in the browser log, I see this message:
```
Error: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=Error%3A%20Unauthorized&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    React 8
    unstable_runWithPriority scheduler.production.min.js:19
    React 5
    sub AppLayout.tsx:112
    RxJS 8
    notify Notifications.tsx:65
    danger Notifications.tsx:81
    queryForTargets Targets.service.tsx:92
    RxJS 59
    checkAuth Api.service.tsx:152
    RxJS 28
    checkAuth Login.tsx:66
    handleSubmit Login.tsx:79
    handleSubmit BearerAuthForm.tsx:57
    React 10
react-dom.production.min.js:209:186
    React 5
    unstable_runWithPriority scheduler.production.min.js:19
    React 4
    unstable_runWithPriority scheduler.production.min.js:19
    React 5
    sub AppLayout.tsx:112
    RxJS 8
    notify Notifications.tsx:65
    danger Notifications.tsx:81
    queryForTargets Targets.service.tsx:92
    RxJS 59
    checkAuth Api.service.tsx:152
    RxJS 28
    checkAuth Login.tsx:66
    handleSubmit Login.tsx:79
    handleSubmit BearerAuthForm.tsx:57
    React 10
```

Visiting the error decoder URL gives the following message: "Objects are not valid as a React child (found: Error: Unauthorized). If you meant to render a collection of children, use an array instead."

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat/issues/687#issuecomment-929564453_

andrewazores:
Fixes #304

The root cause of the bug is that services and components may attempt to display a notification with an Error object as the message to display to the user. This fails because the rendered child component (content of the graphical notification) must be a string. This is a bit tricky because the Notification data type specifies that the message property should be a string, however many/most places where an Error object may be obtained it is typed as 'any', which breaks some of the TS compiler's type-checking guarantees at compile time. To work around this, I apply a simple fix to the notifications service: do a runtime check of the message type, and if it isn't an actual string, apply "JSON.stringify" to serialize it to a string before it is displayed.

### DIFF
--- a/src/app/Notifications/Notifications.tsx
+++ b/src/app/Notifications/Notifications.tsx
@@ -46,7 +46,7 @@ export interface Notification {
   read?: boolean;
   key?: string;
   title: string;
-  message?: string;
+  message?: string | Error;
   category?: string;
   variant: AlertVariant;
   timestamp?: number;
@@ -63,6 +63,11 @@ export class Notifications {
     }
     notification.read = false;
     notification.timestamp = +Date.now();
+    if (notification.message instanceof Error) {
+      notification.message = JSON.stringify(notification.message, Object.getOwnPropertyNames(notification.message));
+    } else if (typeof notification.message !== 'string') {
+      notification.message = JSON.stringify(notification.message);
+    }
     this._notifications.unshift(notification);
     this._notifications$.next(this._notifications);
   }


ebaron:
No more white screen after this fix! There's 3 error notifications, which might not mean anything to users, but it works.

Screenshot 2021-09-28 at 17-33-07 Cryostat

ebaron (approved):

andrewazores:
I think that's to be expected at this point. The first one looks like it comes from the frontend's HTTP service, which automatically displays notifications on 500s. The second is from the component that made that request, trying to give more detail about the specific failure. And the last one would have been there normally due to the messaging service connection failure due to also lacking auths.

PR #308, opened 2021-09-29
chore(deps): upgrade all deps

### LINKED ISSUE #307: Dependencies are old
We're still using webpack v4, older versions of Patternfly and React, etc. These should all be upgraded.

andrewazores:
Fixes #307


jan-law:
I noticed the text font looks slightly different. There's a warning in the console:
image

andrewazores:
Hmm. The right side image on the About modal is also missing now.

andrewazores:
Fonts and About modal image seem to work the same as before the PR now.

jan-law (approved):
Looks good!

PR #310, opened 2021-09-30
fix(notifications): Fix notifications url request failure handling

### LINKED ISSUE #222: NotificationChannelService improperly handles clienturl request failures
https://github.com/cryostatio/cryostat/pull/496#issuecomment-854962881

hareetd:
Fixes #222 

Sample: 

image

### DIFF
--- a/src/app/Shared/Services/NotificationChannel.service.tsx
+++ b/src/app/Shared/Services/NotificationChannel.service.tsx
@@ -107,7 +107,7 @@ export class NotificationChannel {
 
     const notificationsUrl = fromFetch(`${this.apiSvc.authority}/api/v1/notifications_url`)
       .pipe(
-        switchMap(resp => {
+        concatMap(resp => {
           if (!resp.ok) throwError(new Error(resp.status + ' ' + resp.statusText));
           return from(resp.json());
         }),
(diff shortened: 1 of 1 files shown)


src/app/Shared/Services/NotificationChannel.service.tsx
```diff
@@ -107,10 +107,14 @@ export class NotificationChannel {
 
     const notificationsUrl = fromFetch(`${this.apiSvc.authority}/api/v1/notifications_url`)
       .pipe(
-        concatMap(resp => from(resp.json())),
+        concatMap(resp => {
+          if (resp.ok) return from(resp.json());
+          throw new Error(resp.status + ' ' + resp.statusText);
```
andrewazores, line 112: Does `resp` still have a useful response body at this point, even if `!resp.ok`? I think Cryostat is probably still sending some additional info about the failure here, like the stack trace or its backend "cause" message for the exception, so that would be useful to include in the error log and notification if we can get it.

hareetd: Here's what it looks like with the response body included:

Notification:
image

Window console:
image

src/app/Shared/Services/NotificationChannel.service.tsx
```diff
@@ -43,6 +43,7 @@ import { concatMap, distinctUntilChanged, filter, map } from 'rxjs/operators';
 import { Base64 } from 'js-base64';
 import * as _ from 'lodash';
 import { ApiService } from './Api.service';
+import { urlAlphabet } from 'nanoid';
```
andrewazores: This looks unused

andrewazores (approved):

PR #315, opened 2021-10-05
fix(targets): Deselect nonexistent target

### LINKED ISSUE #312: Stored target selection should be cleared if target no longer exists
Currently, the TargetSelect component stores the user's target selection into localStorage. This is restored later if the user refreshes or revisits the web application.

https://github.com/cryostatio/cryostat-web/blob/main/src/app/TargetSelect/TargetSelect.tsx#L81

There is a bug here where the user's selection is stored, the user leaves the application, the selected target disappears, and then the user returns to the application. When this happens, the TargetSelect component will restore a target selection that does not exist. This will lead to network errors (like "no route to host" or resolution failures) if the user navigates to another view, ex. the Active Recordings List, which will attempt to query for active recordings on the invalid restored target.

jan-law:
Fixes #312 

When a previously selected target disappears, the TargetSelect component will display `Select a target` instead of the disconnected target. As before, `localStorage` is updated whenever the user selects a new target.

### DIFF
--- a/src/app/TargetSelect/TargetSelect.tsx
+++ b/src/app/TargetSelect/TargetSelect.tsx
@@ -66,7 +66,10 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props) 
   const addSubscription = useSubscriptions();
 
   React.useEffect(() => {
-    const sub = context.targets.targets().subscribe(setTargets);
+    const sub = context.targets.targets().subscribe((targets) => {
+      setTargets(targets);
+      selectTargetFromCache(targets);
+    });
     return () => sub.unsubscribe();
   }, [context, context.targets, setTargets]);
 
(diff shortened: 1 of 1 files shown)


src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -79,12 +80,16 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props)
   React.useEffect(() => {
     try {
       const previouslyCachedTarget = getCachedTargetSelection();
-      context.target.setTarget(previouslyCachedTarget);
+      const targetStillExists = targets.some(target => _.isEqual(previouslyCachedTarget, target));
+
+      if(targetStillExists) {
+        context.target.setTarget(previouslyCachedTarget);
+      }
```
andrewazores, line 87: Would it make sense to add an `else` here that removes the entry from the cache?

jan-law: It looks like the `useEffect` hook gets triggered twice while the `targets` list is still empty.  Removing entries from the cache might remove a correct entry before `targets` is populated. I'll see if I can restructure the code to make it easier to read.
```
console.log("use effect triggered with target list: ", targets)
      if(targetStillExists) {
        context.target.setTarget(previouslyCachedTarget);
      }
```
image

jan-law: I moved the `setTarget` logic inside a callback function that only runs whenever the observable `Targets[]` changes, then added the `else` statement to remove the invalid cache entry

andrewazores (approved):

PR #321, opened 2021-10-08
fix(auth): Fix repeat auth calls on logout

### LINKED ISSUE #320: Do not retry login after logging out
The following condition `shouldRetryLogin` should only evaluate to true when the web-client has valid credentials and was temporarily disconnected from the backend. However, the code below evaluates to `true` before the user is logged in, where `ready.code` is undefined and `ready.ready` is false. `hasInvalidCredentials` then evaluates to `false` regardless of whether the `CloseStatus` indicated a protocol failure. `shouldRetryLogin` then becomes true, triggering repeated auth calls.

Login.tsx:
```
const hasInvalidCredentials = !!ready.code && ready.code === CloseStatus.PROTOCOL_FAILURE;
const shouldRetryLogin = (!hasInvalidCredentials && !ready.ready)
|| (!!token && ready.ready);
```

jan-law:
Fixes #320

### DIFF
--- a/src/app/Shared/Services/NotificationChannel.service.tsx
+++ b/src/app/Shared/Services/NotificationChannel.service.tsx
@@ -133,9 +133,7 @@ export class NotificationChannel {
             return;
           }
 
-          if(!token) {
-            return;
-          } else if (authMethod === 'Bearer') {
+          if (authMethod === 'Bearer') {
             subprotocol = `base64url.bearer.authorization.cryostat.${Base64.encodeURL(token)}`;
           } else if (authMethod === 'Basic') {
             subprotocol = `basic.authorization.cryostat.${token}`;


andrewazores:
I don't think it really matters what's displayed with the NoopAuthManager, so long as just clicking Submit lets the user in without having to enter some arbitrary faked credentials. Even better would be if NoopAuthManager led to the Login form being automatically dismissed since the default no-credentials case should count as valid authentication there, but that's just a nice-to-have. We aren't expecting anyone to use NoopAuthManager other than developers at this point anyway, and unless we decide to package up and support a "local installation" of Cryostat (either as a bare JVM process or as a local Podman container), I don't see that changing.

> I changed checkAuth('', 'Basic') to checkAuth('', '') in order to display the Basic Login page with their saved credentials until they press Login (instead of the above query automatically logging in the user)

How would this interact/behave with saved credentials for Bearer tokens? It's best to avoid making any assumptions or having some preferential treatment for one login form over the others, I think.

andrewazores:
Yea, unless we do some build-time hacks to strip it out to save a few bytes of storage space, the component would end up included in the shipped Cryostat image, but we would just never expect anyone to actually configure their Cryostat instance to trigger it to show up.

It's not that dissimilar to the downstream productized builds with the Operator only using Bearer auth in practice, but still containing the Basic auth form and code for handling Basic credentials.

PR #346, opened 2021-11-30
fix(snapshot): Cleanly handle empty snapshots

### LINKED ISSUE #345: Cleanly handle Snapshots taken when there is no source data
Front-end change corresponding to https://github.com/cryostatio/cryostat/issues/715

hareetd:
Fixes #345


andrewazores:
Code changes look fine, although I wonder about displaying a modal here. Seems like showing a standard notification would also make sense and be less intrusive. WDYT?

hareetd:
Agreed, a notification would also be more consistent with the rest of the UI alerts.

andrewazores (approved):

PR #350, opened 2021-12-08
feat(auth): Add OAuth redirect authentication

jan-law:
Related https://github.com/cryostatio/cryostat/issues/717

This PR redirects users to the OpenShift Container Platform login page and authenticates users with the resulting access token. A follow-up PR is needed to implement the logout feature.

### DIFF
--- a/src/app/Shared/Services/Login.service.tsx
+++ b/src/app/Shared/Services/Login.service.tsx
@@ -36,7 +36,7 @@
  * SOFTWARE.
  */
 import { Base64 } from 'js-base64';
-import { combineLatest, Observable, ObservableInput, of, ReplaySubject } from 'rxjs';
+import { combineLatest, Observable, ObservableInput, of, ReplaySubject, throwError } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
 import { catchError, concatMap, debounceTime, distinctUntilChanged, first, map, tap } from 'rxjs/operators';
 import { SettingsService } from './Settings.service';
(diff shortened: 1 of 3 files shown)


andrewazores:
I inverted the depends-on relationship between this PR and the Cryostat backend PR.

src/app/Shared/Services/Login.service.tsx
```diff
@@ -80,14 +80,19 @@ export class LoginService {
   }
 
   queryAuthMethod(): void {
+    console.log("querying auth method")
```
andrewazores, line 83: I think this can be removed, right?

src/app/Shared/Services/Login.service.tsx
```diff
@@ -182,12 +196,30 @@ export class LoginService {
...
+    // cached in the oauth server, thus redirecting users to the login page
+    this.queryAuthMethod();
   }
 
   setSessionState(state: SessionState): void {
     this.sessionState.next(state);
   }
 
+  private getTokenFromUrlFragment(): string {
+    var matches = location.hash.match(new RegExp('access_token'+'=([^&]*)'));
```
andrewazores, line 210: I think the regular expression here is worth extracting out to a constant. Also, this is basically just a style thing, but `new RegExp('access_token'+'=([^&]*)')` could simply be `/access_token'+'=([^&]*)/` as a regex literal.

Should the regex use anchors, since it's matching against the browser's URL location hash string already, which should be pretty well-defined/constrained?

andrewazores: Ah, or the hash contains both the `access_token` and the `token_type`, so perhaps anchoring doesn't make sense. Disregard that part.

andrewazores:
Rebase please

src/app/Shared/Services/Login.service.tsx
jan-law: Whoops, forgot to push my fixup.

andrewazores (approved):

PR #355, opened 2021-12-16
feat(auth): Add OAuth logout

jan-law:
Related https://github.com/cryostatio/cryostat/issues/717

In order to log out of the OAuth server, the frontend needs to make a POST request to the OAuth server's `/logout` endpoint, which removes the session cookie stored in the OAuth server. The backend will delete the access token and return the `/logout` url. Note that deleting the access token without removing the session cookie will cause the OAuth server to immediately login a user after logging out.

[OpenShift OAuth logout docs](https://docs.openshift.com/container-platform/4.9/authentication/managing-oauth-access-tokens.html#oauth-delete-tokens_managing-oauth-access-tokens)
[OpenShift OAuth endpoint for deleting access tokens](https://docs.openshift.com/container-platform/4.9/rest_api/oauth_apis/oauthaccesstoken-oauth-openshift-io-v1.html#oauthaccesstoken-oauth-openshift-io-v1)
[OpenShift console UI code explaining logout behaviour](https://github.com/openshift/console/blob/22c6951efe7c4bca87f3f934063b9f4dcb0a4058/frontend/public/module/auth.js#L90)

### DIFF
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -79,8 +79,6 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
   const [notifications, setNotifications] = React.useState([] as Notification[]);
   const [unreadNotificationsCount, setUnreadNotificationsCount] = React.useState(0);
   const [errorNotificationsCount, setErrorNotificationsCount] = React.useState(0);
-  const [token, setToken] = React.useState("");
-  const [authMethod, setAuthMethod] = React.useState("");
   const location = useLocation();
 
   React.useEffect(() => {
@@ -161,18 +159,8 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
   }, [serviceContext.target]);
 
   const handleLogout = React.useCallback(() => {
-      const sub = serviceContext.login.setLoggedOut(token, authMethod).subscribe();
+      const sub = serviceContext.login.setLoggedOut().subscribe();
       return () => sub.unsubscribe();
-  }, [serviceContext.login, token, authMethod]);
-
-  React.useEffect(() => {
-    combineLatest([serviceContext.login.getToken(), serviceContext.login.getAuthMethod()]).subscribe(
-    parts => {
-      const token = parts[0];
-      const method = parts[1];
-      setToken(token);
-      setAuthMethod(method);
-    });
   }, [serviceContext.login]);
 
   const handleUserInfoToggle = React.useCallback(() =>
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -53,6 +53,7 @@ import { SslErrorModal } from './SslErrorModal';
 import { AboutCryostatModal } from '@app/About/AboutCryostatModal';
 import cryostatLogoHorizontal from '@app/assets/logo-cryostat-3-horizontal.svg';
 import { SessionState } from '@app/Shared/Services/Login.service';
+import { combineLatest } from 'rxjs';
 
 interface IAppLayout {
   children: React.ReactNode;
@@ -78,6 +79,8 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
   const [notifications, setNotifications] = React.useState([] as Notification[]);
   const [unreadNotificationsCount, setUnreadNotificationsCount] = React.useState(0);
   const [errorNotificationsCount, setErrorNotificationsCount] = React.useState(0);
+  const [token, setToken] = React.useState("");
+  const [authMethod, setAuthMethod] = React.useState("");
   const location = useLocation();
 
   React.useEffect(() => {
@@ -157,9 +160,20 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
     return () => sub.unsubscribe();
   }, [serviceContext.target]);
 
-  const handleLogout = React.useCallback(() =>
-    serviceContext.login.setLoggedOut(),
-    [serviceContext.login]);
+  const handleLogout = React.useCallback(() => {
+      const sub = serviceContext.login.setLoggedOut(token, authMethod).subscribe();
+      return () => sub.unsubscribe();
+  }, [serviceContext.login, token, authMethod]);
+
+  React.useEffect(() => {
+    combineLatest([serviceContext.login.getToken(), serviceContext.login.getAuthMethod()]).subscribe(
+    parts => {
+      const token = parts[0];
+      const method = parts[1];
+      setToken(token);
+      setAuthMethod(method);
+    });
+  }, [serviceContext.login]);
 
   const handleUserInfoToggle = React.useCallback(() =>
     setShowUserInfoDropdown(v => !v),
--- a/src/app/Shared/Services/Login.service.tsx
+++ b/src/app/Shared/Services/Login.service.tsx
@@ -36,13 +36,12 @@
  * SOFTWARE.
  */
 import { Base64 } from 'js-base64';
-import { combineLatest, from, Observable, ObservableInput, of, ReplaySubject } from 'rxjs';
+import { combineLatest, Observable, ObservableInput, of, ReplaySubject } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
-import { catchError, concatMap, debounceTime, distinctUntilChanged, first, map, switchMap, tap } from 'rxjs/operators';
+import { catchError, concatMap, debounceTime, distinctUntilChanged, first, map, tap } from 'rxjs/operators';
 import { SettingsService } from './Settings.service';
-import { ApiV2Response } from './Api.service';
+import { ApiV2Response, HttpError } from './Api.service';
 import { TargetService } from './Target.service';
-import { delay } from 'lodash';
 
 export enum SessionState {
   NO_USER_SESSION,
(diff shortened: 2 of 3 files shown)


src/app/AppLayout/AppLayout.tsx
```diff
@@ -157,9 +160,20 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({children}) => {
     return () => sub.unsubscribe();
   }, [serviceContext.target]);
 
-  const handleLogout = React.useCallback(() =>
-    serviceContext.login.setLoggedOut(),
-    [serviceContext.login]);
+  const handleLogout = React.useCallback(() => {
+      const sub = serviceContext.login.setLoggedOut(token, authMethod).subscribe();
```
andrewazores, line 164: The `Login.service` already knows the `token` and `authMethod`, doesn't it? So the `setLoggedOut()` call could/should just use those already-known details, and the `AppLayout` component wouldn't need to hold these new pieces of state.

src/app/Shared/Services/Login.service.tsx
```diff
@@ -188,23 +187,66 @@ export class LoginService {
...
+        credentials: 'include',
+        mode: 'no-cors',
+        method: 'POST',
+        body: null,
+        headers: this.getAuthHeaders(token, method),
+      }).pipe(
+          map(response => {
+            this.resetSessionState();
+            this.navigateToLoginPage();
+            return true;
```
andrewazores, line 230: Should this be mapping `response => true` or perhaps `response => response.ok`?

The `resetSessionState()` and `navigaeToLoginPage()` calls would fit better stylistically in a `tap` than a `map`. Perhaps the `tap` should be after the map to boolean or even after the `catchError`, and these actions only performed if the request succeeded?

src/app/Shared/Services/Login.service.tsx
```diff
@@ -36,12 +36,13 @@
...
 import { Base64 } from 'js-base64';
-import { combineLatest, Observable, ObservableInput, of, ReplaySubject, throwError } from 'rxjs';
+import { combineLatest, from, Observable, ObservableInput, of, ReplaySubject } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
-import { catchError, concatMap, debounceTime, distinctUntilChanged, first, map, tap } from 'rxjs/operators';
+import { catchError, concatMap, debounceTime, distinctUntilChanged, first, map, switchMap, tap } from 'rxjs/operators';
 import { SettingsService } from './Settings.service';
 import { ApiV2Response } from './Api.service';
 import { TargetService } from './Target.service';
+import { delay } from 'lodash';
```
andrewazores, line 45: This looks unused

src/app/Shared/Services/Login.service.tsx
```diff
@@ -36,12 +36,13 @@
  * SOFTWARE.
  */
 import { Base64 } from 'js-base64';
-import { combineLatest, Observable, ObservableInput, of, ReplaySubject, throwError } from 'rxjs';
+import { combineLatest, from, Observable, ObservableInput, of, ReplaySubject } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
-import { catchError, concatMap, debounceTime, distinctUntilChanged, first, map, tap } from 'rxjs/operators';
+import { catchError, concatMap, debounceTime, distinctUntilChanged, first, map, switchMap, tap } from 'rxjs/operators';
```
andrewazores, line 41: `switchMap` also looks unused

src/app/Shared/Services/Login.service.tsx
jan-law: Since I realized `mode: no-cors` prevents the frontend from accessing the response data, I found some docs explaining how to use CORS with the OpenShift server. `config.openshift.io` lets you edit the spec `additionalCORSAllowedOrigins` ([APIServer docs](https://docs.openshift.com/container-platform/4.8/rest_api/config_apis/apiserver-config-openshift-io-v1.html)). 

When I tried editing the APIServer config manually ([OCP docs](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.4/html/security/allowing-javascript-based-access-api-server)), I can access the 200 `response.ok` status, but the `kube:admin` session cookie is still stored on the OAuth server. Pressing logout right now immediately logs you back in with a new token, regardless of whether I send `headers: this.getAuthHeaders(...)` with my request.

```
$ oc edit apiserver.config.openshift.io cluster

apiVersion: config.openshift.io/v1
kind: APIServer
metadata:
  annotations:
    include.release.openshift.io/ibm-cloud-managed: "true"
    include.release.openshift.io/self-managed-high-availability: "true"
    include.release.openshift.io/single-node-developer: "true"
    oauth-apiserver.openshift.io/secure-token-storage: "true"
    release.openshift.io/create-only: "true"
  creationTimestamp: "2021-12-17T21:24:58Z"
  generation: 4
  name: cluster
  resourceVersion: "1147467"
  uid: a6c97a3c-d651-48a8-80b0-3d10cfb58ef3
spec:
  additionalCORSAllowedOrigins:
  - cryostat-sample-default\.apps\.jalaw0\.lab\.upshift\.rdu2\.redhat\.com
  audit:
    profile: Default
```

src/app/Shared/Services/Login.service.tsx
```diff
@@ -188,23 +188,64 @@ export class LoginService {
...
+          mode: 'cors',
+          method: 'POST',
+          body: null,
+          headers: this.getAuthHeaders(token, method),
+        });
+      }),
+      concatMap(
+        response => {
+          if(response.status === 302) {
+            const redirectUrl = response.headers.get('X-Location') || '';
```
andrewazores, line 209: What actually happens here if the headers didn't contain the expected `X-Location`? This defaults to an empty string, and then we try a `fromFetch` on that, which will presumably fail. Would it be better to fail more explicitly by throwing an error if the expected header isn't present?

src/app/Shared/Services/Login.service.tsx
```diff
@@ -188,23 +188,64 @@ export class LoginService {
...
+            });
+          } else {
+            return of(response);
+          }
+        }),
+      concatMap(
+        response => {
+          if(response.ok) {
+            this.resetSessionState();
+            this.navigateToLoginPage();
```
andrewazores, line 223: I would still like to see these two calls pulled out into a `tap()`

src/app/Shared/Services/Login.service.tsx
```diff
@@ -188,23 +188,64 @@ export class LoginService {
...
+         }
+          return of(response.ok);
```
andrewazores, line 225: If this was just `return response.ok` then this could be wrapped in a `map` instead of a `concatMap` I think. Just a little simpler and less wrapping/unwrapping.

src/app/Shared/Services/Login.service.tsx
```diff
@@ -188,23 +188,64 @@ export class LoginService {
...
+        });
+      }),
+      concatMap(
+        response => {
+          if(response.status === 302) {
+            const redirectUrl = response.headers.get('X-Location') || '';
+            return fromFetch(redirectUrl, {
+              mode: 'cors',
+              method: 'POST',
+              body: null,
```
jan-law, line 216: My last commit doesn't seem to remove the kube:admin session cookie from the OAuth server because I'm granted a new token immediately after logging out. The same behaviour occurs whether or not I add `headers: this.getAuthHeaders()` in my request. Any ideas?

andrewazores: Was that working previously in this patch set? If so, perhaps you can do a `git bisect` to narrow down which specific change may have broken it.

jan-law: Hmm, [`913a065`](https://github.com/cryostatio/cryostat-web/pull/355/commits/913a065d4bac323ff63518caa2e5888ee95fbaa9) is the first broken commit. In this commit, I piped the `combineLatest` token and method into `setLoggedOut()`, changed `switchMap` to `concatMap`, and removed the auth headers from the redirect url post request. I tried reverting each of these changes individually and the behaviour is still broken. Does anything else look suspicious to you?

andrewazores: Strange. What if you check out/build `913a065` and manually test the workflow with your browser's devtools network traffic monitoring tools open? This should show you exactly what requests are made to Cryostat and to OAuth and what parameters and headers were sent, and what the responses were. Once you know what that flow looks like you can check out the previous commit and run the same test again and compare the traffic traces.

It's the `POST` to the Cryostat backend that is meant to delete that `kube:admin` token, right?

jan-law: > It's the POST to the Cryostat backend that is meant to delete that kube:admin token, right?

Yes, that's correct. I'll let you know what I find from the traffic traces soon

jan-law: It turns out that a cookie (presumably the session cookie?) gets sent with the /logout request in the working commit using `credentials: 'include'`. I had removed `credentials: 'include'` in the broken commit because I didn't think I needed to send any additional information. I'll add that line back in

src/app/Shared/Services/Login.service.tsx
```diff
@@ -85,7 +85,7 @@ export class LoginService {
       });
   }
 
-  checkAuth(token: string, method: string, rememberMe = false): Observable<boolean> {
+  checkAuth(token: string, method: string, rememberMe = true): Observable<boolean> {
```
jan-law, line 88: Even with this change, refreshing the frontend generates an additional access token every time. The network tab shows an additional auth query somewhere that sends an empty token, which prompts the backend to send the login redirect url. I'll keep looking into this.

jan-law: Should be fixed now

andrewazores:
If I log in to my CRC `developer` account without granting it its required permissions, I get this:

image

The error messages appearing due to the failed requests (stemming from permissions denied) are expected. But, I don't have a user menu here and no obvious way to log out from this under-permissioned account.

jan-law:
I've added an empty state component that will only appear when an account doesn't have the required permissions. "Retry Login" should take you to the OpenShift Container Platform login page.
image

andrewazores (approved):
This is excellent, thank you and very well done.

PR #356, opened 2022-01-04
fix(notifications): Show newest unread notification first

### LINKED ISSUE #337: Notification toast displays oldest unread notification
If you complete an action that causes a green toast to pop up, eg create a recording, then wait for the toast to disappear automatically, the notification is still marked as 'unread'. The next time you complete an action that should display a notification, the oldest unread notification appears instead of the newest action notification.
image

jan-law:
Fixes #337


andrewazores (dismissed):
Seems fine as it is written, although I think it might be better (less fragile) for this component to explicitly look for the notification object with the greatest timestamp property. Then if the Notifications service changes how it internally stores the notification objects, or changes how it exposes them on its Observable, we still get the correct/expected behaviour from this component and the pop-up "toast".

andrewazores (approved):

PR #358, opened 2022-01-05
fix(targets): Validate connectUrl before submission

### LINKED ISSUE #348: Custom targets creation form should not allow submission without an entered URL
image

If the user has not entered any text in the required connection URL form field then the `Create` button should be disabled.

A nice enhancement would be to also perform some client-side validation of the URL. It must be either a JMX service URL (related: #347) or a simple `host:port` pair (ex. `localhost:0`, `myapp:1234`). If the entered text does not match this format then the submission button should be disabled *and* some helper text should appear to indicate to the user that their entered text is not a valid connection URL.

jan-law:
Fixes #348 

The create button is disabled until a user enters text into the Connect URL field. If the submission is non-empty and invalid, clicking `Create` will display more helper text instead of posting a request.

### DIFF
--- a/src/app/TargetSelect/CreateTargetModal.tsx
+++ b/src/app/TargetSelect/CreateTargetModal.tsx
@@ -36,7 +36,7 @@
  * SOFTWARE.
  */
 import { Target } from '@app/Shared/Services/Target.service';
-import { ActionGroup, Button, ButtonType, Form, FormGroup, Modal, ModalVariant, TextInput } from '@patternfly/react-core';
+import { ActionGroup, AlertVariant, Button, ButtonType, Form, FormGroup, Modal, ModalVariant, TextInput, ValidatedOptions } from '@patternfly/react-core';
 import * as React from 'react';
 
 export interface CreateTargetModalProps {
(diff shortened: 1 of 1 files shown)


src/app/TargetSelect/CreateTargetModal.tsx
```diff
@@ -47,8 +47,18 @@ export interface CreateTargetModalProps {
...
 export const CreateTargetModal: React.FunctionComponent<CreateTargetModalProps> = (props) => {
   const [connectUrl, setConnectUrl] = React.useState('');
+  const [validConnectUrl, setValidConnectUrl] = React.useState(ValidatedOptions.default);
   const [alias, setAlias] = React.useState('');
 
+  const jmxServiceUrlFormat = new RegExp([
+    /service:jmx:([a-z]+):/
+    ,/\/\/([a-z0-9-]+)\/jndi\/([a-z]+):/
+    ,/\/\/([a-z0-9-]+):([0-9]+)/
+    ,/\/([a-z]+)/g
```
jan-law, line 57: When validating JMX service URLs, how specific should the expression matching be? I've allowed any JMX connector and am not sure if we should require `jndi`.

andrewazores: It should be fairly lenient. We basically just want to prevent the user from trying to add a connection with an empty URL, or a URL that is obviously not a JMX service URL (like one with http:// scheme, or just some garbled text instead of a URL). Different connectors and transports should not be checked here since the frontend doesn't know what the backend actually supports.

IMO the objective for fixing #348 is just to help prevent users from doing something obviously wrong and possibly confusing, like seeing the "Create" button lit up as if it's clickable (implying their form submission is complete and valid), and then immediately receive an error response because they had forgotten to actually enter a URL.

andrewazores (approved):

PR #364, opened 2022-01-27
chore(devserver): set server authority to localhost

### LINKED ISSUE #363: devserver env uses `0.0.0.0` authority instead of `localhost`
Since https://github.com/cryostatio/cryostat/commit/05955511d , the Cryostat development testing setup binds to `localhost`, and not `0.0.0.0` as it used to. The `.env` that the -web devserver uses to set the server location should reflect this change so that the same origin can be used, and things like JWT download tokens can be requested with the correct host.

andrewazores:
Fixes #363

Noticed this one while looking into https://github.com/cryostatio/cryostat/issues/802 . When running the backend with a typical `CRYOSTAT_CORS_ORIGIN=http://localhost:9000 CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true sh smoketest.sh`, JWT token download requests fail because the URL of the requested resource was still using `http://0.0.0.0:8181`, but the server is listening on and expecting requests to use `http://localhost:8181` in this setup. To test, just try to download an event template using a devserver web client. Without this change it fails with an error about the request resourceUrl being invalid. With this change it should succeed.

This can also be worked around by setting the `CRYOSTAT_WEB_HOST` env var when starting the backend - setting it back to the old `0.0.0.0` to override the default `localhost` read out of the `pom.xml` sets it back to how it used to work, when `.env` was created.


jan-law (commented):
I noticed the same thing about the `CRYOSTAT_AUTHORITY` yesterday, except when I changed `CRYOSTAT_AUTHORITY` to use localhost, the successfully downloaded template is always named `TARGET` for me. Is it the same for you?

andrewazores:
Yes, that's just an unfortunate effect of using a CORS setup. When the client origin and the server origin differ, the browser ignores the client's hint as to what to name any downloaded file, and will insist on naming it whatever the last portion of the request URL was. I suppose this is to help users notice if some malicious client is masquerading as the frontend for some other service that the user thinks they're actually using.

jan-law (approved):
>  the browser ignores the client's hint as to what to name any downloaded file, and will insist on naming it whatever the last portion of the request URL was

Ah, I see, that's unfortunate, but not a big deal.

Can this PR also close https://github.com/cryostatio/cryostat/issues/802 then? The backend doesn't need any changes as far as I can tell.

andrewazores:
Oh - does this also fix the https://github.com/cryostatio/cryostat/issues/802 issue? I didn't think that it did. I was able to reproduce that one just running smoketest.sh without the CORS devserver setup.

jan-law:
My bad, https://github.com/cryostatio/cryostat/issues/802 is still an issue without the CORS devserver setup. With this PR though, the templates can download if you use `CRYOSTAT_CORS_ORIGIN=http://localhost:9000`.

PR #366, opened 2022-02-04
fix(archivedrecordings): use notifications to update state without query

### LINKED ISSUE #344: Archived recordings table view does not update on notification
There is a WebSocket Notification emitted when a recording is added or removed from the archives. This should trigger the Archived recordings table to update and add/remove the corresponding entry. A full refresh (GET query) should not be necessary - the state starts from a query when the view is first created, so just applying add/remove operations in sequence when received by notifications should be enough to keep the view in sync with the actual backend state.

andrewazores:
Fixes #344

### DIFF
--- a/src/app/Shared/Services/NotificationChannel.service.tsx
+++ b/src/app/Shared/Services/NotificationChannel.service.tsx
@@ -40,13 +40,12 @@ import { BehaviorSubject, combineLatest, Observable, Subject, timer } from 'rxjs
 import { fromFetch } from 'rxjs/fetch';
 import { webSocket, WebSocketSubject } from 'rxjs/webSocket';
 import { concatMap, distinctUntilChanged, filter } from 'rxjs/operators';
-import { Base64 } from 'js-base64';
 import * as _ from 'lodash';
 import { AuthMethod, LoginService, SessionState } from './Login.service';
-
+import { SavedRecording } from '@app/Shared/Services/Api.service';
 
 interface RecordingNotificationEvent {
-  recording: string;
+  recording: SavedRecording;
   target: string;
 }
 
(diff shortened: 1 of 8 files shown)


src/app/Shared/Services/NotificationChannel.service.tsx
```diff
@@ -93,24 +92,24 @@ export class NotificationChannel {
...
+    this.messages(NotificationCategory.RecordingArchived).subscribe(v => {
       const event: RecordingNotificationEvent = v.message;
-      notifications.success('Recording Stopped', `${event.recording} was stopped`);
+      notifications.success('Recording Archived', `${event.recording.name} was archived`);
     });
 
     this.messages(NotificationCategory.RecordingDeleted).subscribe(v => {
       const event: RecordingNotificationEvent = v.message;
-      notifications.success('Recording Deleted', `${event.recording} was deleted`);
+      notifications.success('Recording Deleted', `${event.recording.name} was deleted`);
```
jan-law, line 112: Deleting archived recordings displays the correct file name in the notification, but when I delete a running/stopped recording, the notification reads `undefined was deleted`.

jan-law (commented):
Besides my last comment, code looks good

src/app/Shared/Services/NotificationChannel.service.tsx
andrewazores: fixed

jan-law: Deleting the active recordings is working well now. When I tried to delete an archived recording, I got a notification reading `[object Object] was deleted`.

andrewazores: Ah. Turns out that deleting an active or an archived recording would each trigger a notification with the same category, but (for now - see also #368) different message contents format. I have updated this PR and the corresponding backend one to use more specific categories for these notifications, and to disambiguate the archived/active deletion.

hareetd (approved):
Tested it out with the corresponding backend and the updated notification categories are looking good. Archived recordings state is updating properly as well.

jan-law (approved):

Issue #349, opened 2021-11-30
Uploading/deleting Event Templates produces no Notifications


andrewazores:
`POST /api/v1/templates` and `DELETE /api/v1/templates/:templateName` endpoints both produce a WebSocket notification on success (as well as indicating success by the HTTP status code in the response). However, neither of these produce a graphical notification in the web-client.

At face value this comes down to the `NotificationChannel.service.tsx` simply not subscribing to the stream of messages with this category. However, this would be a recurring problem with more event categories that are added - and may already be an issue with other categories than the two for Event Templates. There is already some logic in the `Notifications.tsx` for determining whether a WebSocket message indicates some kind of "action" event (response to a user-initiated or automated API request), a "problem" (some API request failed), reflects WebSocket client activity, or indicates that a target JVM (dis)appeared. Rather than the `NotificationsChannel.service.tsx` explicitly opting in to some set of categories for which to display notifications, perhaps it should listen to all incoming messages with all categories, and display notifications for all of the ones that are "actions".

PR #367, opened 2022-02-04
feat(notifications): service handles and displays all incoming WS messages

andrewazores:
Related to #349

This refactors the explicit notification handling in the `NotificationChannel`, extracting out some configuration objects into a `Map` to redefine the existing notifications handling for things like WebSocket client activity, or recording start/stop/archive/delete actions.

This is then further extended by adding handling for notifications that have a category not specified in the above configuration map. The client will still attempt to display these notifications, but the display will include the raw message category from the WebSocket notification message, and the notification body will simply be the body field of the notification message, without any further contextual information, and likely with JSON formatting tokens present. This is just a catch-all handler in case some backend message is emitted that no frontend support was specifically implemented for. Ideally, it should never actually be triggered, because all message types are accounted for with the customized configuration objects.

### DIFF
--- a/src/app/Shared/Services/NotificationChannel.service.tsx
+++ b/src/app/Shared/Services/NotificationChannel.service.tsx
@@ -48,7 +48,7 @@ export enum NotificationCategory {
   WsClientActivity = 'WsClientActivity',
   JvmDiscovery = 'TargetJvmDiscovery',
   ActiveRecordingCreated = 'ActiveRecordingCreated',
-  ActiveRecordingStopped = 'ActiveRecordingStopped',
+  ActiveRecordingStopped = 'RecordingStopped',
   ActiveRecordingSaved = 'ActiveRecordingSaved',
   ActiveRecordingDeleted = 'ActiveRecordingDeleted',
   ArchivedRecordingCreated = 'ArchivedRecordingCreated',
--- a/src/app/Shared/Services/NotificationChannel.service.tsx
+++ b/src/app/Shared/Services/NotificationChannel.service.tsx
@@ -69,6 +69,12 @@ interface ReadyState {
 
 const messageKeys = new Map([
   [
+    // explicitly configure this category with a null mapper.
+    // This is a special case because we do not want to display an alert,
+    // the Targets.service already handles this
+    NotificationCategory.JvmDiscovery, null
+  ],
+  [
     NotificationCategory.WsClientActivity, {
       variant: AlertVariant.info,
       title: 'WebSocket Client Activity',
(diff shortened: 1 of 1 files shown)


src/app/Shared/Services/NotificationChannel.service.tsx
```diff
@@ -82,40 +145,29 @@ export class NotificationChannel {
...
-
-    this.messages(NotificationCategory.ArchivedRecordingCreated).subscribe(v => {
-      const event: RecordingNotificationEvent = v.message;
-      notifications.success('Recording Archived', `${event.recording.name} was uploaded into archives`);
+    messageKeys.forEach((value, key) => {
+      if (!value) {
+        return;
+      }
+      this.messages(key).subscribe((msg: NotificationMessage) => {
+        const message = value.body.call(this, msg);
```
hareetd, line 153: I'm confused on why the `this` context need to be passed into the `body` function? From its definition in the `NotificationMessageWrapper`, it should only need a single `NotificationMessage` parameter, no?

src/app/Shared/Services/NotificationChannel.service.tsx
```diff
@@ -82,40 +145,29 @@ export class NotificationChannel {
...
+        notifications.notify({ title: value.title, message, category: key, variant: value.variant })
+      });
     });
+    this._messages.pipe(
+      filter(msg => !messageKeys.has(msg.meta.category as NotificationCategory))
+    ).subscribe(msg => {
+      const category = NotificationCategory[msg.meta.category as keyof typeof NotificationCategory];
```
hareetd, line 160: Since this subscription handles messages which don't belong to any predefined field in the enum `NotificationCategory`, is this line simply saying "take whatever category `msg.meta.category` is and represent it as a new field in the enum `NotificationCategory`"?

src/app/Shared/Services/NotificationChannel.service.tsx
andrewazores: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call

That's just the `call` function's signature. The first parameter is the object that should be used as the `this` context, and any further parameters are passed after that.

But, you have a point that I don't actually need to use `call`, I should be able to just invoke the `value.body` as a function directly.

src/app/Shared/Services/NotificationChannel.service.tsx
andrewazores: > Since this subscription handles messages which don't belong to any predefined field in the enum NotificationCategory

Not quite - it handles messages for which there is no defined Mapper configuration object. There *should* be a 1:1 relation between NotificationCategory and Mappers, but this is meant to handle any cases where there is a category missing a mapper.

> is this line simply saying "take whatever category msg.meta.category is and represent it as a new field in the enum NotificationCategory"?

It's a typecast of the raw string to a NotificationCategory. If there is no matching enum member of the NotificationCategory equal to the raw string then this still ends up being the same raw string (`msg.meta.category`). Really nothing happens to the raw string value in the end, but this just lets me pretend it has been cast to an enum value so that I can compare it to the enum values (which are really just syntax sugar over string constants here) in the `if-else` for setting the alert variant.

src/app/Shared/Services/NotificationChannel.service.tsx
```diff
@@ -82,40 +145,29 @@ export class NotificationChannel {
...
+      if (!value) {
+        return;
+      }
+      this.messages(key).subscribe((msg: NotificationMessage) => {
+        const message = value.body(msg);
+        notifications.notify({ title: value.title, message, category: key, variant: value.variant })
+      });
     });
+    this._messages.pipe(
+      filter(msg => !messageKeys.has(msg.meta.category as NotificationCategory))
```
jan-law, line 158: Does this filter only allow notifications that have an invalid NotificationCategory? If it does, what kinds of notifications would meet the conditions on lines 163 and 165?

jan-law (commented):
Looks good otherwise

src/app/Shared/Services/NotificationChannel.service.tsx
```diff
@@ -72,6 +67,74 @@ interface ReadyState {
...
+  ],
+  [
+    NotificationCategory.ActiveRecordingCreated, {
+      variant: AlertVariant.success,
+      title: 'Recording Created',
+      body: evt => `${evt.message.recording} created in target: ${evt.message.target}`,
+    } as NotificationMessageMapper
+  ],
+  [
+    NotificationCategory.ActiveRecordingStopped, {
```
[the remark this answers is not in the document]
andrewazores: Ah, yea, that's a detail that leaked over from another branch I'm working on but haven't opened the PR for yet. Whoops.

src/app/Shared/Services/NotificationChannel.service.tsx
andrewazores: It filters notifications that aren't being handled by any of the defined messageKey "mappers". Those could still have known categories, if for some reason a mapper just wasn't provided. It is a little redundant since we do also have a defined mapper for those two categories on 163/165 already - but then again, we also have defined mappers for *every* category right now, so in practice we never expect this filter to actually come up with any results.

src/app/Shared/Services/NotificationChannel.service.tsx
hareetd: Ah, okay that clears things up.

jan-law (approved):

PR #369, opened 2022-02-09
fix(activerecordings): use notifications to update state without query

### LINKED ISSUE #368: Recording start/stop/delete notifications should not trigger full model refresh
See also https://github.com/cryostatio/cryostat-web/issues/344

In the Active Recordings Table component we listen for WebSocket notifications indicating that an active recording has been started, stopped, or deleted. If any of these occur then we trigger a full refresh of the table contents by firing a new HTTP GET request to the backend. Instead, the backend should include all relevant information about the recording that was started/stopped/deleted in the notification message, and the frontend should receive that message and use the recording information within to update the table state directly, without issuing an API request.

andrewazores:
Fixes #368

Much like #366, but for the active recordings table rather than archived.

### DIFF
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -45,7 +45,7 @@ import { Button, Checkbox, Text, Toolbar, ToolbarContent, ToolbarItem } from '@p
 import {  Tbody, Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
 import * as React from 'react';
 import { useHistory, useRouteMatch } from 'react-router-dom';
-import { merge, forkJoin, Observable } from 'rxjs';
+import { combineLatest, merge, forkJoin, Observable } from 'rxjs';
 import { concatMap, filter, first } from 'rxjs/operators';
 import { RecordingActions } from './RecordingActions';
 import { RecordingsTable } from './RecordingsTable';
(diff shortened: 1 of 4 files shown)


jan-law (commented):
Works well.

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -127,15 +124,63 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
...
+      combineLatest([
+        context.target.target(),
+        context.notificationChannel.messages(NotificationCategory.ActiveRecordingCreated),
+      ])
+      .subscribe(parts => {
+        const currentTarget = parts[0];
+        const event = parts[1];
+        if (currentTarget.connectUrl != event.message.target) {
+          return;
+        }
```
hareetd, line 137: Correct me if I'm wrong, but instead of simply subscribing directly to the `ActiveRecordingCreated` notification category, you've done a `combineLatest` with the current target to ensure that a new recording is only added to the table if the target in the notification matches the current target. Is this to ensure that in the (unlikely?) case that the notification is received _after_ the current target has changed, the active recordings table won't be updated with the wrong value?

hareetd: Ah, good point, I didn't think of other clients.

PR #370, opened 2022-02-11
fix(health): handle version separate from Grafana config

### LINKED ISSUE #365: `About` modal version string is undefined when Grafana is not configured
> Report generation and the minimal build setup is working well for me. Did you intend to leave the version string on the About page blank?

Looks like that's actually a bug on the frontend: https://github.com/cryostatio/cryostat-web/blob/16410a78be2a31b07e71739f8252b967504eb5f6/src/app/Shared/Services/Api.service.tsx#L112

Since there is no Grafana configuration, the `Api.service` fails the request, causing the "Grafana configuration not found" error popup. But this also prevents the version string from being correctly populated.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/809#issuecomment-1028321900_

andrewazores:
Fix a bug where missing jfr-datasource or Grafana dashboard URL configuration caused Cryostat version number to not be populated

Fixes #365

### DIFF
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -35,7 +35,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-import { from, Observable, ObservableInput, of, ReplaySubject, forkJoin, throwError, EMPTY } from 'rxjs';
+import { from, Observable, ObservableInput, of, ReplaySubject, forkJoin, throwError, EMPTY, shareReplay } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
 import { catchError, concatMap, first, map, mergeMap, tap } from 'rxjs/operators';
 import { Target, TargetService } from './Target.service';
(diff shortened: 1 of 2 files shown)


src/app/Shared/Services/Api.service.tsx
```diff
@@ -90,36 +90,45 @@ export class ApiService {
...
-            return throwError(() => new Error(message));
-          }}))
+        tap((resp: Response) => {
+          if (!resp.ok) {
+            window.console.error(resp);
+            this.notifications.danger('API /health request failed', resp.statusText);
+          }
+        }),
+        concatMap((resp: Response) => from(resp.json())),
+        shareReplay(),
```
jan-law, line 102: Is the `shareReplay()` needed here because the subscriptions on lines 104/108 subscribe to `health` after the first response gets emitted?

jan-law (commented):
Tested, looks good to me

src/app/Shared/Services/Api.service.tsx
hareetd: So if I'm understanding correctly, the separate subscriptions on lines 105 and 109 both receive and operate on the same cached `Observable` response JSON created on line 101, which is why they don't need to do the `(resp: Response) => from(resp.json())` conversion? And this allows you to extract the version information separately from the datasource/dashboard handling while limiting side effects (i.e. the original `fromFetch` API call) to only the amount needed?

andrewazores: You got it!

hareetd (approved):

PR #371, opened 2022-02-11
fix(notifications): add template notification handling

### LINKED ISSUE #349: Uploading/deleting Event Templates produces no Notifications

andrewazores:
Fixes #349

Since #367 a notification is already shown when uploading or deleting event templates, but this PR adds a specific handler so that the notification is displayed with a nicely formatted title and message.

### DIFF
--- a/src/app/Shared/Services/NotificationChannel.service.tsx
+++ b/src/app/Shared/Services/NotificationChannel.service.tsx
@@ -133,14 +133,14 @@ const messageKeys = new Map([
     NotificationCategory.TemplateCreated, {
       variant: AlertVariant.success,
       title: 'Template Created',
-      body: evt => `${evt.message.template} was created`
+      body: evt => `${evt.message.template.name} was created`
     } as NotificationMessageMapper
   ],
   [
     NotificationCategory.TemplateDeleted, {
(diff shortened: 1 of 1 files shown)


jan-law (approved):
Looks good!

hareetd (approved):

PR #378, opened 2022-02-17
feat(rules): add UI for displaying automated rules

### LINKED ISSUE #375: Automated Rules read-only view

andrewazores:
Fixes #375
Depends on https://github.com/cryostatio/cryostat/pull/837

### DIFF
--- a/src/app/Rules/Rules.tsx
+++ b/src/app/Rules/Rules.tsx
@@ -97,7 +97,7 @@ export const Rules = () => {
       title: 'Maximum Age',
       transforms: [
         info({
-          tooltip: 'The maximum age in seconds for data kept in the JFR recordings started by this rule.'
+          tooltip: 'The maximum age in seconds for data kept in the JFR recordings started by this rule. Values less than 1 indicate no limit.'
         })
       ],
     },
@@ -105,7 +105,7 @@ export const Rules = () => {
       title: 'Maximum Size',
       transforms: [
         info({
-          tooltip: 'The maximum size in bytes for JFR recordings started by this rule.'
+          tooltip: 'The maximum size in bytes for JFR recordings started by this rule. Values less than 1 indicate no limit.'
         })
       ],
     },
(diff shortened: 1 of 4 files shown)


src/app/Rules/Rules.tsx
```diff
@@ -0,0 +1,197 @@
...
+          tooltip: 'The maximum age in seconds for data kept in the JFR recordings started by this rule.'
+        })
+      ],
+    },
+    {
+      title: 'Maximum Size',
+      transforms: [
+        info({
+          tooltip: 'The maximum size in bytes for JFR recordings started by this rule.'
+        })
```
jan-law, line 109: Maybe you could indicate here that `Max Age: 0` and `Max Size: -1` correspond to unlimited age/size, to make it extra clear that `0` and `-1` are special values

src/app/Rules/Rules.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  React.useEffect(() => {
+    if (!context.settings.autoRefreshEnabled()) {
+      return;
+    }
+    const id = window.setInterval(() => refreshRules(), context.settings.autoRefreshPeriod() * context.settings.autoRefreshUnits());
+    return () => window.clearInterval(id);
+  }, []);
+
+  const handleSort = (event, index, direction) => {
```
jan-law, line 150: Could you explain how to sort the table? I tried adding a few rules with different names and clicking the arrow. The rules stayed in the same order.
image

jan-law (commented):
The rest of the code looks good, including the notifications.

src/app/Rules/Rules.tsx
```diff
@@ -0,0 +1,205 @@
...
+  };
+
+  const displayRules = React.useMemo(() => {
+    const { index, direction } = sortBy;
+    let sorted = [...rules];
+    if (typeof index === 'number') {
+      const keys = ['name'];
+      const key = keys[index];
+      sorted = rules
+        .sort((a: Rule, b: Rule): number => (a[key] < b[key] ? -1 : a[key] > b[key] ? 1 : 0));
```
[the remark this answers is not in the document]
andrewazores: Yea, it's a little bit unnecessary since there is only (currently) one sortable column here, but that's the gist of it. `a[key]` is equivalent to `a.name`. When the user clicks on the column header it emits an event that lets us track which column we should sort by, so that's what gives us our `index`. We map that to a property name of the `Rule` objects and sort on it.

jan-law (approved):

Issue #376, opened 2022-02-17
Automated Rules view for evaluating matchExpressions and indicating matched targets


andrewazores:

andrewazores:
Conceptually, something like the OpenShift Console developer topology view. The application graph nodes would highlight in different colours or otherwise indicate whether they match the entered matchExpression. Pushing to 2.2.0.

tthvo:
@andrewazores just came across this. How is this idea fitting the current topology view in #891?

andrewazores:
I think this will probably get pushed back again, to 2.4.0 .

Basically, if there is some visualization of the discovered targets, then there should also be a way to display that visualization that somehow highlights or filters targets by ones that are matched by a given matchExpression. In a Topology view that could look like highlighting the matches or fading out the non-matches. In a list view that could be reordering the elements so that all the matches are at the top, and all the non-matches are at the bottom and faded/disabled.

With that, it should then be possible to update this visualization dynamically as the given matchExpression changes. So if I have a blank matchExpression then every target is faded. If I type `true`, every target is highlighted.

This would be used in views like the Automated Rules or Stored Credentials creation forms, to help users to tailor a matchExpression to the subset of targets that make sense for their usecase. There could be an additional section in the UI that contains this target visualization and dynamically updates as the expression is modified.

Specifically for the Topology view that you are working on, I think some minimized version of that topology graph could be the visualization I'm describing above. The minimized version would not need to be interactable at all - no way to add new targets, no way to view details for a target, etc. Just simply the discovery view rendering as a visual aid component.

tthvo:
Just a question about template select in Rule create form here: Looks like we are fetching templates for the selected target. But when moving to a "topology graph view" of the target, users will be seeing many target of the same time. Should we get the common templates among targets the match expression matches?

![Screenshot from 2023-03-16 14-54-48](https://user-images.githubusercontent.com/68053619/225724840-6bcd571b-60d7-49e7-a5e0-61ca0c17c784.png)

andrewazores:
That sounds like it makes sense, just do a set intersection.

PR #381, opened 2022-02-25
feat(rules): add match expression evaluation component to rule creation form

andrewazores:
Related to #376

### DIFF
--- a/src/app/Rules/MatchExpressionEvaluator.tsx
+++ b/src/app/Rules/MatchExpressionEvaluator.tsx
@@ -44,11 +44,11 @@ import { TargetSelect } from '@app/TargetSelect/TargetSelect';
 import { NoTargetSelected } from '@app/TargetView/NoTargetSelected';
 import {CheckCircleIcon, InfoCircleIcon, QuestionCircleIcon, WarningTriangleIcon} from '@patternfly/react-icons';
 
-export interface TargetJsonProps {
+export interface MatchExpressionEvaluatorProps {
   matchExpression?: string;
 }
 
-export const TargetJson: React.FunctionComponent<TargetJsonProps> = (props) => {
+export const MatchExpressionEvaluator: React.FunctionComponent<MatchExpressionEvaluatorProps> = (props) => {
   const context = React.useContext(ServiceContext);
   const addSubscription = useSubscriptions();
   const [target, setTarget] = React.useState(undefined as Target | undefined);
--- a/src/app/Rules/MatchExpressionEvaluator.tsx
+++ b/src/app/Rules/MatchExpressionEvaluator.tsx
@@ -42,7 +42,7 @@ import { useSubscriptions } from '@app/utils/useSubscriptions';
 import { Target } from '@app/Shared/Services/Target.service';
 import { TargetSelect } from '@app/TargetSelect/TargetSelect';
 import { NoTargetSelected } from '@app/TargetView/NoTargetSelected';
-import {CheckCircleIcon, InfoCircleIcon, QuestionCircleIcon, WarningTriangleIcon} from '@patternfly/react-icons';
+import {CheckCircleIcon, ExclamationCircleIcon, InfoCircleIcon, WarningTriangleIcon} from '@patternfly/react-icons';
 
 export interface MatchExpressionEvaluatorProps {
   matchExpression?: string;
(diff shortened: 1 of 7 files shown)


src/app/Rules/MatchExpressionEvaluator.tsx
```diff
@@ -0,0 +1,141 @@
...
+          { JSON.stringify(body, null, 2) }
+        </CodeBlockCode>
+      </CodeBlock>
+    </>);
+  },[target]);
+
+  return (<>
+    <Stack hasGutter>
+      <StackItem>
+        <TargetSelect isCompact />
```
hareetd, line 115: Is it necessary to provide the `isCompact` prop to `TargetSelect` anymore since it looks like the component leaves the decision of whether it is compact or not to its parent component?

hareetd (commented):
It's working well and the UI looks great too! I just had one question about the code.

src/app/Rules/MatchExpressionEvaluator.tsx
andrewazores: Well spotted, thanks. I've removed the unused property.

src/app/Rules/MatchExpressionEvaluator.tsx
```diff
@@ -0,0 +1,141 @@
...
+          { JSON.stringify(body, null, 2) }
+        </CodeBlockCode>
+      </CodeBlock>
+    </>);
+  },[target]);
+
+  return (<>
+    <Stack hasGutter>
+      <StackItem>
+        <TargetSelect/>
```
jan-law, line 115: In the Create Recording form, the `TargetSelect` is part of the form submission. For the Automated Rules page, the`TargetSelect` and the generated example text isn't part of the form submission, even though you have to enter a match expression in the form to see the `statusLabel` validation. It would be nice if there was a clearer indication that the TargetSelect/examples are not part of the form submission, and that the validation `statusLabel` applies to the MatchExpression field on the left.

Also, to verify that my match expression applies to the right targets, I need to select each target in the `TargetSelect` dropdown and check each `statusLabel`. I think this is fine for this PR, but I was wondering what ideas you had to improve the target selection for multiple targets in the future?

src/app/Rules/MatchExpressionEvaluator.tsx
```diff
@@ -0,0 +1,141 @@
...
+      return;
+    }
+  }, [target, props.matchExpression]);
+
+  const statusLabel = React.useMemo(() => {
+    switch (valid) {
+      case ValidatedOptions.success:
+        return (<Label color="green" icon={<CheckCircleIcon />}>Expression Matches Selected Target</Label>);
+      case ValidatedOptions.warning:
+        return (<Label color="orange" icon={<QuestionCircleIcon />}>Expression Valid, No Match</Label>);
```
jan-law, line 88: For the 'Expression Valid, No Match' label, I originally interpreted this as 'the expression is valid and does not match any targets'. Maybe you could reword it as 'Expression Valid, Does Not Match Selected Target' or something similar?

src/app/Rules/MatchExpressionEvaluator.tsx
```diff
@@ -0,0 +1,141 @@
...
+        return (<Label color="grey" icon={<InfoCircleIcon />}>No Match Expression</Label>);
+    }
+  }, [valid]);
+
+  const exampleExpression = React.useMemo(() => {
+    let body: string;
+    if (!target || !target?.alias || !target?.connectUrl) {
+      body = 'true';
+    } else {
+      body = `target.alias == '${target?.alias}' || target.annotations.cryostat['PORT'] == ${target?.annotations?.cryostat['PORT']}`;
```
jan-law, line 101: Is it possible to remove the "" quotes from the example expression in case someone (like myself) tries to copy/paste the example into the Match Expression field? The Match Expression field input doesn't accept surrounding string quotes "".

jan-law (commented):
Code looks good and the examples look very nice. I just had a couple questions/suggestions to make the form easier to use.

src/app/Rules/MatchExpressionEvaluator.tsx
andrewazores: > In the Create Recording form, the TargetSelect is part of the form submission. For the Automated Rules page, theTargetSelect and the generated example text isn't part of the form submission, even though you have to enter a match expression in the form to see the statusLabel validation. It would be nice if there was a clearer indication that the TargetSelect/examples are not part of the form submission, and that the validation statusLabel applies to the MatchExpression field on the left.

Makes sense. Do you have any ideas on how to give a better conceptual link between the left side matchExpression form field and the right side component? Would adding a header/title to the card on the right help? Or perhaps that right side component should be hidden by default, and should only appear if some helper button/icon next to the matchExpression form field is clicked?

> Also, to verify that my match expression applies to the right targets, I need to select each target in the TargetSelect dropdown and check each statusLabel. I think this is fine for this PR, but I was wondering what ideas you had to improve the target selection for multiple targets in the future?

Definitely. This is something I want to find a better solution for, but I need to think about how to scale this across a large number of targets and have it work visually. Right now I'm thinking of having an entirely separate view, something like a Match Expression Wizard, that has some visualization like a list of target aliases with a checkbox to indicate whether the given expression matches the target. The list items would be expandable to reveal the whole JSON structure of the target like we see in this PR, so that the user can look at the details of all of the target's fields and use that to craft their expression. Collapsing all of the list items so that the view is just the list of checkboxes would give the user a way to evaluate whether the expression is matching all of the expected targets in a compact way. But I'm not completely happy with this solution either because it still means there will be a potentially very long list of targets and a lot of checkboxes to look at, so it will scale up to a point and then become very cumbersome again.

This separate view would probably be something like a modal that can be summoned from within the Create Rule form, and some other similar view could also be on the main Rules page so that users can immediately relate an existing Rule to the targets it applies to.

jan-law: The header and the validation within the match expression text field looks much better. I was going to suggest something similar.

I'd rather keep the match expression evaluator visible at all times. There's enough space on the page to do so, and keeping it as a separate item on the side helps to indicate that it's not part of the form.

jan-law (approved):

PR #383, opened 2022-03-01
feat(recordings): Add metadata recording labels

jan-law:
Related https://github.com/cryostatio/cryostat/issues/709

When creating recordings, users can add their own metadata with key-value label pairs. You can also edit labels for all active and archived recordings.

The labels are stored in the filesystem for now until we find a solution to manage the metadata when Cryostat is restarted (see https://github.com/cryostatio/cryostat/issues/709#issuecomment-1035314270)

image

### DIFF
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -266,9 +266,22 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
       };
 
       const RecordingDuration = (props) => {
-        const str = props.duration === 0 ? 'Continuous' : `${props.duration / 1000}s`
+        const str = props.duration === 0 ? 'Continuous' : `${props.duration / 1000}s`;
         return (<span>{str}</span>);
       };
+
+      const formattedLabels = React.useMemo(() => {
+        const labels = JSON.parse(props.recording.labels);
+        return (
+          <>
+            {Object.entries(labels).map(([k, v]) => (
+              <Text>
+                {k}={v}
+              </Text>
+            ))}
+          </>
+        );
+      }, [props.recording.labels]);
 
       return (
         <Tr key={`${props.index}_parent`}>
@@ -302,13 +315,16 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
           <Td key={`active-table-row-${props.index}_5`} dataLabel={tableColumns[3]}>
             {props.recording.state}
           </Td>
+          <Td key={`active-table-row-${props.index}_6`} dataLabel={tableColumns[4]}>
+            {formattedLabels || '-'}
+          </Td>
           <RecordingActions
             index={props.index}
             recording={props.recording}
             uploadFn={() => context.api.uploadActiveRecordingToGrafana(props.recording.name)}
           />
         </Tr>
-      )
+      );
     }, [
       props.duration,
       props.index,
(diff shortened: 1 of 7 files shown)


jan-law:
As of now, I don't have a UI for appending labels to recordings generated by automated rules. If you think it would be helpful to append labels to these recordings, let me know and I can add it in. 
I didn't add the labels form to the automated rules form because automated rules already have a "description" field where someone could add notes and auto-generated recordings already have the same name as that rule. You could still manually add labels to a specific auto-generated recording with the Recordings table actions if you wanted to.

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -302,13 +328,31 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
...
+                labels={rowLabels} 
+                setLabels={setRowLabels} 
+                usePatchForm={editingMetadata}
+                patchRecordingName={props.recording.name} 
+                onPatchSubmit={handleSubmitLabelPatch}
+                onPatchCancel={handleCancelLabelPatch}
+              />
+              : rowLabels.map(l => (
+                <Text>
+                  {l.key}={l.value}
```
andrewazores, line 343: This doesn't have to just be a `<Text>` element, right? If we can apply some Patternfly styling here I think it would really shine it up visually. https://www.patternfly.org/v4/components/label https://www.patternfly.org/v4/components/label-group . This would also help with the case that there are a lot of labels on a recording.

Also, in the case that the map is empty, could there be some default state here other than a simple empty table cell? Even just  a `-` character or something would be nice.

andrewazores:
Looks great - just a rebase please

andrewazores:
Oh, I guess this will need to be updated to match the backend change I just requested where the `labels` field is nested within a `metadata` field on the recording descriptor.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -587,4 +636,9 @@ export interface RecordingAttributes {
   events: string;
   duration?: number;
   options?: RecordingOptions;
+  labels?: string;
```
andrewazores, line 639: This field addition doesn't seem quite right - shouldn't it be a `Metadata` type to better reflect the actual API response objects?

Also, the metadata field is present in the API responses for both active and archived recordings, right? I think in that case the field should be added to the `ArchivedRecording` base interface rather than here.

jan-law: Are `RecordingAttribute`s used elsewhere besides passing the Create Recording fields to `createRecording(recordingAttributes: RecordingAttributes): Observable<boolean>`? 
I originally wanted to pass `recordingAttributes.labels` to `createRecording`, which I can change to `recordingAttributes.metadata`. I've already added the `metadata: Metadata` field in the `ArchivedRecording` interface.

andrewazores: It's only used in the `CreateRecording` and `CustomRecordingForm`. It just seems that it would be easier to reason about and maintain this code if the labels map is preserved as a map when passed around, rather than being stringified here - similar reasoning to what we discussed on the backend.

jan-law: WDYT about passing the `Metadata` object around in my last commit, where the object only gets stringified at `form.append('metadata', JSON.stringify(recordingAttributes.metadata))`?

andrewazores: Looks better, thanks

andrewazores (approved):

andrewazores:
rebase please

andrewazores (approved):

PR #385, opened 2022-03-04
fix(health): process health configured/available results

### LINKED ISSUE #384: `minimal` deployments display a "Grafana configuration not found" error
We build https://github.com/cryostatio/cryostat with a standard and a "minimal" build profile. The "minimal" profile originally meant that no web-client was included, but nowadays "minimal" for a Cryostat deployment means that the standard Cryostat image is deployed, but there is no accompanying jfr-datasource/grafana-dashboard. When an image is deployed in a minimal scenario and the web-client is visited in a browser however, the "Grafana configuration not found" error notification is displayed, despite the fact that this is actually the expected case for this scenario.

andrewazores:
Use new health "configured" status fields to determine if other
services components are expected to be available, in addition to the
existing "available" status fields indicating actual reachability for
configured components

Fixes #384

An easy way to test this and the associated backend fix is to use the backend devserver setup or smoketest script, and modify them to vary between a) datasource env var set and container started within pod, b) datasource env var set but container not started, c) datasource env var not set. This can be repeated for the dashboard alone or in conjunction. If both are configured and can be reached over the network then no notification should be displayed. If either are configured by env var but cannot be reached then the red error notification will be displayed. If either are not configured at all then a yellow warning notification will be displayed.

### DIFF
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -108,19 +108,43 @@ export class ApiService {
       });
     health
       .pipe(concatMap((jsonResp: any) => {
-        if (jsonResp.datasourceAvailable && jsonResp.dashboardAvailable) {
-          return forkJoin([getDatasourceURL, getDashboardURL]);
+        const toFetch: Observable<any>[] = [];
+        const unconfigured: string[] = [];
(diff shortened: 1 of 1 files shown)


ebaron (approved):
Works as expected

PR #116, opened 2020-08-10
JMX Auth UX

andrewazores:
Fixes #117

Based on top of #110

container-jfr-web-jmx-auth-2020-08-10_17 48 15

PR #389, opened 2022-03-14
feat(credentials): Manage target JMX credentials

### LINKED ISSUE #388: UI for storing and removing JMX credentials
> We might also want to build a simple API and UI for listing targets that have stored credentials - without revealing any information about what those credentials are of course - so that users can delete stored credentials graphically. For example, for old targets that they no longer need to connect to, or for targets that have been redeployed and where the user knows in advance that the required credentials have changed.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-web/issues/257#issuecomment-1014721088_

jan-law:
Fixes #388 
Adds a table listing targets with stored credentials in the Security tab. Also refactors the Security panel to make it easier to add more cards in the future.

image

### DIFF
--- a/src/app/SecurityPanel/CreateJmxCredentialModal.tsx
+++ b/src/app/SecurityPanel/CreateJmxCredentialModal.tsx
@@ -54,8 +54,8 @@ export const CreateJmxCredentialModal: React.FunctionComponent<CreateJmxCredenti
       onClose={props.onClose}
       title="Store JMX Credentials"
       description="Creates stored credentials for a given target. 
-      These are used for automated rules processing - if a Target JVM requires JMX authentication, 
-      Cryostat will use stored credentials when attempting to open JMX connections to the target."
+        If a Target JVM requires JMX authentication, Cryostat will use stored credentials 
+        when attempting to open JMX connections to the target."
     >
       <TargetSelect />
       <br />
--- a/src/app/SecurityPanel/CreateJmxCredentialModal.tsx
+++ b/src/app/SecurityPanel/CreateJmxCredentialModal.tsx
@@ -1,8 +1,8 @@
 /*
  * Copyright The Cryostat Authors
- *
+ * 
  * The Universal Permissive License (UPL), Version 1.0
- *
+ * 
  * Subject to the condition set forth below, permission is hereby granted to any
  * person obtaining a copy of this software, associated documentation and/or data
  * (collectively the "Software"), free of charge and under any and all copyright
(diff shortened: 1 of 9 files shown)


jan-law:
Should I go ahead and remove the `Delete` button in the JMX `AuthModal`? It seems redundant now that we have a table for deleting credentials. (see https://github.com/cryostatio/cryostat-web/issues/116#issuecomment-675743591_)

andrewazores:
> Should I go ahead and remove the `Delete` button in the JMX `AuthModal`? It seems redundant now that we have a table for deleting credentials. (see [#116 (comment)](https://github.com/cryostatio/cryostat-web/pull/116#issuecomment-675743591)_)

Sure, please go ahead.

src/app/SecurityPanel/CreateJmxCredentialModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const CreateJmxCredentialModal: React.FunctionComponent<CreateJmxCredentialModalProps> = (props) => {
+  return (
+    <Modal
+      isOpen={props.visible}
+      variant={ModalVariant.large}
+      showClose={true}
+      onClose={props.onClose}
+      title="Store JMX Credentials"
+      description="Creates stored credentials for a given target. 
+      These are used for automated rules processing - if a Target JVM requires JMX authentication,
```
andrewazores, line 57: These are not only used for automated rules processing anymore, right? Since this change:

https://github.com/cryostatio/cryostat/pull/858/files#diff-5135fbe0affe1bfd7a1f6338186396f7b3a4aad88ceeee4d5f8c4f6382e4ba54R126

any API request will check for and use stored credentials if they are available and the request doesn't contain overriding credentials.

src/app/SecurityPanel/CreateJmxCredentialModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+          isRequired
+          fieldId="target-select"
+          helperTextInvalid="Select a target"
+          validated={validTarget}
+        >
+          <TargetSelect />
+        </FormGroup>
+      </Form>
+      <br/>
+      <JmxAuthForm onSave={onSave} onDismiss={props.onClose} />
```
jan-law, line 93: The form focus on this modal starts at `Username` instead of `Target`, which means someone using the keyboard needs to tab through the entire modal before the focus lands on the `TargetSelect` dropdown. Adding/removing the `Form` component doesn't change the behaviour. Any ideas?

jan-law (commented):
Besides my one question about tabbing through the `CreateJmxCredentialModal`, this PR is ready for review

src/app/SecurityPanel/CreateJmxCredentialModal.tsx
andrewazores: I think that's kind of okay actually, since the `TargetSelect` will have automatically selected whichever target the user was looking at last.

Users can use shift-tab to move upward from the username field to the target selection dropdown too if they really want to change it.

andrewazores (approved):

PR #391, opened 2022-03-17
fix(notifications): handle labels nested within metadata

andrewazores:
Related to https://github.com/cryostatio/cryostat/pull/864


jan-law (commented):
I realized that I configured the frontend label validation to exclude commas and equal signs. We should probably make the validations more lenient to accommodate `template=Profiling,type=TARGET`:
https://github.com/cryostatio/cryostat-web/blob/main/src/app/CreateRecording/EditRecordingLabels.tsx#L65

jan-law (approved):

PR #397, opened 2022-03-21
feat(archives): split views for per-target and universal archives

### LINKED ISSUE #395: Archived recordings should be viewable without selecting a target

### LINKED ISSUE #396: Archived recordings should be able to filter by the selected target

andrewazores:
Fixes #395
Fixes #396

This adds a new `Archives` view accessible via the left side navigation panel. This view does not have a `TargetSelect` and displays all of the recordings in the archives, for all targets, even targets which are not currently discoverable. The existing "Archived Recordings" table view under a tab in the `Recording` view now uses a GraphQL query to populate its data instead, so it displays only archived recordings sourced from the selected target.

A table is not the best view for the new "All Targets Archives", but it's what works for now and is easily implemented since it's largely a copy-paste. Ideally however I think this should either be a Tree view (https://www.patternfly.org/v4/components/tree-view), or some kind of split-pane file explorer view like `midnight commander` or `ranger`. I'm not sure yet how to best incorporate the recording metadata/labels or other properties alongside a tree view, along with bulk operations for deleting recordings or whole folders. This is a stretch goal for this release and can be held for the next one.


src/app/Shared/Services/Api.service.tsx
```diff
@@ -474,13 +480,22 @@ export class ApiService {
...
+          return from(v.text());
+        }
+        throw from(v.text());
+      }),
       tap({
         next: () => window.onbeforeunload = null,
-        error: () => window.onbeforeunload = null,
+        error: (err) => {
+          window.onbeforeunload = null;
+          err.subscribe(msg => this.notifications.danger('Upload Failed', msg, NotificationCategory.ArchivedRecordingCreated));
```
jan-law, line 496: Is the Re-Upload Recording modal intended to stay open if you upload an invalid file? If so, I think the error toast gets hidden while the modal is open.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -401,6 +401,12 @@ export class ApiService {
     return this.sendRequest(apiVersion, path, { method: 'GET' }).pipe(map(resp => resp.json()), concatMap(from), first());
   }
 
+  graphql<T>(query: string): Observable<T> {
+    const headers = new Headers();
+    headers.set('Content-Type', 'application/graphql');
+    return this.sendRequest('beta', 'graphql', { method: 'POST', body: query, headers }).pipe(map(resp => resp.json()), concatMap(from), first());
```
jan-law, line 407: Is it just me, or does smoketest with `BasicAuthManager` return a `401` for the graphql query?

jan-law (commented):
All happy path archive-related actions are working well with `NoopAuthManager`

hareetd (commented):
Works well. Since the `ArchivedRecordingsTable` and `AllArchivedRecordingsTable` share much of the same code, do you think it's worth doing some refactoring for code re-use?

> A table is not the best view for the new "All Targets Archives", but it's what works for now and is easily implemented since it's largely a copy-paste. Ideally however I think this should either be a Tree view (https://www.patternfly.org/v4/components/tree-view), or some kind of split-pane file explorer view like midnight commander or ranger. I'm not sure yet how to best incorporate the recording metadata/labels or other properties alongside a tree view, along with bulk operations for deleting recordings or whole folders. This is a stretch goal for this release and can be held for the next one.

I can take a look into improving the view for the "All Target Archives" if you'd like. It seems like a good opportunity for me to get my hands dirty with some more involved React code, especially since it's a stretch goal and I can take my time with it?

src/app/Shared/Services/Api.service.tsx
andrewazores: Hmm. The toast is still visible for me while the modal is open...

image

andrewazores:
> Works well. Since the ArchivedRecordingsTable and AllArchivedRecordingsTable share much of the same code, do you think it's worth doing some refactoring for code re-use?

I thought about that, but figured it wasn't worthwhile if the longer term plan is to rip out the `AllArchivedRecordingsTable` implementation and replace it with a `Tree`-based or other implementation anyway.

> I can take a look into improving the view for the "All Target Archives" if you'd like. It seems like a good opportunity for me to get my hands dirty with some more involved React code, especially since it's a stretch goal and I can take my time with it?

Sure, that would be great.

src/app/Shared/Services/Api.service.tsx
hareetd: It's still visible for me as well.

jan-law: Ah - I didn't realize that the "Enable or disable notifications by category" feature preserves your settings even after I restarted the backend. I disabled a bunch of notifications while testing another feature a few days ago.

If the `ArchivedRecordingUploaded` notification is disabled, I don't see the error message.

andrewazores: Oh, yea. The notification settings go into browser local storage.

jan-law: I think this is good enough as is - I can still see the red notification badge when I close the modal.

src/app/Shared/Services/Api.service.tsx
andrewazores: Actually, it looks like this same bug also affects the `Automated Rules` view, which makes a request in pretty much the same way. When I enable Basic auth I also get a 401 response when trying to create a rule. I'll file a separate bug.

jan-law (approved):
Code looks good

PR #405, opened 2022-04-12
fix(jmx): populate credentials table via notification only (#404)

andrewazores:
Clean backport of #404


jan-law (approved):
Looks good to me

Issue #403, opened 2022-04-11
JMX credentials table renders empty view after deleting one item


jan-law:
To reproduce: store two or more credentials to populate the stored JMX credentials table. Check one item and press delete. The table appears empty. Refreshing the page will render the correct table.

PR #408, opened 2022-04-13
chore(tests): Add credentials table test

jan-law:
Related #403 

Adds unit tests for the Stored JMX Credentials Table and some refactoring to allow Jest access to the table components via imports and component labels.

~~You'll notice that the enums in StoreJmxCredentials were changed to "'TargetCredentialsStored' as NotificationCategory.TargetCredentialsStored".~~ ~~This is a workaround due to the fact that Jest doesn't support exported const enums when testing ([link to issue](https://github.com/Microsoft/TypeScript/issues/5243#issuecomment-147766222))~~
Nevermind, fixed the `messageKeys` import by adding `...jest.requireActual('@app/Shared/Services/NotificationChannel.service'),` to the mocked `NotificationChannel` service.

To make the tests easier to read, I also split up the 'Delete credentials' test case into two tests, ie `'removes the correct table entry when a deletion notification is received'` and `'makes a delete request when deleting one credential'`. The first test receives a mocked notification before asserting that the table state was updated, and the second test clicks the Add/Delete button before asserting that the API calls were made. (The equivalent API call assertion for adding stored credentials isn't tested here because the call is made in `AuthModal`.)


src/test/SecurityPanel/StoreJmxCredentials.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+jest.mock('@app/Shared/Services/Target.service', () => {
+  return {
+    TargetService: jest.fn(() => {
+      return {
+        deleteCredentials: jest.fn(),
+      };
+    }),
+  };
+});
```
jan-law, line 127: @andrewazores I think I'm missing the correct service name for `TargetService` based on the error below. I've tried replacing `TargetService: jest.fn()` with `TargetInstance jest.fn()` and `target: jest.fn()` with the same results. Any ideas? 
```
● <StoreJmxCredentials /> › removes a single table entry when the selected credentials are deleted

    TypeError: Cannot read properties of undefined (reading 'deleteCredentials')

      127 |         handleRowCheck(false, idx);
      128 |         tasks.push(context.api.deleteTargetCredentials(t));
    > 129 |         context.target.deleteCredentials(t.connectUrl);
          |                        ^
      130 |       }
      131 |     });
      132 |     addSubscription(forkJoin(tasks).subscribe());

      at src/app/SecurityPanel/StoreJmxCredentials.tsx:129:24
```

andrewazores: @hareetd ^ ?

Judging by the type error it looks like your `context.target` reference is undefined. So there is a `context` object in scope, but it's either a completely empty object or at least the `.target` property is undefined/null/uninitialized.

If I understand the mocking system here, I think what you've done seems correct, and by providing these mocks the real `Services.tsx` (https://github.com/cryostatio/cryostat-web/blob/main/src/app/Shared/Services/Services.tsx) should populate the `context` correctly via the `ServiceContext`, but that looks to me like the thing that's broken in this instance.

hareetd: Hmm, it looks Jest has issues when it comes to mocking exported `const` members of modules. This line is what seems to be issue: https://github.com/cryostatio/cryostat-web/blob/4a02275ee9f839bbb4d3820880e86728ae89dd88/src/app/Shared/Services/Target.service.tsx#L113. There's probably a workaround out there; I'll keep looking into this as well.

hareetd: Never mind, I figured out the real issue. Your current approach to mocking the `@app/Shared/Services/Target.service` module replaces the module definition with a mocked `TargetService` but any other exported members of the original module (i.e. `NO_TARGET`, `Target` and `TargetInstance`) are left undefined. In order to ensure these other exported members are preserved, you'll need a `jest.requireActual` call in your mock definition:

```
jest.mock('@app/Shared/Services/Target.service', () => {
  return {
    ...jest.requireActual('@app/Shared/Services/Target.service'),
    TargetService: jest.fn(() => {
      return {
        deleteCredentials: jest.fn(),
      };
    }),
  };
});
```
You would've needed to do this anyways in order to use the `Target` import in your mock of the `TargetsService`.

jan-law: Thanks! I think I still need to import the `Target` interface because I made a mock `Target` return object. The tests are compiling now.

hareetd (dismissed):
Looks good!

Judging from your mock on line 94, it seems the `NotificationChannel.messages` subscription is invoked three times in `StoreJmxCredentials`. For my own understanding, could you explain why this happens? I thought there would be only two invocations (one for a stored notification and the other for a deleted notification) once the render is complete and the `useEffect` calls are triggered.

src/app/SecurityPanel/StoreJmxCredentials.tsx
```diff
@@ -94,7 +94,7 @@ const Component = () => {
   }, [context, context.notificationChannel, targets, setStoredTargets]);
```
jan-law, line 94: > Judging from your mock on line 94, it seems the NotificationChannel.messages subscription is invoked three times in StoreJmxCredentials. For my own understanding, could you explain why this happens? I thought there would be only two invocations (one for a stored notification and the other for a deleted notification) once the render is complete and the useEffect calls are triggered.

@hareetd, the list of discoverable `targets` is a dependency in the `useEffect` hook that subscribes to `TargetCredentialsStored` notifications, which causes this hook to be re-run when `targets` gets populated from another `useEffect` hook during render. This hook checks that the target parsed from the notification still exists in `targets` before adding it to the table.

src/app/SecurityPanel/StoreJmxCredentials.tsx
```diff
@@ -94,7 +94,7 @@ const Component = () => {
...
 
   React.useEffect(() => {
-    const sub = context.notificationChannel.messages(NotificationCategory.TargetCredentialsDeleted).subscribe((v) => {
+    const sub = context.notificationChannel.messages('TargetCredentialsDeleted' as NotificationCategory.TargetCredentialsDeleted).subscribe((v) => {
```
andrewazores, line 97: Same q

src/app/SecurityPanel/StoreJmxCredentials.tsx
```diff
@@ -83,7 +83,7 @@ const Component = () => {
   }, [context, context.targets, setTargets, refreshStoredTargetsList]);
 
   React.useEffect(() => {
-    const sub = context.notificationChannel.messages(NotificationCategory.TargetCredentialsStored).subscribe((v) => {
+    const sub = context.notificationChannel.messages('TargetCredentialsStored' as NotificationCategory.TargetCredentialsStored).subscribe((v) => {
```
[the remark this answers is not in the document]
jan-law: Jest unfortunately doesn't support exported const enum imports when compiling tests. I added a note in the PR description along with the link to the issue. Not sure if there's a better workaround for this.

src/app/SecurityPanel/StoreJmxCredentials.tsx
jan-law: Also see my note in the PR description

src/app/SecurityPanel/StoreJmxCredentials.tsx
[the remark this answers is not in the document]
andrewazores: Ah right, thanks. That's annoying...

src/app/SecurityPanel/StoreJmxCredentials.tsx
hareetd: Makes sense, thanks!

src/app/SecurityPanel/StoreJmxCredentials.tsx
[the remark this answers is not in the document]
hareetd: I actually ran into this issue as well when setting up the unit testing framework, specifically with the `TextVariants.h4` enum used in the `Recordings` component. However, after I fixed/updated the dependencies, it stopped being an issue (you'll see the enum is used "normally" in `Recordings`). It's a bit puzzling why it's being an issue again with this file but not that one.

src/test/SecurityPanel/StoreJmxCredentials.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+        >
+          Jmx Auth Form
+        </Modal>
+      );
+    }),
+  };
+});
+
+jest.mock('@app/Shared/Services/Api.service', () => {
+  const MOCK_TARGET = { connectUrl: 'service:jmx:rmi://someUrl', alias: 'fooTarget' } as Target;
```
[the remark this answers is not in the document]
jan-law: Technically, I should be able to prefix the variables with `mock` to have them accessible inside `jest.mock()` [(jest docs)](https://jestjs.io/docs/es6-class-mocks#calling-jestmock-with-the-module-factory-parameter), but I'm still getting the same error as above. Also tried variations of casting `as Target`, `as Target[]` and using the spread operator `.mockReturnValueOnce(of([{...mockTarget} as Target])`.
```
const mockTarget = { connectUrl: 'service:jmx:rmi://someUrl', alias: 'fooTarget' } as Target;
const mockAnotherTarget = { connectUrl: 'service:jmx:rmi://anotherUrl', alias: 'anotherTarget' } as Target;
...
ApiService: jest.fn(() => {
      return {
        getTargetsWithStoredJmxCredentials: jest
          .fn()
          .mockReturnValueOnce(of([mockTarget as Target]))
...
```

Any suggestions?

src/app/SecurityPanel/StoreJmxCredentials.tsx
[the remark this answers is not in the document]
jan-law: When you updated the dependencies, did you need to do anything else besides `npm run yarn:frzinstall`?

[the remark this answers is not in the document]
hareetd: Never mind, `jest.requireActual` strikes again :)

It's not actually a const enum import issue like I had before the versioning was updated. Without the `jest.requireActual` call, the exported enum `NotificationCategory` becomes undefined when you mock out the `NotificationChannel.service. module. Insert the following line into your `NotificationChannel.service` mock, above the `NotificationChannel: jest.fn()` declaration:

```
 ...jest.requireActual('@app/Shared/Services/NotificationChannel.service'),
```
Then change the enum usage in `StoreJmxCredentials` back to normal. It should work now. 

The issue I had was different as the enum `TextVariants` from `@patternfly/react-core` was showing up as undefined even though I wasn't mocking out the module it belonged to. I believe that was an actual case of the issue you linked in the PR description, likely resolved by fixing the dependency versioning.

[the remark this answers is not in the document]
jan-law: Ah, makes sense. Thanks!

src/test/SecurityPanel/StoreJmxCredentials.test.tsx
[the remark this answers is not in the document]
hareetd: Since the issue is caused by `jest.mock()` calls being hoisted to the top of the file above any variable declarations, then [`jest.doMock`](https://jestjs.io/docs/jest-object#jestdomockmodulename-factory-options) is another option. Since `doMock` isn't hoisted to the top of the file, theoretically any out-of-scope variables should be defined. I haven't used this function yet though but let me know if you need help.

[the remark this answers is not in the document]
jan-law: After trying several [workarounds](https://github.com/facebook/jest/issues/2567) using `jest.mock`, I've been working on getting the tests to run with `jest.doMock`. The docs you linked seem to duplicate the `jest.doMock` and ` const moduleName = require('../moduleName');` for every test though. 

@hareetd Do you know if putting all of my `jest.doMock()`s inside a `beforeAll` will result in the same behaviour as the docs, but without duplicating code?

I've committed what I have so far - right now the test compilation fails with `Error: Uncaught [TypeError: Cannot read properties of undefined (reading 'subscribe')]`. Seems like Jest can't find the mocked service functions?

andrewazores:
Check on that latest commit please - the `package-lock.json` exploded.

src/test/SecurityPanel/StoreJmxCredentials.test.tsx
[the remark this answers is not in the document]
hareetd: Hey, it looks like putting `jest.doMock` or even `jest.mock` inside a `beforeAll`/`beforeEach` should [work](https://github.com/facebook/jest/issues/2582#issuecomment-284514827). I think the issue is that since the `ApiService` and `TargetsService` mocks are now being defined in the `describe` block, they show up as undefined when the `StoreJmxCredentials` import happens on line 46. 

I wonder if importing `StoreJmxCredentials` inside the `beforeAll` block would fix it; might be worth experimenting with.

src/test/SecurityPanel/StoreJmxCredentials.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+      };
+    }),
+  };
+});
+
+describe('<StoreJmxCredentials />', () => {
+  beforeAll(() => {
+    jest.resetModules();
+    const ApiService = require('@app/Shared/Services/Api.service').default;
+    const TargetsService = require('@app/Shared/Services/Api.service').default;
```
hareetd, line 117: The import path for `TargetsService` is incorrect here (the error is still the same though after fixing this).

jan-law:
I rebased my last commit and removed the extra dependencies - it turns out that you don't need dynamic importing for `jest.doMock` to work unless you are specifically using `import("../components/Component")` inside your tests.

src/test/SecurityPanel/StoreJmxCredentials.test.tsx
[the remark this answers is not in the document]
jan-law: Thanks, fixed it by moving the `doMock()`s outside of the `describe` block, then adding the `StoreJmxCredentials`, `ServiceContext`, and `defaultServices` imports after the `doMock()`s and before `describe`. I should add a note in the readme for future reference.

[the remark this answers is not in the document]
hareetd: Nice, glad it's working! Yeah, this is definitely a good thing to note in the `TESTING.md` file.

src/test/SecurityPanel/StoreJmxCredentials.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+    }),
+  };
+});
+
+jest.doMock('@app/Shared/Services/Targets.service', () => {
+  return {
+    ...jest.requireActual('@app/Shared/Services/Targets.service'),
+    TargetsService: jest.fn(() => {
+      return {
+        targets: jest.fn().mockReturnValue(of([mockTarget, mockAnotherTarget])),
```
hareetd, line 139: This can be changed to a simple `jest.fn()` implementation like `deleteTargetCredentials` above (i.e. no need for `mockReturnValue` since there's no chaining going on), just for consistency reasons.

src/test/SecurityPanel/StoreJmxCredentials.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+    ...jest.requireActual('@app/Shared/Services/Target.service'),
+    TargetService: jest.fn(() => {
+      return {
+        deleteCredentials: jest.fn(),
+      };
+    }),
+  };
+});
+
+jest.doMock('@app/Shared/Services/Api.service', () => {
```
hareetd, line 111: Just realized that this can be changed back to `jest.mock`. Even though it'll be hoisted above the `const mockTarget` and `const mockAnotherTarget` variable definitions, the actual `jest.fn` implementations inside the mocks aren't invoked until we do the imports on line 145/146, and by that time the variables would've been defined.

src/test/SecurityPanel/StoreJmxCredentials.test.tsx
hareetd: Up to you though, whichever format you prefer.

src/test/SecurityPanel/StoreJmxCredentials.test.tsx
jan-law: Oh nice, I'll change it back

hareetd (approved):

andrewazores:
@jan-law ready to merge?

PR #418, opened 2022-04-27
feat(notfound): generate NotFound from routes

### LINKED ISSUE #913: [Task] Interactive matchExpression builder
Following on from https://github.com/cryostatio/cryostat-web/issues/912

Various target object fields displayed in the details view could be clickable, which would append the field as a condition in a built matchExpression. The expression would start as the empty string. Clicking the first field would produce an expression like `target.alias == 'foo'` or `target.labels['mylabel'] == 'something'`, according to what it was the user clicked. Clicking on more fields would simply add more terms to the expression. Perhaps each field would have an `AND` and `OR` button next to it so the user can specify whether the term should be added to the expression as a conjunction or a disjunction. An expression like `(A && B) || C)` could be built by clicking `A`, `AND B`, `OR C` controls.

andrewazores:
Fixes #913

### DIFF
--- a/src/app/NotFound/NotFound.tsx
+++ b/src/app/NotFound/NotFound.tsx
@@ -38,7 +38,7 @@
 import * as React from 'react';
 import { Link } from 'react-router-dom';
 import { NotFoundCard } from './NotFoundCard';
-import { 
+import {
   Button,
   EmptyState,
   EmptyStateIcon,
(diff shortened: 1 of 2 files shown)


hareetd (approved):
Looks good, I'll keep in mind that automation is the way to go in the future :)

jan-law (approved):

hareetd:
It looks like the issue referenced (#913) in the PR description doesn't exist.

jan-law:
Good catch, should be https://github.com/cryostatio/cryostat/issues/913

andrewazores:
Oh, thanks for catching that - I filed the issue in the wrong repo.

andrewazores:
> It looks like the issue referenced (#913) in the PR description doesn't exist.
> 
> Edit: Is this PR still able to be included in the 2.1 release? If so, I can update the screenshot of the 404 page in the Release Notes.

Although it's quite a small and seemingly low-impact patch, we're in the window where only bugfixes should be backported for 2.1 release now, so this will have to wait for 2.2.

PR #427, opened 2022-04-28
fix(snapshots): remove successful snapshot creation notification handling

### LINKED ISSUE #426: Snapshot creation notification handling should be delegated to the backend
Frontend equivalent of [#916](https://github.com/cryostatio/cryostat/issues/916).

hareetd:
Fixes #426

### DIFF
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -43,10 +43,10 @@ import { ServiceContext } from '@app/Shared/Services/Services';
 import { NO_TARGET } from '@app/Shared/Services/Target.service';
 import { useSubscriptions} from '@app/utils/useSubscriptions';
 import { Button, Checkbox, Label, Text, Toolbar, ToolbarContent, ToolbarItem } from '@patternfly/react-core';
-import {  Tbody, Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
+import { Tbody, Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
 import * as React from 'react';
 import { useHistory, useRouteMatch } from 'react-router-dom';
-import { combineLatest, forkJoin, Observable, of } from 'rxjs';
+import { combineLatest, forkJoin, merge, Observable } from 'rxjs';
 import { concatMap, filter, first } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


andrewazores:
Once the backend does emit a notification when a snapshot is created then the frontend will need a [notification category and mapper ](https://github.com/cryostatio/cryostat-web/blob/5a9b7be1febcb0698a2a31144c64b159cbfe302a/src/app/Shared/Services/NotificationChannel.service.tsx#L80) added. Would you like to do that within this PR or create a second one for that separately?

hareetd:
I think doing it in this one would make sense, I'll create notification categories for "Snapshot Created" and "Snapshot Deleted".

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -129,7 +129,10 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
     addSubscription(
       combineLatest([
         context.target.target(),
-        context.notificationChannel.messages(NotificationCategory.ActiveRecordingCreated),
+        merge(
```
andrewazores, line 132: What does the `merge` here do?

hareetd: Without the `merge` the web client doesn't properly update the table state upon receiving `ActiveRecordingDeleted` or `SnapshotDeleted` notifications. I included it here as well for consistency reasons even though the `ActiveRecordingCreated` and `SnapshotCreated` notifications don't exhibit the same behavior, although that's likely because a refresh is triggered upon returning from the `/create` route and not because the table is properly updating its state based on the notifications alone.

I think the reason why `combineLatest` doesn't work on its own is that it won't emit an initial value until each of the Observables comprising it doesn't emit at least once. For example, that means the first time an `ActiveRecordingDeleted` notification is received, `combineLatest` won't emit until it also receives a `SnapshotDeleted` notification, which leads to weird state behavior in the displayed table. 

However, by merging snapshot and active recording notification Observables into one, it ensures that `combineLatest` will emit soon as either a snapshot or active recording-related notification is received.

andrewazores: Ahh, that makes sense. Thanks.

andrewazores (approved):

Issue #414, opened 2022-04-25
[Task] Search/filters in Recordings tables


andrewazores:
The Active/Archived recordings tables should be searchable/filterable by the recording's name, state, and labels. Before/after a datetime would also be nice.

andrewazores:
@jan-law I filed a second issue, #441, that I think better encapsulates the discussion we had on #428 on this topic. I think #441 and this one are very complementary (so they're Tasks grouped in a Story now), but it'd be easier to develop and review these as two separate PRs I think.

PR #428, opened 2022-05-04
feat(labels): Bulk edit recording labels

### LINKED ISSUE #415: Bulk-edit Recording labels
A user should be able to check multiple recordings in the Active or Archived tables and perform a bulk-edit action on the labels of the selected recordings. This would allow the user to add a new label to all of the selected recordings. The set intersection of all of the recordings' labels should also be displayed and editable so that the user can update common labels across all of the selected recordings.

### LINKED ISSUE #436: [Task] Bulk editing of labels for active or archived recordings within one selected target

jan-law:
Fixes #436

If you would like to edit labels for a single recording, select the recording by clicking the checkbox and click Edit Labels in the toolbar. The recording labels side panel will appear.
You can also select multiple recordings and add a label to all recordings at once. The side panel displays labels that are common to all selected recordings, as well as an Edit button to add/remove a label to/from all selected recordings. 

The ellipsis menu item 'Edit Metadata' has been removed and replaced with the side panel. There will be a follow-up PR to add the recording labels side panel in the All Archives view.

image

~~Adds a collapsible section above the recordings table for adding and deleting labels for multiple selected recordings. I also refactored the old recording labels components into `RecordingLabelFields` and `EditableLabelCell` to reuse the input fields in the new `BulkEditLabels` component. The individual label editing form and the validation logic is the same except that an empty key or value will be highlighted with a warning symbol.~~ (implementation has changed since opening the PR)

Depends on #438


src/test/RecordingMetadata.tsx/RecordingLabelFields.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+    userEvent.click(screen.getAllByTestId('remove-label-button')[0]);
+
+    expect(screen.getAllByLabelText('label key').length).toBe(1);
+    expect(screen.getAllByLabelText('label value').length).toBe(1);
+
+    expect(screen.getAllByDisplayValue('someLabel').length).toBe(0);
+    expect(screen.getAllByDisplayValue('someValue').length).toBe(0);
+    expect(screen.getAllByDisplayValue('anotherLabel')[0]).toBeInTheDocument();
+    expect(screen.getAllByDisplayValue('anotherValue')[0]).toBeInTheDocument();
+  });
```
jan-law, line 152: These two tests are failing because the `Add` and `Delete` label buttons don't change the size of the `mockLabels` array.  `await waitFor(() => expect(screen.getAllByLabelText('label key').length).toBe(1));` and `await waitFor(() => expect(screen.getAllByLabelText('label key').length).toBe(1), { timeout: 10});` also results in the same output. The buttons work fine when manually testing them.

@hareetd Do you know if there's anything I should be doing differently when passing props or setting the `mockLabels` array with Jest?

```
 ● <RecordingLabelFields /> › adds a label when Add Label is clicked

    expect(received).toBe(expected) // Object.is equality

    Expected: 3
    Received: 2

      131 |     userEvent.click(screen.getByText('Add Label'));
      132 |
    > 133 |     expect(screen.getAllByLabelText('label key').length).toBe(3);
          |                                                          ^
      134 |     expect(screen.getAllByLabelText('label value').length).toBe(3);
      135 |   });
      136 |

      at Object.<anonymous> (src/test/RecordingMetadata.tsx/RecordingLabelFields.test.tsx:133:58)
      at TestScheduler.scheduleTests (node_modules/@jest/core/build/TestScheduler.js:333:13)
      at runJest (node_modules/@jest/core/build/runJest.js:387:19)
      at _run10000 (node_modules/@jest/core/build/cli/index.js:408:7)
      at runCLI (node_modules/@jest/core/build/cli/index.js:261:3)

  ● <RecordingLabelFields /> › removes the correct label when Delete button is clicked

    expect(received).toBe(expected) // Object.is equality

    Expected: 1
    Received: 2

      143 |     userEvent.click(screen.getAllByTestId('remove-label-button')[0]);
      144 |
    > 145 |     expect(screen.getAllByLabelText('label key').length).toBe(1);
          |                                                          ^
      146 |     expect(screen.getAllByLabelText('label value').length).toBe(1);
      147 |
      148 |     expect(screen.getAllByDisplayValue('someLabel').length).toBe(0);
```

hareetd: So the thing with RTL is that it's designed purely from a user perspective. In this situation where the parent component's state and state handlers are passed in as props to the component under test, all we need to confirm is that:

1) Is the passed-in prop data rendered correctly?
2) Do changes in the component result in the appropriate prop handler being called?

From your other tests, we can confirm the passed-in labels are indeed correctly rendered so all that's left to confirm is `props.setLabels` is called when the add and delete buttons are clicked. We assume that the parent component correctly updates the labels when `props.setLabels` is called and since we know `RecordingLabelFields` will render the updated labels correctly, the test for this workflow is complete. The actual behavior of `props.setLabels` can be tested in the unit tests for the parent components themselves. 

If you think about it from a user perspective, when interacting with the `RecordingLabelFields` component they see the rendered labels and then either press "Add Label" or the "X" to delete a label, all of which you've fully tested for. The user has no idea about the internal state or props of a component, they simply interact with the render as described in the previous sentence. 

This is one of the reasons why unit tests can be tricky with RTL. If Enzyme was still an option we would be able to easily mock the internal props and state of a component but RTL doesn't allow that, which is why RTL exists on a sort of continuum between unit testing and integration testing. You could test `RecordingLabelFields` and its parent components in the same test, where the "unit" in the unit test is no longer each isolated component but instead, the entire label workflow, which would help solve these state mocking issues. However, I still think it's worthwhile to isolate components and unit test them in a traditional manner.

jan-law: > In this situation where the parent component's state and state handlers are passed in as props to the component under test

Ah thanks, that makes sense. I've moved the part of the test that checks if the state is actually updated to the parent component tests, and changed the child component test to verify that `props.setLabels` is called with the right arguments.

src/app/RecordingMetadata/BulkEditLabels.tsx
```diff
@@ -0,0 +1,163 @@
...
+    commonLabels,
+    parseLabels,
+    editing,
+    setEditing,
+    context,
+    context.api,
+  ]);
+
+  const handleEditLabels = React.useCallback(() => {
+    setEditing(!editing);
```
andrewazores, line 103: This should be `setEditing(editing => !editing)`

andrewazores:
Works really nicely. There's also a minor bug where if I select some recording(s), show the label options, and then deselect all recordings, the label editing section is still visible with empty content but the button to hide/collapse that section is disabled.

But, I'm not sure about the UI aspect of having the hidden collapsible section above the table - feels a little awkward to me. I thought of having it as a popup modal but then the modal needs to be dismissed in order to change which recordings are selected, and I like that right now it's very fluid and you can change the selection with the editor section open. What if the editor section is placed in its own card within the view instead of being an invisible collapsible section in the same card?

Is it still worth having "Edit Metadata" in the three-dots action menu when we have this UI instead? Maybe we can remove that and change the label from "(Show|Hide) Label Options" on the new toolbar button to just "Edit Labels".

Also we'd need to have some version of this present in the All Archives view as well, but maybe it's worth waiting for Hareet's rework of that view to be done first, and then we retrofit this change there to avoid dealing with the merge/rebase conflicts.

hareetd:
> Also we'd need to have some version of this present in the All Archives view as well, but maybe it's worth waiting for Hareet's rework of that view to be done first, and then we retrofit this change there to avoid dealing with the merge/rebase conflicts.

Sounds good to me

andrewazores:
Another related idea that maybe feeds into the UI here:

Having these labels in the UI is good for informational purposes, but it becomes really useful when the user can interact with the labels and use them to perform selections of recordings. So by clicking on a label in the table, belonging to any recording, the label is added to a selection set. There would somehow be an `AND|OR` operator and an `invert` option to apply to the selection set, and the result of evaluating this would be that the table rows for the matching recordings are checked. This would let the user do things like:

1. Go to Active Recordings 
2. Click the `template.name: Continuous` label
3. Click the `mylabel: whatever` label
4. Click the `AND` operator
5. Click the `Archive` button
All the recordings on the selected target which had both of those labels are archived.

1. Go to Active Recordings
2. Click the `rule: myrule` label
3. Select the `invert label selection` option
4. Click the `Delete` button
All the recordings on the selected target which were not started by `myrule` are deleted.

I think this flow is already achievable on the backend using GraphQL label selectors.

jan-law:
> Having these labels in the UI is good for informational purposes, but it becomes really useful when the user can interact with the labels and use them to perform selections of recordings. So by clicking on a label in the table, belonging to any recording, the label is added to a selection set. There would somehow be an AND|OR operator and an invert option to apply to the selection set, and the result of evaluating this would be that the table rows for the matching recordings are checked. This would let the user do things like:

I think your examples would go well with #414. I can make a follow-up PR to make the recording labels clickable, which will select recordings containing the clicked label.

jan-law:
> But, I'm not sure about the UI aspect of having the hidden collapsible section above the table - feels a little awkward to me. I thought of having it as a popup modal but then the modal needs to be dismissed in order to change which recordings are selected, and I like that right now it's very fluid and you can change the selection with the editor section open. What if the editor section is placed in its own card within the view instead of being an invisible collapsible section in the same card?

Good point. I originally hid the editor section because I wanted the 'Create' recording button to be the first item you see on the table, but if we have some sort of search bar / recording selector UI from #414 at the top, then the labels editor would fit nicely underneath the recording selector UI.

> Is it still worth having "Edit Metadata" in the three-dots action menu when we have this UI instead? Maybe we can remove that and change the label from "(Show|Hide) Label Options" on the new toolbar button to just "Edit Labels".

Yeah, I can remove it. Most people would notice the Edit Labels button before looking in the ellipsis menu anyway.

src/app/RecordingMetadata/BulkEditLabels.tsx
jan-law: I fixed 2 out of 3 `setEditing` calls. The `props.setEditing` call in `BulkEditLabels` link 94 only seems to compile with `props.setEditing(!props.editing)`. If I change it to `props.setEditing(editing => !editing)`, the error reads `Argument of type '(editing: any) => boolean' is not assignable to parameter of type 'boolean'`.

andrewazores: I think you just need to update this:

```typescript
  setEditing: (editing: boolean) => void;
```

in the `BulkEditLabelsProps` to suit the new signature.

andrewazores:
label-editor-cancel-bug-2022-05-06_09 00 33

There's a bug where editing, removing a label, cancelling, and then editing again causes the label to still be removed from the label editor. I expected that after clicking "cancel", clicking "edit" again would result in a fresh new state.

I'm still not really sure what to do about the placement of this editor component. I'm not a fan of how it feels being within the card and above the table right now, but I still don't have a better alternative. When we do get a better search/filter toolbar for the table I think it'll just end up being even more visually busy and cluttered feeling if we have that, the current toolbar action buttons, and also the label editor section all on top.

jan-law:
> I'm still not really sure what to do about the placement of this editor component. I'm not a fan of how it feels being within the card and above the table right now, but I still don't have a better alternative. When we do get a better search/filter toolbar for the table I think it'll just end up being even more visually busy and cluttered feeling if we have that, the current toolbar action buttons, and also the label editor section all on top.

Yeah, I agree. I also tried moving the labels editor to the right side and added a dummy search bar on top. The right panel also starts to look cluttered and compresses the table width. If we make the right panel collapsible or add tabs like the recordings table, I'm concerned it will be harder to locate everything. Maybe this screenshot will trigger more ideas?
image

andrewazores:
Or rather than hiding the Labels column in the table and having a "right side panel" implemented as an additional card in a split layout, it could be a [Drawer](https://www.patternfly.org/v4/components/drawer) on the right side. This is kind of like the modal idea I mentioned earlier but I think it fits better.

andrewazores:
Nice, I like that layout a lot better. It feels nice to use and it keeps the complexity of the child component hidden away when it isn't needed, as well as providing a nice visual separation between things when it is displayed.

jan-law:
The `Drawer` adds a scrollbar instead of letting the overflow ellipsis menu spill over the edge of the drawer. We could potentially use Patternfly's [CSS variables](https://www.patternfly.org/v4/developer-resources/global-css-variables/#using-the-variable-system), but I'm not sure how maintainable that would be. Any ideas?
image

andrewazores:
That's a little annoying. Using the CSS variables should be fine, though, if you can find a nice way to make that work better. `src/app/app.css` should do it. Maybe there's a way to set the z-index of that overflow menu to bring it up higher in the stack?

src/test/RecordingMetadata.tsx/BulkEditLabels.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+} as NotificationMessage;
+const mockDeleteNotification = { message: { target: mockConnectUrl, recording: mockRecording } } as NotificationMessage;
+
+jest.spyOn(defaultServices.target, 'target').mockReturnValue(of(mockTarget));
+jest.spyOn(defaultServices.api, 'doGet').mockReturnValue(of([mockRecording]));
+jest.spyOn(defaultServices.api, 'postTargetRecordingMetadata').mockReturnValue(of('updatedLabels'));
+jest.spyOn(defaultServices.api, 'postRecordingMetadata').mockReturnValue(of('updatedLabels'));
+jest.spyOn(defaultServices.notificationChannel, 'messages').mockReturnValue(of());
+jest
+  .spyOn(defaultServices.notificationChannel, 'messages')
```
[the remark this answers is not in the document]
andrewazores: I'm okay with either way. If I remember/understand correctly, `mock` is a whole replacement for the mocked function(s) that you can provide with mock behaviour and verify how/when it was called after. `spyOn` on the other hand by default is just wrapping around the real function implementation but lets you inspect how/when it was called, so you don't have to provide a mock behaviour definition if you're okay with the real impl being run.

There's a time and a place for both so if spies fit the case, go for it.

andrewazores (approved):

Issue #430, opened 2022-05-09
[Task] ArchivedRecordingsTable does not properly update view state using notifications 


hareetd:
When a notification is issued, before the table state is updated we check whether the current target in context matches the one described in the notification (see [here](https://github.com/cryostatio/cryostat-web/blob/d4e92765e1238e15766427f6845824d37ef18dff/src/app/Recordings/ArchivedRecordingsTable.tsx#L154)). However, unlike the active recordings notifications, the archived recordings notifications do not contain a `target` field for matching the `connectUrl` of the current target so the state is never updated. 

Known cases:

1) `Archived Recording Deleted` notifications do not update the table state [[1](https://github.com/cryostatio/cryostat-web/pull/431#issuecomment-1149070884), [2](https://github.com/cryostatio/cryostat-web/pull/431#issuecomment-1149071976)]
2) Unable to differentiate between uploaded vs. saved archived recordings, resulting in bad table state in #431 [[1](https://github.com/cryostatio/cryostat-web/pull/431#issuecomment-1149067098)]

andrewazores:
This only happens on the archived recordings table under the Recordings view. The Archives view for all targets behaves as expected.

maxcao13:
> When a notification is issued, before the table state is updated we check whether the current target in context matches the one described in the notification

I'm not quite familiar enough with this repo or React even but why do we check this in the ArchivedRecordingsTable and not the AllArchivedRecordingsTable?

hareetd:
The `ArchivedRecordingsTable` is target-specific so when we receive a notification from the backend, such as "Archived Recording Deleted", we need to make sure that the target to which the deleted recording belongs to is the same target that's currently selected in the view (i.e. the `ArchivedRecordingsTable` instance we're looking at in the web client belongs to the same target from the notification). 

The `AllArchivedRecordingsTable`, on the other hand, is _not_ target-specific. It simply lists the archived recordings belonging to all targets, meaning, for example, when an "Archived Recording Deletion" is received, we can simply remove the recording the `AllArchivedRecordingsTable` without having to check which target it belongs to.

maxcao13:
> Correct, the fix will require a patch to the backend to emit this information as well as a patch to the frontend to handle it accordingly.

So does this mean I have to file a new issue in [cryostat] to patch this then?

andrewazores:
Yep.

maxcao13:
[...]

PR #431, opened 2022-05-10
feat(archives): Redesign "All Archives" view

### LINKED ISSUE #398: Re-uploading a saved recording does not modify the second file name
Behaviour occurs on the `main` branch as well as #397 

To reproduce bug:
1. Create a recording
2. Archive the recording
3. Download the recording you just archived
4. Re-upload the downloaded recording
5. The second file name is the same as the first one. Any "Delete" or "Edit Metadata" actions affect both recordings.

image

This alternative code path works fine - I'm assuming this is the intended behaviour:
Click the 'plus' icon and upload the same recording file twice, the second recording gets a number appended to it.
image

### LINKED ISSUE #400: [Task] Revamp/Redesign "All Archives" view
https://github.com/cryostatio/cryostat-web/pull/397#issue-1175999375

> A table is not the best view for the new "All Targets Archives", but it's what works for now and is easily implemented since it's largely a copy-paste. Ideally however I think this should either be a Tree view (https://www.patternfly.org/v4/components/tree-view), or some kind of split-pane file explorer view like midnight commander or ranger. I'm not sure yet how to best incorporate the recording metadata/labels or other properties alongside a tree view, along with bulk operations for deleting recordings or whole folders. This is a stretch goal for this release and can be held for the next one.

hareetd:
Fixes #400
Fixes #398


hareetd:
[Adding a query](https://github.com/cryostatio/cryostat/pull/924) for all archived recordings as you suggested and using it to populate both the uploaded and target-specific archived recordings tables solves the issue of authentication-related `Exceptions` occurring. The current implementation of the `ArchivedRecordingsTable` populates the table by fetching the `targetNodes.recordings.archived` data from the backend. This causes the aforementioned `Exceptions` to occur because the `RecordingsFetcher` also creates a target connection to fetch the `targetNodes.recordings.active` data at the same time. 

I'm also thinking of adding new notifications for when targets are added/deleted in order to keep the targets list up-to-date.

andrewazores:
> The current implementation of the ArchivedRecordingsTable populates the table by fetching the targetNodes.recordings.archived data from the backend. This causes the aforementioned Exceptions to occur because the RecordingsFetcher also creates a target connection to fetch the targetNodes.recordings.active data at the same time.

Nice, that makes sense. I wonder if there is some better way to structure the implementation of that query so that it only opens the target connection when actually needed, ie querying active recordings.

> I'm also thinking of adding new notifications for when targets are added/deleted in order to keep the targets list up-to-date.

Is this something different than https://github.com/cryostatio/cryostat/blob/ed9ff7e2d13da4d6c1d51a3325098e4169845295/src/main/java/io/cryostat/platform/internal/MergingPlatformClient.java#L70 ?

hareetd:
> Nice, that makes sense. I wonder if there is some better way to structure the implementation of that query so that it only opens the target connection when actually needed, ie querying active recordings.

I can look into this in a separate PR, for now, the All Archives view can use the new query.

> Is this something different than https://github.com/cryostatio/cryostat/blob/ed9ff7e2d13da4d6c1d51a3325098e4169845295/src/main/java/io/cryostat/platform/internal/MergingPlatformClient.java#L70 ?

Oops, just a notification for target deletion then?

hareetd:
Never mind, took a look at the frontend handling for `NotificationCategory.TargetJvmDiscovery` and it includes both `FOUND` and `LOST` events. I assumed from the category name that it was only for targets being added.

hareetd:
TODO once I'm back from PTO:

1. More efficient data handling. Currently, the nested `ArchivedRecordingsTable` instances for each target in the `All Targets` table are all populated as soon as one navigates to the `Archives` page, even though none of the target rows have been expanded yet to show the underlying tables. The same behavior is observed when a target row is expanded or collapsed. This results in a bunch of unnecessary GraphQL queries which can quickly become expensive depending on the number of targets. The proper behavior should be that a nested `ArchivedRecordingsTable` is only populated when its parent target row is expanded. 
2. Search function (search for both targets and recordings?)
3. Better code re-use; there's a lot of overlap between the `ArchivedRecordingsTable` and the `UploadedArchivedRecordingsTable`.
4. Unit testing

Canter5290 (dismissed):

hareetd:
The new Archives page:

image

image

image

Thoughts?

I only have tests left to write. Let me know what you think about my design approach with the implementation itself.

andrewazores:
Looks and feels good to use. I noticed a bug when I went to "Archives > Uploads". I uploaded an existing recording, but I also had an automated rule running in the background.

image

My uploaded recording appears in the table, but so do newly archived recordings when notifications come in. Switching tabs to All Archives and back to Uploads does a full re-query and repopulation of the table, so the content goes back to being as expected - until the next archived recording notification comes in.

I think you mentioned something about finding this bug already, but I took a quick look around and didn't see if it's already filed somewhere. The root cause seems to be that the notifications emitted when archives are created/deleted doesn't include information about which archive subdirectory the file was in (which target it originated from). If there is no issue for this bug already then please file one and it can be fixed separately.

Also as an idea for a separate enhancement, it might be nice to give some indication to the user how many archives belong to each target in the Archives > All Targets table. For example adding some graphical element containing a count of recordings for that target. And perhaps if that count is 0 then the table row for that target is not expandable?

andrewazores:
I also saw this error when trying to delete all archives belonging to a specific target:

image

The error notification appears and the table does not update, though if I navigate away and back then the archived recordings do appear to be deleted as expected.

andrewazores:
Selecting each recording file and trying to individually delete it also doesn't result in the table updating. I get the success notification but the archived recording entry is still in the table and selected.

hareetd:
> I think you mentioned something about finding this bug already, but I took a quick look around and didn't see if it's already filed somewhere. The root cause seems to be that the notifications emitted when archives are created/deleted doesn't include information about which archive subdirectory the file was in (which target it originated from). If there is no issue for this bug already then please file one and it can be fixed separately.

I think the [issue](https://github.com/cryostatio/cryostat-web/issues/430) I filed should cover this case as well, even though the original bug I encountered had to do with `ArchivedRecordingDeleted` notifications not properly updating state. In both cases, it has to do with the archive subdirectory (i.e. "target") not being properly provided/processed inside the notification.

> Also as an idea for a separate enhancement, it might be nice to give some indication to the user how many archives belong to each target in the Archives > All Targets table. For example adding some graphical element containing a count of recordings for that target. And perhaps if that count is 0 then the table row for that target is not expandable?

Good idea, I'll take a look into that.

> The error notification appears and the table does not update, though if I navigate away and back then the archived recordings do appear to be deleted as expected.

> Selecting each recording file and trying to individually delete it also doesn't result in the table updating. I get the success notification but the archived recording entry is still in the table and selected.

This is another case of the issue I mentioned above. I had originally left this since you marked it as a `good first issue` for Max and Thuan, but if you like I could iron out all the issues we're having with archived recordings (both on the backend and frontend, as we discussed on IRC last week).

andrewazores:
> This is another case of the issue I mentioned above. I had originally left this since you marked it as a "good first issue" for Max and Thuan, but if you like I could iron out all the issues we're having with archived recordings (both on the backend and frontend, as we discussed on IRC last week), after this PR.

Makes sense. I'll leave it up to the three of you to decide - @maxcao13 @tthvo any interest in working with Hareet on this bug?

maxcao13:
> > This is another case of the issue I mentioned above. I had originally left this since you marked it as a "good first issue" for Max and Thuan, but if you like I could iron out all the issues we're having with archived recordings (both on the backend and frontend, as we discussed on IRC last week), after this PR.
> 
> Makes sense. I'll leave it up to the three of you to decide - @maxcao13 @tthvo any interest in working with Hareet on this bug?

Sure, I will take a look at it.

hareetd:
@maxcao13 Sounds good, let me know if you have any questions.

hareetd:
I'm a bit stumped on how best to approach keeping a count of the archived recordings per target in the `AllTargetsArchivedRecordingsTable` (parent table). Ideally, I would like to keep the querying for recordings and subsequent handling encapsulated inside the `ArchivedRecordingsTable` (child table), allowing the archived recordings data of a target to be queried/loaded only when it's been clicked. However, this means I need a way to trigger an update of the parent table's count state from inside the child table but I'm having trouble with this (see past 5 commits).

Another way would be to add a new GraphQL query that returns a `Map` consisting of `target:count` pairs to set the initial state and then use notifications to update the count as archived recordings are added/deleted. Thoughts?

andrewazores:
> Ideally, I would like to keep the querying for recordings and subsequent handling encapsulated inside the ArchivedRecordingsTable (child table), allowing the archived recordings data of a target to be queried/loaded only when it's been clicked. However, this means I need a way to trigger an update of the parent table's count state from inside the child table but I'm having trouble with this (see past 5 commits).

> Another way would be to add a new GraphQL query that returns a Map consisting of target:count pairs to set the initial state and then use notifications to update the count as archived recordings are added/deleted. Thoughts?

There are lots of good ideas in here, and I think I agree with the general high-level points overall. But, I think that for the purposes of this feature, the "least evil" is for each sub-table to perform its own query and update only when it's expanded, but also for the overall table to know how to perform essentially the same query but only use the aggregate data (counts). I think this is probably the ideal because it means that on first visit to the view, the main table only ever has to perform one query to populate all *N* of its rows. If the main table has to rely on the child tables to perform a query computer the count, and send that data back to the parent then we need to make *N* GraphQL queries when visiting this view just for the main table to fully populate.

```bash
$ http :8181/api/v2.2/graphql?query="query { targetNodes { name recordings { archived { name } } } }"
HTTP/1.1 200 OK
content-encoding: gzip
content-length: 239
vary: origin

{
    "data": {
        "targetNodes": [
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:10000/jmxrmi",
                "recordings": {
                    "archived": []
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi",
                "recordings": {
                    "archived": [
                        {
                            "name": "io-cryostat-Cryostat_self_20220610T200712Z.jfr"
                        }
                    ]
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi",
                "recordings": {
                    "archived": [
                        {
                            "name": "es-andrewazor-demo-Main_auto_myrule_20220610T200817Z.jfr"
                        },
                        {
                            "name": "es-andrewazor-demo-Main_auto_myrule_20220610T200834Z.jfr"
                        }
                    ]
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9094/jmxrmi",
                "recordings": {
                    "archived": [
                        {
                            "name": "es-andrewazor-demo-Main_auto_myrule_20220610T200817Z.jfr"
                        },
                        {
                            "name": "es-andrewazor-demo-Main_auto_myrule_20220610T200834Z.jfr"
                        }
                    ]
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9095/jmxrmi",
                "recordings": {
                    "archived": [
                        {
                            "name": "es-andrewazor-demo-Main_auto_myrule_20220610T200817Z.jfr"
                        },
                        {
                            "name": "es-andrewazor-demo-Main_auto_myrule_20220610T200833Z.jfr"
                        }
                    ]
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9096/jmxrmi",
                "recordings": {
                    "archived": []
                }
            }
        ]
    }
}
```

There's still plenty of room to improve this by using a filter on the `targetNodes` query and implementing pagination, but it's always a single query that can populate all of the high-level information needed by the parent table - it has a name for each target (should correspond 1:1 to what's in the frontend targets.service you're currently referencing for this data) as well as a list of archives belonging to each of those. That nested `archived` array can just be mapped to its length and displayed into the table `Count` column on the client side.

My problem with this solution I'm presenting is that it means the client will be receiving a lot of data about archived recordings that it's just going to end up discarding - all we want is the number of recordings in each target archive, but we're getting a whole JSON object with at least the recording's name string in it. This could add up to a lot of wasted bandwidth if each target has a lot of archived recordings. To solve that, here's some reading with ideas:

https://stackoverflow.com/questions/55940123/possible-to-count-result-list-elements-in-type-graphql
https://stackoverflow.com/questions/34321688/can-graphql-return-aggregate-counts

This would suggest that, for example, we could modify our [GraphQL Schema](https://github.com/cryostatio/cryostat/blob/2b052d65717f71c2f00c43b3539c5edbc2133ed7/src/main/resources/types.graphqls#L85) for the response type of that `recordings` structure. Instead of just containing a plain list/array of recording objects, we could restructure that and return a composed object containing the list as well as a "summary" or "aggregates" substructure, which would contain properties like the size of the list.

The query above could then hypothetically look like this instead:

```bash
$ http :8181/api/v2.2/graphql?query="query { targetNodes { name recordings { archived { aggregate { count } } } } }"
HTTP/1.1 200 OK
content-encoding: gzip
content-length: 239
vary: origin

{
    "data": {
        "targetNodes": [
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:10000/jmxrmi",
                "recordings": {
                    "archived": {
                        "aggregate": {
                            count: 0
                        }
                    }
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi",
                "recordings": {
                    "archived": {
                        "aggregate": {
                            count: 1
                        }
                    }
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi",
                "recordings": {
                    "archived": {
                        "aggregate": {
                            count: 2
                        }
                    }
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9094/jmxrmi",
                "recordings": {
                    "archived": {
                        "aggregate": {
                            count: 2
                        }
                    }
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9095/jmxrmi",
                "recordings": {
                    "archived": {
                        "aggregate": {
                            count: 2
                        }
                    }
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9096/jmxrmi",
                "recordings": {
                    "archived": {
                        "aggregate": {
                            count: 0
                        }
                    }
                }
            }
        ]
    }
}
```

I think updating the parent table in this way is fairly close to an ideal solution. This also means the child tables can continue using the same query and backend implementation as the parent table really - they just use a `filter` on the top-level query to narrow the view down to their specific target, and they use the actual archived recordings list rather than the `aggregate.count` property.

hareetd:
Awesome, appreciate the links/info. That aggregate count approach should work perfectly. Thanks!

hareetd:
The counts are working now. I still have some implementation work left to do but I'd appreciate any feedback on how I've designed the `AllTargetsArchivedRecordingsTable`, particularly with how the `Map<Target, number>` state representing the targets and their archived recordings counts is maintained and updated.

hareetd:
It looks like I've fixed the efficiency issue. The deep comparison alone wasn't enough, I needed to separate the processing of the parent and child rows from each other so in the case where the `searchedCounts` are updated but not the `searchedTargets`, only the parent row and not the child row are updated. 

However, I noticed a new issue (it occurs on old commits as well, not just after these new changes) where if you perform a search while a parent row is expanded, and then delete the search characters, it messes up the row ordering so that parent rows and child rows are mismatched. It also results in the data of any expanded tables being re-queried even though the table is already open. 

Since I essentially used and modified the search implementation in `EventTemplates.tsx`, I'll need to investigate further and ensure everything works seamlessly with regard to expanded rows.

hareetd:
Alright, so from an efficiency standpoint, I'm happy with the implementation now. The data inside expanded recordings tables is no longer re-queried for if you perform a search, or when any recordings-related notifications are received. I've stress-tested using automated rules but I'd appreciate it if you could perform QA as well in case I missed anything.

However, it still isn't perfect. When a nested table contains a lot of recordings (30 in the example video below), scrolling causes the view to lag, with parts disappearing and reappearing seemingly at random. Compare this to the target with only 10 recordings which has no issues when scrolling. 

Interestingly, if you leave the Archives, go to the Recordings view for the target with 30 recordings, and click the Archived Recordings tab, there are no issues when scrolling, even though it's the same `ArchivedRecordingsTable` shown in both cases (albeit different instances). Also, since the Rule responsible for the 30 recordings was already deleted before the start of the video, there shouldn't be any Reactive interactions going on (i.e. the table is effectively static) so I'm not sure why this is happening.

It could be a Patternfly issue, especially since the nested tables also have expandable rows themselves but more likely it's an issue with my implementation. Do you think it's worth tinkering around more to try to resolve this or since we plan on implementing pagination at some point, wait for it to be resolved then?

https://user-images.githubusercontent.com/41028544/179082181-c86ab4a7-3222-4a9d-b2b9-51ecab8b8f4e.mov

andrewazores:
Just reading your comments and looking at the video, I suspect it's a React rendering issue - there's probably some component that is being re-rendered too often, or is too costly to re-render. It could very well have to do with Patternfly or how we are using a Patternfly component, too.

https://reactjs.org/blog/2019/08/15/new-react-devtools.html

The React devtools profiler is something to investigate here. I think it should help trace if that is really a re-rendering issue, and if so, how often the renders are occurring and (hopefully) why they are occurring.

I'll review the code later, but from what I've seen before and see in this video the functionality looks great. If it isn't spamming the backend with unnecessary GraphQL queries then the frontend performance is something we can revisit in a follow-up PR, just so this isn't lingering forever (and so #478 isn't blocked forever).

hareetd:
I used the React devtools profiler and it doesn't look like there's any rendering going on during the scrolling. I tried a few different things but no dice. The latest idea is to try implementing scrolling inside the nested table using [sticky headers](https://www.patternfly.org/v4/components/table/#composable-sticky-columns-and-header), which requires updating the Patternfly packages to the latest backward-compatible ones. Regardless, I'm not going to spend much more time on this, the main priority at this point is to get unit tests added, and then the PR should be ready to go.

andrewazores:
Just a thought: if the row has a count of 0 in the main table, should we maybe just not create a row for that? There's no action available to add a new content directly from here (and I don't think there should be), and the empty row can't even be expanded or otherwise interacted with, so it's just kind of visual clutter.

image

andrewazores:
Or maybe there can be a main table toolbar with a checkbox for hiding the empty rows?

andrewazores:
Latest commits look and feel great to use. Awesome work. Could you rebase this and then I'll work through reviewing the implementation?

hareetd:
Before I rebase and add the unit tests, note that there is a bug still. In order to solve the issue where scrolling leads to temporary whitespace in the table, I made the header sticky and wrapped the table in an `OuterScrollContainer` and `InnerScrollContainer` so that the scrolling in the recordings table is separate from that of the main page. 

Unfortunately, this leads to the recording actions dropdown being cut off by the boundaries of the containers which means users have to scroll down to see the full dropdown:

https://user-images.githubusercontent.com/41028544/179568164-15a5da81-c2ae-47b7-a8dc-ba1907e205d4.mov

It's a fairly minor bug so I'll file an issue after this PR has been merged.

andrewazores:
Not sure if it's just me, but:

image

hareetd:
Hmm, weird the video works for me. How about if I attach it to this comment?

https://user-images.githubusercontent.com/41028544/179603070-c9a5ac92-0dec-4f95-b402-64e21381b8e0.mov

andrewazores:
Ah yep, I've got my video issue fixed. It was just missing codecs.

andrewazores:
Not sure what happened to the CI build action. Looks like it's just hanging on reporting even though I can see completed runs around the right timeframe: https://github.com/cryostatio/cryostat-web/actions?query=actor%3Ahareetd

This branch needs some merge conflicts fixed anyway, so hopefully after those are fixed and the branch is updated, CI will figure out what it's doing.

andrewazores (approved):
Wonderful work, thanks.

PR #443, opened 2022-05-26
fix(recordings): disable actions after deleting all recordings

### LINKED ISSUE #437: 'Archive' and 'Delete' buttons still enabled after deleting all recordings
Create one recording and select it. The Archive and Delete buttons should become enabled. Then open a second web-client in a new tab and delete that recording. Now both tabs show an empty flight recording table, but the first web client's Archive and Delete buttons are still enabled. Clicking those buttons doesn't do anything.

image

jan-law:
Fixes #437 

The problem with the previous fix was that `handleRowCheck` correctly removes the index of the deleted recording from the `checkedIndices` list, but  the indices of the remaining recordings after the deleted row still needed to be decremented by 1. The `checkedIndices` were also modified twice for the web client that manually deleted the recording and received the deletion notification.

This fix will use the index of the deleted recording to update `checkedIndices` only when a deletion notification is received. If an unrelated web client selected recordings, those `checkedIndices` should be preserved to avoid re-selecting recordings every time a recording gets deleted.

This PR is affected by https://github.com/cryostatio/cryostat-web/issues/430 - you can test the archived recordings table by commenting out lines 161-163 in `ArchivedRecordingsTable` in the meantime.

### DIFF
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -167,18 +167,24 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
         if (currentTarget.connectUrl != event.message.target) {
           return;
         }
-        
-        setRecordings(old => old.filter((r, idx) => {
-          if (r.name == event.message.recording.name) {
-            handleRowCheck(false, idx);
-            return false;
-          } else {
+        let deleted;
+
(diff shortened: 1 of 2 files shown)


jan-law:
If you close and re-open an issue, the project board doesn't move the issue back to the In Progress column. I manually moved it to see what happens after this gets merged

andrewazores (approved):

PR #453, opened 2022-06-09
chore(webpack): upgrade version and plugins

### LINKED ISSUE #452: [Task] Upgrade webpack and plugins
Possibly related #451

andrewazores:
Fixes #452

The line deletions in `webpack.dev` are because those configuration options are no longer supported by this version of the devserver. They don't seem to be needed for anything anymore - manually testing the devserver setup seems to show everything working as normal with these config lines simply deleted.


hareetd (approved):

ebaron:
@andrewazores what causes the change in yarn.lock from yarnpkg.com to npmjs.org?

andrewazores:
@ebaron hmm, good question. I'm not sure actually. I'll see what happens if I replace all those URLs back to how they were.

PR #460, opened 2022-06-14
fix(navigation): fix unncessary page refresh when following breadcrumbs

### LINKED ISSUE #459: Longer reload time when following breadcrumbs
When creating rules, if a user clicks on `Automated Rules` link to get back to rule listings, the screen takes much longer time to reload as opposed to clicking on the the side tab. This also happens when creating recordings. 

screenshot

tthvo:
Fixes #459 

Fixes view refresh taking a long time when following breadcrumbs.

### DIFF
--- a/src/app/BreadcrumbPage/BreadcrumbPage.tsx
+++ b/src/app/BreadcrumbPage/BreadcrumbPage.tsx
@@ -36,6 +36,7 @@
  * SOFTWARE.
  */
 import * as React from 'react';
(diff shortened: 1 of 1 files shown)


tthvo:
I chose the `<Link>` approach since I think it would be a more centralized fix. I read the history router approach but it seems we need to check and pass correct callbacks to `onclick`.

andrewazores (approved):

PR #463, opened 2022-06-16
feat(ui): add deletion prompt when deleting an active or archived recording

### LINKED ISSUE #382: [Story] Deletions should prompt for confirmation
When the user deletes an active or archived recording or an automated rule, they should be presented with a prompt confirming this destructive action. The prompt should contain some contextual information about what is being removed and the consequences for doing so (eg explain what it means to delete an active recording, or to delete an archived recording). For the case of automated rules there should also be a checkbox or toggle to set the value of the `?clean={boolean}` query parameter, with an explanation of what "cleaning" means when deleting a rule.

Finally, there should also be a checkbox on the prompt that says something like "do not ask me for confirmation again", which stores a user preference in local storage to skip displaying the prompt in the future and to implicitly pass. This preference should be exposed in the user settings menu as well so that the user can enable/disable the prompts at will.

Tasks:
- [ ] #462
- [ ] [Task] Add deletion prompt for custom Event Templates
- [ ] [Task] Add deletion prompt for Automated Rules
- [ ] [Task] Add deletion prompt for JMX Credentials
- [ ] add task for "do not ask again" feature

### LINKED ISSUE #462: [Task] Add deletion prompt when deleting an active or archived recording
Parent: https://github.com/cryostatio/cryostat-web/issues/382

maxcao13:
Fixes #462 
Also fixes #382

I didn't add a task or mention to add a prompt for deleting rules yet because I wanted to take it steady with React. If anyone thinks it should be included in this PR and the associated issue, please feel free. 

Any feedback and thoughts on what I've done?

### DIFF
--- a/src/app/Archives/AllArchivedRecordingsTable.tsx
+++ b/src/app/Archives/AllArchivedRecordingsTable.tsx
@@ -247,15 +247,15 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
     else {
       handleDeleteRecordings();
     }
-  }, [handleDeleteRecordings])
+  }, [context, context.settings, setWarningModalOpen, handleDeleteRecordings])
 
   const handleWarningModalClose = React.useCallback(() => {
     setWarningModalOpen(false);
   }, [setWarningModalOpen]);
 
-  const disableDeletionDialog = () => {
+  const disableDeletionDialog = React.useCallback(() => {
     context.settings.setDeletionDialogsEnabledFor(DeleteWarningType.DeleteArchivedRecordings, false);
-  };
+  }, [context, context.settings]);
 
   const RecordingsToolbar = () => {
     const deleteArchivedWarningModal = React.useMemo(() => {
--- a/src/app/Archives/AllArchivedRecordingsTable.tsx
+++ b/src/app/Archives/AllArchivedRecordingsTable.tsx
@@ -40,14 +40,14 @@ import { ArchivedRecording } from '@app/Shared/Services/Api.service';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { NotificationCategory } from '@app/Shared/Services/NotificationChannel.service';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
-import { Button, Checkbox, Toolbar, ToolbarContent, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
+import { Button, Checkbox, Label, Text, Toolbar, ToolbarContent, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
 import { Tbody, Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
 import { RecordingActions } from '@app/Recordings/RecordingActions';
 import { RecordingsTable } from '@app/Recordings/RecordingsTable';
 import { ReportFrame } from '@app/Recordings/ReportFrame';
 import { Observable, forkJoin, merge } from 'rxjs';
 import { first } from 'rxjs/operators';
-import { UploadIcon } from '@patternfly/react-icons';
+import { PlusIcon, UploadIcon } from '@patternfly/react-icons';
 import { ArchiveUploadModal } from './ArchiveUploadModal';
 import { parseLabels } from '@app/RecordingMetadata/RecordingLabel';
 import { LabelCell } from '@app/RecordingMetadata/LabelCell';
@@ -96,7 +96,7 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
     setIsLoading(true);
     addSubscription(
       context.api.doGet<ArchivedRecording[]>('recordings', 'v1')
-        .subscribe(handleRecordings)
+      .subscribe(handleRecordings)
     );
   }, [addSubscription, context, context.api, setIsLoading, handleRecordings]);
 
@@ -129,6 +129,7 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
           o => o.name == v.message.recordingName ? { ...o, metadata: { labels: v.message.metadata.labels } } : o)))
     );
   }, [addSubscription, context, context.notificationChannel, setRecordings]);
+
 
   const handleDeleteRecordings = () => {
     const tasks: Observable<any>[] = [];
@@ -164,7 +165,7 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
       return parseLabels(props.recording.metadata.labels);
     }, [props.recording.metadata.labels]);
 
-    const expandedRowId = `archived-table-row-${props.index}-exp`;
+    const expandedRowId =`archived-table-row-${props.index}-exp`;
     const handleToggle = () => {
       toggleExpanded(expandedRowId);
     };
@@ -178,7 +179,7 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
     };
 
     const parentRow = React.useMemo(() => {
-      return (
+      return(
         <Tr key={`${props.index}_parent`}>
           <Td key={`archived-table-row-${props.index}_0`}>
             <Checkbox
@@ -189,21 +190,21 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
             />
           </Td>
           <Td
-            key={`archived-table-row-${props.index}_1`}
-            id={`archived-ex-toggle-${props.index}`}
-            aria-controls={`archived-ex-expand-${props.index}`}
-            expand={{
-              rowIndex: props.index,
-              isExpanded: isExpanded,
-              onToggle: handleToggle,
-            }}
-          />
+              key={`archived-table-row-${props.index}_1`}
+              id={`archived-ex-toggle-${props.index}`}
+              aria-controls={`archived-ex-expand-${props.index}`}
+              expand={{
+                rowIndex: props.index,
+                isExpanded: isExpanded,
+                onToggle: handleToggle,
+              }}
+            />
           <Td key={`archived-table-row-${props.index}_2`} dataLabel={tableColumns[0]}>
             {props.recording.name}
           </Td>
           <Td key={`active-table-row-${props.index}_3`} dataLabel={tableColumns[1]}>
-            <LabelCell
-              labels={parsedLabels}
+            <LabelCell 
+              labels={parsedLabels} 
             />
           </Td>
           <RecordingActions
@@ -223,9 +224,9 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
             dataLabel={"Content Details"}
             colSpan={tableColumns.length + 3}
           >
-            <ExpandableRowContent>
-              <ReportFrame isExpanded={isExpanded} recording={props.recording} width="100%" height="640" />
-            </ExpandableRowContent>
+          <ExpandableRowContent>
+            <ReportFrame isExpanded={isExpanded} recording={props.recording} width="100%" height="640" />
+          </ExpandableRowContent>
           </Td>
         </Tr>
       )
@@ -254,8 +255,11 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
 
   const RecordingsToolbar = () => {
     const deleteArchivedWarningModal = React.useMemo(() => {
-      return <DeleteWarningModal
+      const filtered = recordings.filter((r: ArchivedRecording, idx: number) => checkedIndices.includes(idx));
+      const items = filtered.map((r) => `${r.name}`);
+      return <DeleteWarningModal 
         warningType={DeleteWarningType.DeleteArchivedRecordings}
+        items={items}
         visible={warningModalOpen}
         onAccept={handleDeleteRecordings}
         onClose={handleWarningModalClose}
@@ -271,19 +275,19 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
           </ToolbarGroup>
           <ToolbarGroup variant="icon-button-group">
             <ToolbarItem>
-              <Button key="upload" variant="secondary" onClick={() => setShowUploadModal(true)}>
-                <UploadIcon />
-              </Button>
+                <Button key="upload" variant="secondary" onClick={() => setShowUploadModal(true)}>
+                  <UploadIcon/>
+                </Button>
             </ToolbarItem>
           </ToolbarGroup>
-          {deleteArchivedWarningModal}
+          { deleteArchivedWarningModal }
         </ToolbarContent>
       </Toolbar>
     );
   };
 
   const recordingRows = React.useMemo(() => {
-    return recordings.map((r, idx) => <RecordingRow key={idx} recording={r} index={idx} />)
+    return recordings.map((r, idx) => <RecordingRow key={idx} recording={r} index={idx}/>)
   }, [recordings, expandedRows, checkedIndices]);
 
   const handleModalClose = React.useCallback(() => {
@@ -293,19 +297,19 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
 
   return (<>
     <RecordingsTable
-      tableTitle="Archived Flight Recordings"
-      toolbar={<RecordingsToolbar />}
-      tableColumns={tableColumns}
-      isHeaderChecked={headerChecked}
-      onHeaderCheck={handleHeaderCheck}
-      isLoading={isLoading}
-      isEmpty={!recordings.length}
-      errorMessage=''
+        tableTitle="Archived Flight Recordings"
+        toolbar={<RecordingsToolbar />}
+        tableColumns={tableColumns}
+        isHeaderChecked={headerChecked}
+        onHeaderCheck={handleHeaderCheck}
+        isLoading={isLoading}
+        isEmpty={!recordings.length}
+        errorMessage=''
     >
       {recordingRows}
     </RecordingsTable>
 
-    <ArchiveUploadModal visible={showUploadModal} onClose={handleModalClose} />
+    <ArchiveUploadModal visible={showUploadModal} onClose={handleModalClose}/>
   </>);
 };
 
--- a/src/app/Archives/AllArchivedRecordingsTable.tsx
+++ b/src/app/Archives/AllArchivedRecordingsTable.tsx
@@ -252,10 +252,6 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
     setWarningModalOpen(false);
   }, [setWarningModalOpen]);
 
-  const disableDeletionDialog = React.useCallback(() => {
-    context.settings.setDeletionDialogsEnabledFor(DeleteWarningType.DeleteArchivedRecordings, false);
-  }, [context, context.settings]);
-
   const RecordingsToolbar = () => {
     const deleteArchivedWarningModal = React.useMemo(() => {
       return <DeleteWarningModal
@@ -263,7 +259,6 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
         visible={warningModalOpen}
         onAccept={handleDeleteRecordings}
         onClose={handleWarningModalClose}
-        disableDialog={disableDeletionDialog}
       />
     }, [recordings, checkedIndices]);
     return (
--- a/src/app/Archives/AllArchivedRecordingsTable.tsx
+++ b/src/app/Archives/AllArchivedRecordingsTable.tsx
@@ -40,7 +40,7 @@ import { ArchivedRecording } from '@app/Shared/Services/Api.service';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { NotificationCategory } from '@app/Shared/Services/NotificationChannel.service';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
-import { Button, Checkbox, Label, Text, Toolbar, ToolbarContent, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
+import { Button, Checkbox, Toolbar, ToolbarContent, ToolbarGroup, ToolbarItem } from '@patternfly/react-core';
 import { Tbody, Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
 import { RecordingActions } from '@app/Recordings/RecordingActions';
 import { RecordingsTable } from '@app/Recordings/RecordingsTable';
@@ -259,7 +259,6 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
       const items = filtered.map((r) => `${r.name}`);
       return <DeleteWarningModal 
         warningType={DeleteWarningType.DeleteArchivedRecordings}
-        items={items}
         visible={warningModalOpen}
         onAccept={handleDeleteRecordings}
         onClose={handleWarningModalClose}
--- a/src/app/Archives/AllArchivedRecordingsTable.tsx
+++ b/src/app/Archives/AllArchivedRecordingsTable.tsx
@@ -240,25 +240,34 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
     );
   };
 
-  const handleDeleteButton = () => {
+  const handleDeleteButton = React.useCallback(() => {
     if (context.settings.deletionDialogsEnabledFor(DeleteWarningType.DeleteArchivedRecordings)) {
       setWarningModalOpen(true);
     }
     else {
       handleDeleteRecordings();
     }
-  }
+  }, [handleDeleteRecordings])
+
+  const handleWarningModalClose = React.useCallback(() => {
+    setWarningModalOpen(false);
+  }, [setWarningModalOpen]);
+
+  const disableDeletionDialog = () => {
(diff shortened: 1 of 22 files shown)


andrewazores (commented):
This is looking really, really good so far. Nice work. Looking forward to getting this merged :-)

maxcao13:
Btw, I notice that passing clean=true does not actually delete any active recordings created by the Automated Rule as mentioned here https://github.com/cryostatio/cryostatio.github.io/issues/36#issuecomment-1059145447. Was this intended to be implemented later after this task was resolved?

andrewazores:
`?clean=true` is implemented: https://github.com/cryostatio/cryostat/blob/f83117582d5404a3d1590a09cbabbd67c0bea120/src/main/java/io/cryostat/net/web/http/api/v2/RuleDeleteHandler.java#L149

It doesn't delete active recordings, it just stops them. That's intentional, so that the runtime throughput overhead of the recording is removed, but the latest JFR data is still available for the user to grab if they need it.

maxcao13:
Unrelated, but how much does the "react-axe" dependency matter for our development? It seems to popup a lot for me in the web developer console in my browser detailing accessibility issues, but I don't see much fuss about it, just wondering.

andrewazores:
Accessibility is important and something we should paid more attention to. Internationalization/localization, too.

src/app/Modal/DeleteWarningUtils.tsx
```diff
@@ -47,41 +47,47 @@ export interface DeleteWarning {
     id: DeleteWarningType;
     title: string;
     description: string;
+    descriptionPartTwo: string;
```
andrewazores, line 50: Are these used anywhere yet?

src/app/Modal/DeleteWarningModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+        ]}
+      >
+        <Stack hasGutter key="modal-checkboxes-stack">
+          {(realWarningType === DeleteAutomatedRules) && (typeof setCheckbox !== 'undefined') &&
+          <Checkbox id="clean-rule-enabled" 
+            label="Clean" 
+            description={`Clean will stop any Active Recordings that ${items![0]} created.`}
+            isChecked={checkbox} 
+            onChange={(checked) => setCheckbox(checked)}
+          />}
```
andrewazores, line 98: Hmm. I'm not a fan of having subtype-specific logic hardcoded into the parent type like this. Maybe this can be accomplished by refactoring this component into two components? One generic one that can be used everywhere else, and one specialized subtype for Automated Rules that also adds this checkbox control?

src/app/Modal/DeleteWarningModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const [doNotAsk, setDoNotAsk] = useState(false);
+  const realWarningType = getFromWarningMap(warningType);
+
+  const description = `${realWarningType?.description}${(typeof items === 'undefined' || items.length <= 1) ?  "":"s"}: [${items?.join(", ")}]. ${realWarningType?.descriptionPartTwo}`
+
+  const onAcceptClose = () => {
+    onAccept();
+    onClose();
+    if (doNotAsk) {
+      setShowDialog();
```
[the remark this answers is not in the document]
maxcao13: My thinking was that the dialog checkbox would always set the showDialog to be false since it wouldn't make sense if the dialog option was false, and setting the checkmark would make it true because we wouldn't be able to see the dialog in the first place, if that makes sense.

src/app/Modal/DeleteWarningModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+          </Stack> 
+        ]}
+      >
+        <Stack hasGutter key="modal-checkboxes-stack">
+          {(realWarningType === DeleteAutomatedRules) && (typeof setCheckbox !== 'undefined') &&
+          <Checkbox id="clean-rule-enabled" 
+            label="Clean" 
+            description={`Clean will stop any Active Recordings that ${items![0]} created.`}
+            isChecked={checkbox} 
+            onChange={(checked) => setCheckbox(checked)}
```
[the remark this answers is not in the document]
maxcao13: Checking my understanding here:

I wrote this: 
```
  const disableRuleDialog = React.useCallback(() => {
    context.settings.setDeletionDialogsEnabledFor(DeleteWarningType.DeleteAutomatedRules, false);
  }, [context.settings.setDeletionDialogsEnabledFor])
...
...
            <RuleDeleteWarningModal 
            ...                 
            ...
                disableDialog={disableRuleDialog}
            />
```

1. Would there be a point in using useCallback here? The disableDialog function is only invoked when the Delete button is clicked with the checkmark checked, and in that case, we would probably want a re-render anyways, right? 

2. If not then, when doing this
```
            <RuleDeleteWarningModal 
                 ...
                onAccept={() => { handleDelete(rowDeleteData, cleanRuleEnabled); } }
                ....
            />
```
vs 
```
const foo = () => {
     handleDelete(...);
}
...
...
            <RuleDeleteWarningModal 
                 ...
                onAccept={foo}
                ....
            />

```
is there any difference between passing in an anonymous prop function in the first example vs the second?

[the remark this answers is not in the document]
andrewazores: In this particular example, where the "real" function being called ("setDeletionDialogsEnabledFor") doesn't depend on any other piece of state from the component, I don't think it ends up mattering in the end so far as React and rendering goes.

If it did depend on some other piece of state, then I think `useCallback` would be the way to go. This way the function body is properly updated to use fresh state data that it needs. There isn't much reason *not* to use `useCallback` here, other than it being unnecessary and adding a bit of extra boilerplate code. At least this way, if the function definition is changed in the future and begins to depend on some piece of state, it's already nicely set up to do that by just adjusting the dependency array.

If you don't want to use `useCallback`, `const foo = () =>` is better just because it moves code out of the JSX template and into the normal-feeling TypeScript part of the component definition. If you wanted to add some other behaviours to that `onAccept` event handler, it gets really messy doing it the `onAccept={() => { doSomething(); const value = doAComputation(); andSomethingElse(value); } }` way.

src/app/Archives/AllArchivedRecordingsTable.tsx
```diff
@@ -240,25 +240,34 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
...
-  const handleDeleteButton = () => {
+  const handleDeleteButton = React.useCallback(() => {
     if (context.settings.deletionDialogsEnabledFor(DeleteWarningType.DeleteArchivedRecordings)) {
       setWarningModalOpen(true);
     }
     else {
       handleDeleteRecordings();
     }
-  }
+  }, [handleDeleteRecordings])
```
andrewazores, line 250: `context`, `context.settings`, and `setWarningModalOpen` should all also be dependencies here.

src/app/Archives/AllArchivedRecordingsTable.tsx
```diff
@@ -240,25 +240,34 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
...
+
+  const handleWarningModalClose = React.useCallback(() => {
+    setWarningModalOpen(false);
+  }, [setWarningModalOpen]);
+
+  const disableDeletionDialog = () => {
+    context.settings.setDeletionDialogsEnabledFor(DeleteWarningType.DeleteArchivedRecordings, false);
```
andrewazores, line 257: Any particular reason to not make this one a `useCallback()` too?

src/app/Events/EventTemplates.tsx
```diff
@@ -256,14 +255,27 @@ export const EventTemplates = () => {
     setSortBy({ index, direction });
   };
 
-  const handleDeleteButton = () => {
+  const handleDeleteButton = (rowData) => {
```
andrewazores, line 258: This one can be a `useCallback()` too.

src/app/Events/EventTemplates.tsx
```diff
@@ -256,14 +255,27 @@ export const EventTemplates = () => {
...
     else {
-      handleDelete(rowDeleteData)
+      handleDelete(rowData);
     }
-  }
+  };
+
+  const handleWarningModalAccept = React.useCallback(() => {
+    handleDelete(rowDeleteData);
+  }, [rowDeleteData]);
```
andrewazores, line 270: `handleDelete` should be in the deps array

src/app/Events/EventTemplates.tsx
```diff
@@ -256,14 +255,27 @@ export const EventTemplates = () => {
...
+
+  const handleWarningModalClose = React.useCallback(() => {
+    setWarningModalOpen(false);
+  }, [setWarningModalOpen]);
+
+  const disableDeletionDialog = () => {
```
andrewazores, line 276: `useCallback()` here too?

src/app/Modal/DeleteWarningModal.tsx
```diff
@@ -46,62 +46,58 @@ export interface DeleteWarningProps {
...
-export const DeleteWarningModal = ({ warningType , items, visible, onAccept, onClose, setShowDialog, checkbox, setCheckbox}: DeleteWarningProps): JSX.Element => {
+export const DeleteWarningModal = ({ warningType , items, visible, onAccept, onClose, disableDialog}: DeleteWarningProps): JSX.Element => {
   const [doNotAsk, setDoNotAsk] = useState(false);
   const realWarningType = getFromWarningMap(warningType);
 
-  const description = `${realWarningType?.description}${(typeof items === 'undefined' || items.length <= 1) ?  "":"s"}: [${items?.join(", ")}]. ${realWarningType?.descriptionPartTwo}`
+  const description = 
+    <TextContent>
+      <Text>
+        {realWarningType?.description}{(typeof items === 'undefined' || items.length <= 1) ?  "":"s"}: [{items?.join(", ")}]
```
andrewazores, line 59: The `typeof` check here is a little unconventional. Does `items?.length <= 1` fit the need here?

src/app/Modal/DeleteWarningModal.tsx
```diff
@@ -46,62 +46,58 @@ export interface DeleteWarningProps {
...
-          />}
-          <Checkbox id="do-not-ask-enabled" 
-            label="Don't ask me again" 
-            isChecked={doNotAsk} 
-            onChange={(checked) => setDoNotAsk(checked)}
-          />
-        </Stack>
-      </Modal>  
+    <Modal
+      title={`Permanently ${realWarningType?.title}?`}
```
andrewazores, line 75: I think it'd be better to move the `Permanently ` prefix into the actual "warning type title". This keeps the component more generic, and easier to localize in the future.

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -427,14 +427,22 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
...
-  const handleDeleteButton = () => {
+  const handleDeleteButton = React.useCallback(() => {
     if (context.settings.deletionDialogsEnabledFor(DeleteWarningType.DeleteActiveRecordings)) {
       setWarningModalOpen(true);
     }
     else {
       handleDeleteRecordings();
     }
-  }
+  }, [handleDeleteRecordings]);
```
andrewazores, line 437: Missing some deps on this one, too

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -427,14 +427,22 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
...
+
+  const handleWarningModalClose = React.useCallback(() => {
+    setWarningModalOpen(false);
+  }, [setWarningModalOpen]);
+
+  const disableDeletionDialog = () => {
```
andrewazores, line 443: `useCallback()`?

src/app/Recordings/ArchivedRecordingsTable.tsx
```diff
@@ -303,25 +303,34 @@ export const ArchivedRecordingsTable: React.FunctionComponent<ArchivedRecordings
...
-  const handleDeleteButton = () => {
+  const handleDeleteButton = React.useCallback(() => {
     if (context.settings.deletionDialogsEnabledFor(DeleteWarningType.DeleteArchivedRecordings)) {
       setWarningModalOpen(true);
     }
     else {
       handleDeleteRecordings();
     }
-  }
+  }, [handleDeleteRecordings])
```
andrewazores, line 313: deps

src/app/Recordings/ArchivedRecordingsTable.tsx
```diff
@@ -303,25 +303,34 @@ export const ArchivedRecordingsTable: React.FunctionComponent<ArchivedRecordings
...
+
+  const handleWarningModalClose = React.useCallback(() => {
+    setWarningModalOpen(false);
+  }, [setWarningModalOpen]);
+
+  const disableDeletionDialog = () => {
```
andrewazores, line 319: `useCallback()`?

src/app/Rules/Rules.tsx
```diff
@@ -226,19 +226,19 @@ export const Rules = () => {
...
-  const handleDeleteButton = (rowData) => {
+  const handleDeleteButton = React.useCallback((rowData) => {
     if (context.settings.deletionDialogsEnabledFor(DeleteWarningType.DeleteAutomatedRules)) {
       setWarningModalOpen(true);
     }
     else {
       handleDelete(rowData, cleanRuleEnabled)
     }
-  }
+  }, [cleanRuleEnabled]);
```
andrewazores, line 236: deps

src/app/Rules/Rules.tsx
```diff
@@ -269,6 +269,18 @@ export const Rules = () => {
...
 
+  const handleWarningModalAccept = React.useCallback(() => {
+    handleDelete(rowDeleteData, cleanRuleEnabled);
+  }, [rowDeleteData, cleanRuleEnabled]);
+
+  const handleWarningModalClose = React.useCallback(() => {
+    setWarningModalOpen(false);
+  }, [setWarningModalOpen]);
+
+  const disableDeletionDialog = () => {
```
andrewazores, line 280: `useCallback()`?

src/app/SecurityPanel/StoreJmxCredentials.tsx
```diff
@@ -141,18 +141,26 @@ export const StoreJmxCredentials = () => {
     addSubscription(forkJoin(tasks).subscribe());
   };
 
-  const handleModalClose = () => {
+  const handleAuthModalClose = () => {
```
andrewazores, line 144: `useCallback()`?

src/app/SecurityPanel/StoreJmxCredentials.tsx
```diff
@@ -141,18 +141,26 @@ export const StoreJmxCredentials = () => {
...
-  const handleDeleteButton = () => {
+  const handleDeleteButton = React.useCallback(() => {
     if (context.settings.deletionDialogsEnabledFor(DeleteWarningType.DeleteJMXCredentials)) {
       setWarningModalOpen(true);
     }
     else {
       handleDeleteCredentials();
     }
-  }
+  }, [handleDeleteCredentials]);
```
andrewazores, line 155: deps

src/app/Modal/DeleteWarningModal.tsx
```diff
@@ -46,62 +46,58 @@ export interface DeleteWarningProps {
...
   const [doNotAsk, setDoNotAsk] = useState(false);
   const realWarningType = getFromWarningMap(warningType);
 
-  const description = `${realWarningType?.description}${(typeof items === 'undefined' || items.length <= 1) ?  "":"s"}: [${items?.join(", ")}]. ${realWarningType?.descriptionPartTwo}`
+  const description = 
+    <TextContent>
+      <Text>
+        {realWarningType?.description}{(typeof items === 'undefined' || items.length <= 1) ?  "":"s"}: [{items?.join(", ")}]
+        <br />       
+        {realWarningType?.descriptionPartTwo}
```
andrewazores, line 61: Can we rename this property something other than `descriptionPartTwo`? Perhaps it can even just be part of the description, if it's possible to use things like `<br>` in the description text and have that render properly (not sure). If that doesn't work directly then rather than the `description` property being just a `string`, maybe it can be `JSX.Element` and you can pass in a `<Text>` element with the intended formatting.

src/app/Modal/DeleteWarningModal.tsx
maxcao13: image
I was previously mapping the title to be the label text that's shown on the setting, should I use the id instead like this, or would you rather me just add a new field in the DeletionWarning object called `label` or something?

src/app/SecurityPanel/StoreJmxCredentials.tsx
maxcao13: I didn't even write this one, I just noticed that it had a very similar name to what I was doing :laughing: , but yeah I will change it.

src/app/Archives/AllArchivedRecordingsTable.tsx
andrewazores: Yea, sorry, it "doesn't matter" in practical rendering terms - at least, given the rest of the structure of the component, at this point in time. It might be a slight case of premature optimization to apply `useCallback` here, but I think the penalty for it is small. I would rather err on the side of applying it when it's unnecessary and adding a little extra unneeded overhead, rather than not applying it and leaving it open to bugs in the future after this component is refactored or used in a new way.

src/app/Modal/DeleteWarningModal.tsx
andrewazores: Hmm, that's fair. Just using the `id` works well enough for now, but again that'd be a deficiency later on when we get around to localizing this thing and need that to display in other languages.

I guess a `label` and a `title` property would work, where ex. `label: Delete Automated Rule` and `title: Permanently delete Automated Rule`. This looks a little redundant but it's also clear and unambiguous, and we can very cleanly extract those strings and localize them.

src/app/SecurityPanel/StoreJmxCredentials.tsx
andrewazores: Sorry again :-) I like to follow the boyscout's rule: "Always leave the code better than you found it."

Since you're working in this area at the moment, I'm holding you responsible for cleaning up things I see at the edges of the ~~campsite~~ diffs ;-)

src/app/Modal/DeleteWarningModal.tsx
andrewazores: I don't mind how that last screenshot looks. The `items?.join` isn't *that* worth holding onto here, IMO. The whole `[thing1, thing2]` is kind of nice to have, but seems ultimately unnecessary given that this modal is only going to appear after the user has explicitly selected some item(s) and then clicked a Delete button. If they clicked a select-all checkbox on a very long table of items then this dialog's list becomes really long and unwieldy, too.

The last screenshot with the `: [sdfh, sdfhfds]` removed would still totally do the job, IMO.

maxcao13:
Sorry, I used a code formatter on VSCode and it touched a lot of random places like whitespace.

andrewazores:
> Sorry, I used a code formatter on VSCode and it touched a lot of random places like whitespace.

No worries. It's best to contain that to a single commit if and when practical, but sometimes this happens. FWIW there's also `npm run format` that you can do from your shell.

maxcao13:
> > Sorry, I used a code formatter on VSCode and it touched a lot of random places like whitespace.
> 
> No worries. It's best to contain that to a single commit if and when practical, but sometimes this happens. FWIW there's also `npm run format` that you can do from your shell.

When I run `npm run format` it doesn't detect and fix things that my VSCode extension does, for example misaligned tabs or extra whitespace. Is there some sort of config file that @magic/format uses? I notice that there is a `.prettierrc`, but I'm not sure if that's it. I also notice that the format script in `package.json` runs 
```
"prettier --check --write ./src/**/*.{tsx,ts} && license-check-and-add add -f license-config.json"
```
but I don't what config it uses. Sorry this may be better as a question on a separate issue.

andrewazores:
> > > Sorry, I used a code formatter on VSCode and it touched a lot of random places like whitespace.
> > 
> > 
> > No worries. It's best to contain that to a single commit if and when practical, but sometimes this happens. FWIW there's also `npm run format` that you can do from your shell.
> 
> When I run `npm run format` it doesn't detect and fix things that my VSCode extension does, for example misaligned tabs or extra whitespace. Is there some sort of config file that @magic/format uses? I notice that there is a `.prettierrc`, but I'm not sure if that's it. I also notice that the format script in `package.json` runs
> 
> ```
> "prettier --check --write ./src/**/*.{tsx,ts} && license-check-and-add add -f license-config.json"
> ```
> 
> but I don't what config it uses. Sorry this may be better as a question on a separate issue.

What is `@magic/format`?

`prettier` uses the `.prettierrc`, that's one of the defaults it looks for and is a common config file convention.

```bash
$ npx prettier --help
Usage: prettier [options] [file/dir/glob ...]

By default, output is written to stdout.
Stdin is read if it is piped to Prettier and no files are given.

Output options:

  -c, --check              Check if the given files are formatted, print a human-friendly summary
                           message and paths to unformatted files (see also --list-different).
  -l, --list-different     Print the names of files that are different from Prettier's formatting (see also --check).
  -w, --write              Edit files in-place. (Beware!)

Format options:

  --arrow-parens <always|avoid>
                           Include parentheses around a sole arrow function parameter.
                           Defaults to always.
  --bracket-same-line      Put > of opening tags on the last line instead of on a new line.
                           Defaults to false.
  --no-bracket-spacing     Do not print spaces between brackets.
  --embedded-language-formatting <auto|off>
                           Control how Prettier formats quoted code embedded in the file.
                           Defaults to auto.
  --end-of-line <lf|crlf|cr|auto>
                           Which end of line characters to apply.
                           Defaults to lf.
  --html-whitespace-sensitivity <css|strict|ignore>
                           How to handle whitespaces in HTML.
                           Defaults to css.
  --jsx-single-quote       Use single quotes in JSX.
                           Defaults to false.
  --parser <flow|babel|babel-flow|babel-ts|typescript|espree|meriyah|css|less|scss|json|json5|json-stringify|graphql|markdown|mdx|vue|yaml|glimmer|html|angular|lwc>
                           Which parser to use.
  --print-width <int>      The line length where Prettier will try wrap.
                           Defaults to 80.
  --prose-wrap <always|never|preserve>
                           How to wrap prose.
                           Defaults to preserve.
  --quote-props <as-needed|consistent|preserve>
                           Change when properties in objects are quoted.
                           Defaults to as-needed.
  --no-semi                Do not print semicolons, except at the beginning of lines which may need them.
  --single-quote           Use single quotes instead of double quotes.
                           Defaults to false.
  --tab-width <int>        Number of spaces per indentation level.
                           Defaults to 2.
  --trailing-comma <es5|none|all>
                           Print trailing commas wherever possible when multi-line.
                           Defaults to es5.
  --use-tabs               Indent with tabs instead of spaces.
                           Defaults to false.
  --vue-indent-script-and-style
                           Indent script and style tags in Vue files.
                           Defaults to false.

Config options:

  --config <path>          Path to a Prettier configuration file (.prettierrc, package.json, prettier.config.js).
  --no-config              Do not look for a configuration file.
  --config-precedence <cli-override|file-override|prefer-file>
                           Define in which order config files and CLI options should be evaluated.
                           Defaults to cli-override.
  --no-editorconfig        Don't take .editorconfig into account when parsing configuration.
  --find-config-path <path>
                           Find and print the path to a configuration file for the given input file.
  --ignore-path <path>     Path to a file with patterns describing files to ignore.
                           Defaults to .prettierignore.
  --plugin <path>          Add a plugin. Multiple plugins can be passed as separate `--plugin`s.
                           Defaults to [].
  --plugin-search-dir <path>
                           Custom directory that contains prettier plugins in node_modules subdirectory.
                           Overrides default behavior when plugins are searched relatively to the location of Prettier.
                           Multiple values are accepted.
                           Defaults to [].
  --with-node-modules      Process files inside 'node_modules' directory.

Editor options:

  --cursor-offset <int>    Print (to stderr) where a cursor at the given position would move to after formatting.
                           This option cannot be used with --range-start and --range-end.
                           Defaults to -1.
  --range-end <int>        Format code ending at a given character offset (exclusive).
                           The range will extend forwards to the end of the selected statement.
                           This option cannot be used with --cursor-offset.
                           Defaults to Infinity.
  --range-start <int>      Format code starting at a given character offset.
                           The range will extend backwards to the start of the first line containing the selected statement.
                           This option cannot be used with --cursor-offset.
                           Defaults to 0.

Other options:

  --no-color               Do not colorize error messages.
  --no-error-on-unmatched-pattern
                           Prevent errors when pattern is unmatched.
  --file-info <path>       Extract the following info (as JSON) for a given file path. Reported fields:
                           * ignored (boolean) - true if file path is filtered by --ignore-path
                           * inferredParser (string | null) - name of parser inferred from file path
  -h, --help <flag>        Show CLI usage, or details about the given flag.
                           Example: --help write
  -u, --ignore-unknown     Ignore unknown files.
  --insert-pragma          Insert @format pragma into file's first docblock comment.
                           Defaults to false.
  --loglevel <silent|error|warn|log|debug>
                           What level of logs to report.
                           Defaults to log.
  --require-pragma         Require either '@prettier' or '@format' to be present in the file's first docblock comment
                           in order for it to be formatted.
                           Defaults to false.
  --stdin-filepath <path>  Path to the file to pretend that stdin comes from.
  --support-info           Print support information as JSON.
  -v, --version            Print Prettier version.
```

Issue #465, opened 2022-06-17
[Story] Enhance JMX Credentials UI for matchExpressions


andrewazores:
Parent: https://github.com/cryostatio/cryostat/issues/895

https://github.com/cryostatio/cryostat/pull/1000 adds a new API call that should be used for a revamped JMX Credentials view within the Security section. This structure seems well-suited to a table where each row corresponds to a `matchExpression`, and the row can be expanded to reveal a nested table of `target`s.

The `matchExpression` rows should be selectable, including with a header checkbox to select all, so that the row can be deleted. This deletion operation is partially supported in https://github.com/cryostatio/cryostat/pull/1000 - the internal call exists, but there is no `v2.2` API handler for the deletion yet.

There should also be a `Create` button to add new credentials with a user-defined `matchExpression`. This should navigate the user to a subview similar to `Automated Rules > Create` with a Match Expression Evaluator and validation, and form fields for the credential username and password. This is also partially supported in https://github.com/cryostatio/cryostat/pull/1000 , with an internal call but no API handler yet.

Tasks:
- [x] #466
- [x] #467

```bash
$ http :8181/api/v2.2/credentials
HTTP/1.1 200 OK
content-encoding: gzip
content-length: 328
content-type: application/json
vary: origin
```
```json
{
    "data": {
        "result": [
            {
                "matchExpression": "target.alias == \"es.andrewazor.demo.Main\"",
                "targets": [
                    {
                        "alias": "es.andrewazor.demo.Main",
                        "annotations": {
                            "cryostat": {
                                "HOST": "cryostat",
                                "JAVA_MAIN": "es.andrewazor.demo.Main",
                                "PORT": "9094"
                            },
                            "platform": {}
                        },
                        "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9094/jmxrmi",
                        "labels": {}
                    },
                    {
                        "alias": "es.andrewazor.demo.Main",
                        "annotations": {
                            "cryostat": {
                                "HOST": "cryostat",
                                "JAVA_MAIN": "es.andrewazor.demo.Main",
                                "PORT": "9095"
                            },
                            "platform": {}
                        },
                        "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9095/jmxrmi",
                        "labels": {}
                    },
                    {
                        "alias": "es.andrewazor.demo.Main",
                        "annotations": {
                            "cryostat": {
                                "HOST": "cryostat",
[...]

PR #475, opened 2022-06-28
feat(security): use API v2.2 JMX credentials

andrewazores:
Related to #465

switch API requests to v2.2, matchExpression-based credentials definitions


hareetd (approved):
Works well.

PR #476, opened 2022-06-30
feat(rules): implement initialDelay

andrewazores:
Related to https://github.com/cryostatio/cryostat/issues/1014
As usual, this needs to be tested along with the corresponding backend support in https://github.com/cryostatio/cryostat/pull/1015 . For submodule/CI reasons the "depends on" relationship is inverted.

Adds support for the new "initial delay" parameter for automated rule definitions.

Previously, rules only had an "archival period". So when a rule was created with period *t* - or when a new matching target appeared - at time *x* archived recordings would be copied at time *x+t*, *x+2t*, *x+3t*, ... , *x+nt*.

Now, rules can be created with both an archival period and an initial delay. If the rule is created or a target appears at time *x*, and the rule has archival period *t* and initial delay *d*, then archived recordings are copied at *x+d*, *x+d+t*, ... , *x+d+nt*.

It is also possible to have an initial delay defined >0 but the archival period of 0. This has the effect of only copying an archived recording at time *x+d*, exactly once.

If the initial delay is left as 0 and the archival period is >0, the backend will treat this the same way that it used to, with archives being copied at *x+nt*. There will be no copy created immediately at *x*.

### DIFF
--- a/src/app/Rules/Rules.tsx
+++ b/src/app/Rules/Rules.tsx
@@ -56,6 +56,7 @@ export interface Rule {
   matchExpression: string;
   eventSpecifier: string;
   archivalPeriodSeconds: number;
+  initialDelaySeconds: number;
   preservedArchives: number;
   maxAgeSeconds: number;
   maxSizeBytes: number;
@@ -101,7 +102,15 @@ export const Rules = () => {
       title: 'Archival Period',
       transforms: [
         info({
-          tooltip: 'Period in seconds. Cryostat will connect to matching targets at this interval and copy the relevant recording data into its archives. Values less than 1 prevent data from being copied into archives - recordings will be started and remain only in target JVM memory.'
+          tooltip: 'Period in seconds. Cryostat will connect to matching targets at this interval and copy the relevant recording data into its archives. Values less than 1 prevent data from being repeatedly copied into archives - recordings will be started and remain only in target JVM memory.'
+        })
+      ],
+    },
+    {
+      title: 'Initial Delay',
+      transforms: [
+        info({
+          tooltip: 'Initial delay in seconds. Cryostat will wait this amount of time before first copying recording data into its archives. Values less than 0 default to equal to the Archival Period. You can set a non-zero Initial Delay with a zero Archival Period, which will start a recording and copy it into archives exactly once after a set delay.'
         })
       ],
     },
@@ -167,9 +176,9 @@ export const Rules = () => {
     return () => window.clearInterval(id);
   }, []);
 
-  const handleSort = (event, index, direction) => {
+  const handleSort = React.useCallback((event, index, direction) => {
     setSortBy({ index, direction });
-  };
+  }, [setSortBy]);
 
   const handleCreateRule = React.useCallback(() => {
     routerHistory.push(`${url}/create`);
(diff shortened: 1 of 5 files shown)


src/test/Rules/CreateRule.test.tsx
```diff
@@ -178,4 +179,4 @@ describe('<CreateRule/>', () => {
     expect(createSpy).toHaveBeenCalledTimes(1);
```
tthvo, line 179: Wondering if we need to add a userEvent to type a value for `InitialDelaySeconds`? Either way it is still 0 as in the mock but just to be consistent with other forms?

andrewazores: Sure, good idea. It's Canada Day though, stop reviewing PRs :-)

src/app/Rules/Rules.tsx
```diff
@@ -96,7 +97,15 @@ export const Rules = () => {
...
-          tooltip: 'Period in seconds. Cryostat will connect to matching targets at this interval and copy the relevant recording data into its archives. Values less than 1 prevent data from being copied into archives - recordings will be started and remain only in target JVM memory.'
+          tooltip: 'Period in seconds. Cryostat will connect to matching targets at this interval and copy the relevant recording data into its archives. Values less than 1 prevent data from being repeatedly copied into archives - recordings will be started and remain only in target JVM memory.'
+        })
+      ],
+    },
+    {
+      title: 'Initial Delay',
+      transforms: [
+        info({
+          tooltip: 'Initial delay in seconds. Cryostat will wait this amount of time before first copying recording data into its archives. Values less than 0 default to equal to the Archival Period. You can set a non-zero Initial Delay with a zero Archival Period, which will start a recording and copy it into archives exactly once after a set delay.'
```
tthvo, line 108: > Values less than 0 default to equal to the Archival Period.

The description is for the last paragraph [above](https://github.com/cryostatio/cryostat-web/pull/476/#issue-1290602429) right? This might be mistaken as:  `initialDelaySeconds` will be set equal to the value `ArchivePeriod` if left as 0?

I am thinking maybe it could be like `Values less than 0 has no effect?` or maybe another better suggestion.

src/test/Rules/CreateRule.test.tsx
tthvo: Happy Canada Day :D i guess haha

tthvo:
I notice the first 3 forms for `CreateRule` was missing `aria-label` and using `aria-describeby` instead. Not sure if its intentional but I guess adding a label would give a better accessibility support?

src/app/Rules/Rules.tsx
andrewazores: > This might be mistaken as: initialDelaySeconds will be set equal to the value ArchivePeriod if left as 0?

[That's correct, though.](https://github.com/cryostatio/cryostat/pull/1015/files#diff-35494c48ebb256b9fb6036f153781a729d47eaa5e68cc8ed6b65664b164f16c6R78) If the backend really treated `initialDelaySeconds: 0` at face value, then that should indicate that an archived copy will be made at time `x+0`, but the backend specifically avoids that case.

tthvo: Ahhh yesss that makes sense! Maybe I got a bit mixed up there :D

tthvo:
Ahh got it! Should other forms also have `arial-describedby` too tho?

andrewazores:
> Ahh got it! Should other forms also have `arial-describedby` too tho?

https://cccaccessibility.org/web-1/web-developer-tutorials/aria-labelledby-vs-aria-describedby-vs-aria-label

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

I need to do some more reading about this. `aria-label` and `aria-labelledby` both do the same job, just with a direct label vs pointing to another element that acts as the label. `aria-describedby` provides a description rather than a label, but I'm not totally clear on when one or the other or both should be used.

hareetd (dismissed):
Works well

tthvo (approved):

andrewazores:
Merge conflict resolved and rebased.

hareetd (approved):

PR #473, opened 2022-06-23
fix(labels): edit labels for distinct recordings with identical file names

jan-law:
Fixes #442 

Uses the backend fix from https://github.com/cryostatio/cryostat/pull/971 to modify and retrieve archived recording labels saved from a specified target. 

I'll add another follow-up PR to apply the same fix to the all-archives view once Hareet merges his.

PR #478, opened 2022-07-05
fix(package.json): fix broken prettier config, add other useful scripts

### LINKED ISSUE #477: `npm run format` doesn't check or format any source files
https://github.com/cryostatio/cryostat-web/pull/463#issuecomment-1175063266

andrewazores:
Fixes #477
Depends on #473 #463 #431

- fix(package.json): quote prettier targets
- split format into format:apply
- add check-only format script
- don't re-exec npm directly, use execpath
- reorder license check/apply
- use npm-run-all
- linting also runs format check
- also lint .ts
- add script to apply lint fixes
- ci checks code format, not only license


package.json
```diff
@@ -7,18 +7,26 @@
...
+    "lint": "npm-run-all -l -p format:check eslint",
+    "format": "npm-run-all format:apply",
+    "format:check": "npm-run-all -l license:check prettier:check",
+    "format:apply": "npm-run-all -l -p license:apply prettier:apply",
+    "prettier:check": "prettier --check './src/**/*.{tsx,ts}'",
+    "prettier:apply": "prettier --write './src/**/*.{tsx,ts}'",
     "build:bundle-profile": "webpack --config webpack.prod.js --profile --json > stats.json",
-    "bundle-profile:analyze": "npm run build:bundle-profile && webpack-bundle-analyzer ./stats.json",
+    "build:bundle-analyze": "webpack-bundle-analyzer ./stats.json",
+    "bundle-profile:analyze": "npm-run-all -l build:bundle-profile build:bundle-analyze",
```
tthvo, line 29: `"bundle-profile:analyze"` still falls under `build:` right? I guess something like `build:bundle-analyze-profile` or a shorter version?

package.json
```diff
@@ -7,18 +7,26 @@
...
-    "eslint": "eslint --ext .tsx,.js ./src/",
+    "eslint": "npm-run-all eslint:check",
+    "eslint:check": "eslint --ext .ts,.tsx,.js ./src/",
+    "eslint:apply": "eslint --ext .ts,.tsx,.js --fix ./src/",
     "license:check": "license-check-and-add check -f license-config.json",
-    "lint": "npm run license-check-and-add && npm run eslint",
-    "format": "prettier --check --write ./src/**/*.{tsx,ts} && license-check-and-add add -f license-config.json",
+    "license:apply": "license-check-and-add add -f license-config.json",
+    "lint": "npm-run-all -l -p format:check eslint",
+    "format": "npm-run-all format:apply",
```
tthvo, line 22: Here `format` is alias for `format:apply` but above, `eslint` is alias for `eslint:check`? Should we make it consistent?

package.json
```diff
@@ -7,18 +7,26 @@
...
     "test:ci": "jest --maxWorkers=50%",
-    "eslint": "eslint --ext .tsx,.js ./src/",
+    "eslint": "npm-run-all eslint:check",
+    "eslint:check": "eslint --ext .ts,.tsx,.js ./src/",
+    "eslint:apply": "eslint --ext .ts,.tsx,.js --fix ./src/",
     "license:check": "license-check-and-add check -f license-config.json",
-    "lint": "npm run license-check-and-add && npm run eslint",
-    "format": "prettier --check --write ./src/**/*.{tsx,ts} && license-check-and-add add -f license-config.json",
+    "license:apply": "license-check-and-add add -f license-config.json",
+    "lint": "npm-run-all -l -p format:check eslint",
```
tthvo, line 21: I think we can use `format` for short?

andrewazores: `format` is already the alias for checking licenses and code style (indentation, whitespaces, etc). This one does that plus linting, which is the next step and kinda analogous to `spotbugs:check` on the main repo.

tthvo: oh sorry I mean within the command, we can use `format` instead of `format:check` for short :D

andrewazores: Oh, yea, that works but I just find it neater to not use `npm-run-all foo` where `foo` also starts with invoking `npm-run-all`. It doesn't really matter too much, though.

tthvo (approved):
Looks great to me! These scripts will be super helpful!

Issue #466, opened 2022-06-17
[Task] Implement nested table for `matchExpression` Credentials


andrewazores:
Parent: https://github.com/cryostatio/cryostat-web/issues/465

> https://github.com/cryostatio/cryostat/pull/1000 adds a new API call that should be used for a revamped JMX Credentials view within the Security section. This structure seems well-suited to a table where each row corresponds to a matchExpression, and the row can be expanded to reveal a nested table of targets.
>
> The matchExpression rows should be selectable, including with a header checkbox to select all, so that the row can be deleted. This deletion operation is partially supported in https://github.com/cryostatio/cryostat/pull/1000 - the internal call exists, but there is no v2.2 API handler for the deletion yet.

hareetd:
![image](https://user-images.githubusercontent.com/41028544/182946151-2ecb9409-d4b5-4a1c-9447-bd4c74915526.png)

I'm thinking of adding a count of the number of matched targets for each match expression. Would it be fine to modify the `CredentialsGetHandler` to also return such a count alongside the `id` and `matchExpression`? Or would a new GraphQL query be preferable? 

I'm following a similar workflow to the `AllTargetsArchivedRecordingsTable` when it comes to data querying. Expanding a match expression row sends a request to the `CredentialGetHandler` for the targets list.

andrewazores:
I think you can just add the count inside the response from the existing handler in this case.

PR #481, opened 2022-07-11
feat(credentials): implement form with matchExpression

### LINKED ISSUE #467: [Task] Implement enhanced Credentials creation form with Match Expression Evaluator
Parent: #465

> There should also be a Create button to add new credentials with a user-defined matchExpression. This should navigate the user to a subview similar to Automated Rules > Create with a Match Expression Evaluator and validation, and form fields for the credential username and password. This is also partially supported in https://github.com/cryostatio/cryostat/pull/1000 , with an internal call but no API handler yet.

andrewazores:
Fixes #467

- fix(credentials): use new API v2.2 credentials format
- chore(credentials): extract to subdirectory
- feat(jmx): remove cached credentials, use backend storage only
- feat(auth): add link from auth modal to /security
- feat(credentials): user can enter matchExpression for credentials
- update test
- fixup! feat(auth): add link from auth modal to /security

Test with https://github.com/cryostatio/cryostat/pull/1019 on the backend.


src/app/AppLayout/AuthModal.tsx
```diff
@@ -50,25 +52,35 @@ export interface AuthModalProps {
...
-  const handleDismiss = () => {
-    props.onDismiss();
-  };
-
-  const onSave = (username: string, password: string) => {
-    context.api.postTargetCredentials(username, password).pipe(first()).subscribe();
-    props.onSave();
-  };
+  const onSave = React.useCallback((username: string, password: string): Promise<void> => {
+    return new Promise((resolve) => {
```
tthvo, line 56: Should we use Observable pattern here too instead of Promise?

src/app/AppLayout/JmxAuthForm.tsx
```diff
@@ -52,36 +51,33 @@ export const JmxAuthForm: React.FunctionComponent<JmxAuthFormProps> = (props) =>
...
-  const handleDismiss = () => {
+  const handleDismiss = React.useCallback(() => {
     clear();
     props.onDismiss();
-  };
+  }, [clear, props.onDismiss]);
 
-  const handleKeyUp = (event: React.KeyboardEvent): void => {
+  const handleKeyUp = React.useCallback((event: React.KeyboardEvent): void => {
     if (event.keyCode === EnterKeyCode) {
```
tthvo, line 73: This line gives me `'keyCode' is deprecated` as said at https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode. Should we change it to something like:

```ts
if (event.key == 'Enter') { ... }
```

This is recommended as https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

src/app/SecurityPanel/Credentials/CreateJmxCredentialModal.tsx
```diff
@@ -74,23 +70,35 @@ export const CreateJmxCredentialModal: React.FunctionComponent<CreateJmxCredenti
...
+      <JmxAuthForm onSave={onSave} onDismiss={props.onClose}>
         <FormGroup
-          label="Target"
+          label="Match Expression"
           isRequired
-          fieldId="target-select"
-          helperTextInvalid="Select a target"
-          validated={validTarget}
+          fieldId="match-expression"
+          helperText={
```
tthvo, line 82: Should the helper text be right below the text input for match expression tho? Here it shows the helper text at the end of the form?

Screenshot from 2022-07-12 16-50-23

src/app/AppLayout/AuthModal.tsx
tthvo: Also do we need to `useSubscriptions`  for any api call on modals? Same for `src/app/SecurityPanel/Credentials/CreateJmxCredentialModal.tsx` too?

https://github.com/cryostatio/cryostat-web/blob/837a7208534191befc1b21413d333b7f0d1f1de5/src/app/utils/useSubscriptions.ts#L41

maxcao13 (commented):
In general, I'm curious if you have seen the consensus on using React.FunctionComponent here https://github.com/facebook/create-react-app/pull/8177 .

It seems that the React developers say that using React.FC is unnecessary because of the issues they detail. Wanted to get your thoughts on that.

maxcao13 (commented):
image

I'm wondering if there should also be the hint like in the Automated Rule create wizard? For me I didn't know what a match expression was supposed to look like so it may be helpful.

src/app/SecurityPanel/Credentials/StoreJmxCredentials.tsx
```diff
@@ -278,8 +252,6 @@ export const StoreJmxCredentials = () => {
 
 export const StoreJmxCredentialsCard: SecurityCard = {
   title: 'Store JMX Credentials',
-  description: `Targets for which Cryostat has stored JMX credentials are listed here.
-    If a Target JVM requires JMX authentication, Cryostat will use stored credentials
-    when attempting to open JMX connections to the target.`,
+  description: `Credentials that Cryostat uses to connect to target JVMs over JMX are stored here. Expand an item to see which known target applications the credential applies to.`,
```
tthvo, line 255: We cannot see the expanded yet right?

src/test/SecurityPanel/Credentials/StoreJmxCredentials.test.tsx
```diff
@@ -237,19 +223,12 @@ describe('<StoreJmxCredentials />', () => {
...
 
-    expect(screen.getByText('anotherTarget')).toBeInTheDocument();
-    expect(screen.getByText('fooTarget')).toBeInTheDocument();
+    expect(screen.getByText(mockAnotherCredential.matchExpression)).toBeInTheDocument();
+    expect(screen.getByText(mockCredential.matchExpression)).toBeInTheDocument();
 
     const checkboxes = screen.getAllByRole('checkbox');
     const selectAllCheck = checkboxes[0];
     userEvent.click(selectAllCheck);
     userEvent.click(screen.getByText('Delete'));
```
tthvo, line 232: I wonder if we also need to do some `expect` to check if api calls are made?

src/test/SecurityPanel/Credentials/StoreJmxCredentials.test.tsx
```diff
@@ -98,26 +94,18 @@ jest.mock('@app/Shared/Services/NotificationChannel.service', () => {
...
   return {
     ApiService: jest.fn(() => {
       return {
-        getStoredJmxCredentials: jest
+        deleteCredentials: jest
+          .fn()
+          .mockReturnValue(of(true)),
+        getCredentials: jest
           .fn()
           .mockReturnValueOnce(of([mockCredential])) // 'renders correectly'
```
tthvo, line 106: typo? I guess its just a comment anyway :D

src/app/AppLayout/AuthModal.tsx
andrewazores: > Should we use Observable pattern here too instead of Promise? Or is there a reason for using Promise tho?

Promise just seemed a little simpler, since this behaviour only needs to specify a callback that occurs once, not a potentially recurring observable stream. I could change it over to Observable if you prefer.

> Also do we need to useSubscriptions for any api call on modals? Same for src/app/SecurityPanel/Credentials/CreateJmxCredentialModal.tsx too?

Good catch. I'll fix that.

src/app/SecurityPanel/Credentials/CreateJmxCredentialModal.tsx
andrewazores: image

src/app/SecurityPanel/Credentials/StoreJmxCredentials.tsx
andrewazores: Not yet, that's https://github.com/cryostatio/cryostat-web/issues/466 . I can alter the text here so that it isn't referencing that feature that isn't implemented just yet.

andrewazores:
> In general, I'm curious if you have seen the consensus on using React.FunctionComponent here [facebook/create-react-app#8177](https://github.com/facebook/create-react-app/pull/8177) .
> 
> It seems that the React developers say that using React.FC is unnecessary because of the issues they detail. Wanted to get your thoughts on that.

I haven't seen that before. I don't really have any thoughts on the matter or opinion one way or the other.

https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/

I tend to like the explicit/static types where possible, but sometimes it isn't really necessary - and if the type signature is actually lying to you, like it is about the implicit `children`, then it does more harm than good.

If you'd like to go through and refactor to remove all those type declarations later then I wouldn't be opposed.

src/app/SecurityPanel/Credentials/CreateJmxCredentialModal.tsx
tthvo: Looks great to me :D

src/app/AppLayout/AuthModal.tsx
tthvo: Ohh I think it makes sense to use Promise for simpler use :D 

Also do you the two calls to `context.target.target` and `context.api.postCredentials` can be piped using `mergeMap` like here? That way we only 1 subscription added to `addSubcription`. Just seeing which style we prefer^^

https://github.com/cryostatio/cryostat-web/blob/837a7208534191befc1b21413d333b7f0d1f1de5/src/app/Rules/RulesUploadModal.tsx#L108

andrewazores: Sure, good idea.

src/app/Shared/MatchExpressionEvaluator.tsx
```diff
@@ -125,13 +120,28 @@ export const MatchExpressionEvaluator: React.FunctionComponent<MatchExpressionEv
       </StackItem>
       <StackItem>
         { statusLabel }
+        {
+          !props.inlineHint ?
+            <Tooltip content={<div>Hint: try an expression like <br />{ exampleExpression }</div>}>
+              <HelpIcon noVerticalAlign />
```
tthvo, line 126: Maybe some space between the help icon and the status label?

Screenshot from 2022-07-13 11-50-53

andrewazores: image

tthvo: Looks good!

maxcao13 (approved):
Other than the match expression hint thing, which is up to you of course, looks great. 
Good work!

andrewazores:
> Other than the match expression hint thing, which is up to you of course, looks great. Good work!

There is a hint now, but it's tucked away inside of a tooltip icon so that the modal contents aren't too large and the user doesn't have to scroll too much.

maxcao13:
> I haven't seen that before. I don't really have any thoughts on the matter or opinion one way or the other.
> 
> https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/
> 
> I tend to like the explicit/static types where possible, but sometimes it isn't really necessary - and if the type signature is actually lying to you, like it is about the implicit children, then it does more harm than good.
> 
> If you'd like to go through and refactor to remove all those type declarations later then I wouldn't be opposed.

Maybe I will do a refactor in the future, but for now I think it may be best to stop using React.FC in `-web` in the future and stick with the usual 
```
const Foo = ( {prop1, prop2}: FooProps ): JSX.Element => {
   ...
   return ...
}
``` 
so that everything is using the same and preferred style. 

If that's okay with you anyways. I'm not sure what is the best way to let everyone know though, haha.

andrewazores:
> I'll open an issue for it, if that's okay with you anyways. This isn't a big deal by any means ^^. I'm not sure what is the best way to let everyone know though, haha.

The best way is probably just to remove it from the codebase. New code added will naturally tend to follow the existing style found as new developers onboard and use what they read as a reference for what they write. We could also add it to a style guide document, but if it's just a written suggestion and not enforced by formatting/linting tools then it's likely to just get ignored or missed.

tthvo:
@andrewazores I just have a quick question. 

If I enter the form with the exact same inputs (i.e. match expression, username, password - I guess this could be duplicate) multiple times, the credentials are appended as new. Is this intentional?

andrewazores:
Not exactly intentional, but harder to guard against than it seems.

The username and password can definitely be duplicated across different definitions.

The `matchExpression` should be unique, conceptually, but that's not as easy as it sounds.

```
A || B
A    ||    B
A||B
!(!A && !B)
A || (A || B)
```
are all "the same expression", but we won't catch those by simply doing `string1.equals(string2)`.

We could respond with a `400` and reject cases where the exact same `matchExpression` is entered a second time, where the String representation is identical, but that's only a tiny sliver of semantically duplicated `matchExpression`s that could be entered.

If we could catch all of those cases then for sure, I would want to reject semantically duplicated expressions. But right now we don't have any engine for parsing and evaluating the semantics of the expression to determine logical equivalence, so I leave it as permissive and allow the user to do silly things by duplicating expressions, rather than explicitly rejecting only certain cases of identical strings.

tthvo:
Ahh make sense! It is really hard to check those.

> But right now we don't have any engine for parsing and evaluating the semantics of the expression to determine logical equivalence

Have we considered any engine yet?

andrewazores:
No, and to be honest I'm not even sure such a thing exists for this use case. This is a kind of formal verification, which is an advanced topic, and JavaScript (which is what our matchExpressions really are) seems particularly hard to formally verify because of all its dynamic behaviours and types. And if it does exist, we would need it to be available as a Java package we can import as a dependency, or at least as some C library we can call out to over JNI. That's a very tight set of requirements to find anything that fits, and if it does exist I'm not sure it's even worth the effort/cost.

andrewazores:
Oh... and I forgot about some more "easy" and "obvious" permutations too!

```
A || B
A    ||    B
A||B
!(!A && !B)
A || (A || B)

B || A
B || A || false
(A || B) && true
```

You can almost fake a solution by testing the new `matchExpression` against all known targets and creating a `Set<ServiceRef>` where the set contains all and only `ServiceRef`s where the expression evaluates truthy. Then, test every other already-known `matchExpression` in storage the same way. The new expression appears to be a duplicate of an existing expression if the `Set<ServiceRef>` `.equals()` each other. But if you have a lot of known targets and/or a lot of already defined `matchExpressions` this gets really costly. And even worse, it still isn't correct, because if for example I'm running in OpenShift and have labelled every Pod I have with `mylabel=foo`, then the expression `target.labels.mylabel === \"foo\"` and the expression `true` appear to be semantically equivalent!

So the real solution is to parse the expressions and create something like an abstract syntax tree. You also need to analyze the tree and prune any tautological branches like `|| false` and `&& true` which are trivial and add no semantic meaning to the expression. Then you need to compare the pruned trees for each existing expression to the tree for the new expression, allowing for things like `A || B` and `B || A` to be equivalent (logical OR operator with propositions A and B, order doesn't matter because OR is commutative).

Maybe you can propose this as a final year research project in a formal verification or compilers course :-)

tthvo:
Yehhh that is very niceee ideaa! I have been actually thinking about this issue for a while but never gets to play around with it (too much thinking haha). I guess maybe I could do it as a side-quest every now and then too :rofl:

PR #483, opened 2022-07-15
fix(rules): automated rule creation now list eventTemplates based on target selector

### LINKED ISSUE #433: [Task] Automated Rules form should list Event Templates for the selected target
Not sure if this is a bug or not, but if you start an automated rule using the web UI and smoketest with MatchExpression `true` and `template.name: Cryostat`, I only see one recording started on the Cryostat container. Is this intentional?

maxcao13:
Fixes #433 

It works but if a user clicks on a certain target and if there's any errors with SSL authentication or JMX authentication `e.g. HTTP 427: JMX Authentication Failure`, the `ServiceApi` will stop working and get stuck on the error, and hence subsequent clicks on the dropdown targets won't do anything 

I'm not sure why I even get these errors since I'm running 
```
CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true CRYOSTAT_CORS_ORIGIN=http://localhost:9000 sh smoketest.sh
```

### DIFF
--- a/src/app/Rules/CreateRule.tsx
+++ b/src/app/Rules/CreateRule.tsx
@@ -170,7 +170,7 @@ const Comp = () => {
     archivalPeriod, archivalPeriodUnits, initialDelay, initialDelayUnits,
     preservedArchives, maxAge, maxAgeUnits, maxSize, maxSizeUnits]);
 
-  // FIXME Error 427 JMX Authentication is handled differently than Error 502 Untrusted SSL
+  // FIXME Error 427 JMX Authentication is handled differently than Error 502 Untrusted SSL.
   const refreshTemplateList = React.useCallback(() => {
     addSubscription(
       context.target.target()
@@ -208,7 +208,6 @@ const Comp = () => {
 
   return (
     <BreadcrumbPage pageTitle='Create' breadcrumbs={breadcrumbs} >
-      <Button onClick={() => {context.api.doGet<EventTemplate[]>(`targets/localhost:0/templates`).subscribe()}}/>
       <Grid hasGutter>
         <GridItem xl={7}>
           <Card>
--- a/src/app/Rules/CreateRule.tsx
+++ b/src/app/Rules/CreateRule.tsx
@@ -38,7 +38,7 @@
 import * as React from 'react';
 import { ActionGroup, Button, Card, CardBody, CardHeader, CardHeaderMain, Form, FormGroup, FormSelect, FormSelectOption, Grid, GridItem, Split, SplitItem, Text, TextInput, TextVariants, ValidatedOptions } from '@patternfly/react-core';
 import { useHistory, withRouter } from 'react-router-dom';
-import { first } from 'rxjs/operators';
+import { concatMap, filter, first, mergeMap, toArray} from 'rxjs/operators';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { NotificationsContext } from '@app/Notifications/Notifications';
 import { BreadcrumbPage, BreadcrumbTrail } from '@app/BreadcrumbPage/BreadcrumbPage';
@@ -47,6 +47,7 @@ import { EventTemplate } from '../CreateRecording/CreateRecording';
 import { Rule } from './Rules';
 import { MatchExpressionEvaluator } from '../Shared/MatchExpressionEvaluator';
 import { FormSelectTemplateSelector } from '../TemplateSelector/FormSelectTemplateSelector';
+import { NO_TARGET, Target } from '@app/Shared/Services/Target.service';
 
 // FIXME check if this is correct/matches backend name validation
 export const RuleNamePattern = /^[\w_]+$/;
@@ -62,6 +63,7 @@ const Comp = () => {
   const [description, setDescription] = React.useState('');
   const [matchExpression, setMatchExpression] = React.useState('');
   const [matchExpressionValid, setMatchExpressionValid] = React.useState(ValidatedOptions.default);
+  const [target, setTarget] = React.useState(undefined as Target | undefined);
   const [templates, setTemplates] = React.useState([] as EventTemplate[]);
   const [template, setTemplate] = React.useState(null as string | null);
   const [templateType, setTemplateType] = React.useState(null as string | null);
@@ -172,9 +174,13 @@ const Comp = () => {
   // or at least make them write the name manually and choose TARGET/CUSTOM from a dropdown?
   React.useEffect(() => {
     addSubscription(
-      context.api.doGet<EventTemplate[]>(`targets/localhost:0/templates`).subscribe(setTemplates)
+      context.target.target()
+      .pipe(
+        filter(target => target !== NO_TARGET),
+        concatMap(target => context.api.doGet<EventTemplate[]>(`targets/${encodeURIComponent(target.connectUrl)}/templates`)),
+      ).subscribe(setTemplates)
     );
-  }, [addSubscription, context, context.api]);
+  }, [addSubscription, context, context.api, context.target]);
 
   const breadcrumbs: BreadcrumbTrail[] = [
     {
--- a/src/app/Rules/CreateRule.tsx
+++ b/src/app/Rules/CreateRule.tsx
@@ -170,8 +170,7 @@ const Comp = () => {
     archivalPeriod, archivalPeriodUnits, initialDelay, initialDelayUnits,
     preservedArchives, maxAge, maxAgeUnits, maxSize, maxSizeUnits]);
 
-  // FIXME we query ourselves to populate the list of templates, since Rules can apply to any target. Is this better than making the user write the event specifier manually,
-  // or at least make them write the name manually and choose TARGET/CUSTOM from a dropdown?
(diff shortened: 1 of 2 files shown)


andrewazores:
> I'm not sure why I even get these errors since I'm running

The `HTTP 427` you're seeing is about Cryostat failing to connect to the selected target JVM.

> CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true CRYOSTAT_CORS_ORIGIN=http://localhost:9000 sh smoketest.sh

These environment variables are telling Cryostat to start itself up with its own webserver running in HTTP mode rather than HTTPS (`DISABLE_SSL`), to not require JMX credentials when JMX clients attempt to connect to it (including itself as the connecting client - `DISABLE_JMX_AUTH`), and to allow cross-origin connections (needed for the `cryostat-web` devserver to properly interact, `CORS_ORIGIN`). If the target JVM you're connecting to requires JMX credentials then that's that, they're required, you can't disable credentials on a different target by asking Cryostat not to use credentials ;)

> It works but if a user clicks on a certain target and if there's any errors with SSL authentication or JMX authentication e.g. HTTP 427: JMX Authentication Failure, the ServiceApi will stop working and get stuck on the error, and hence subsequent clicks on the dropdown targets won't do anything

Is this problem reproducible on `main` without any changes applied? If that's an existing bug then we need to file and address it separately.

src/app/Rules/CreateRule.tsx
```diff
@@ -62,6 +64,7 @@ const Comp = () => {
   const [description, setDescription] = React.useState('');
   const [matchExpression, setMatchExpression] = React.useState('');
   const [matchExpressionValid, setMatchExpressionValid] = React.useState(ValidatedOptions.default);
+  const [target, setTarget] = React.useState(undefined as Target | undefined);
```
andrewazores, line 67: I think you could probably default this to the `NO_TARGET` special value, rather than `undefined`. If so then I think that'd be preferred since you don't need to worry about null/undefined dereferences in that case.

src/app/Rules/CreateRule.tsx
```diff
@@ -170,11 +173,25 @@ const Comp = () => {
...
   // FIXME we query ourselves to populate the list of templates, since Rules can apply to any target. Is this better than making the user write the event specifier manually,
   // or at least make them write the name manually and choose TARGET/CUSTOM from a dropdown?
+
   React.useEffect(() => {
     addSubscription(
-      context.api.doGet<EventTemplate[]>(`targets/localhost:0/templates`).subscribe(setTemplates)
+      context.target.target()
+      .pipe(
+        concatMap(target => 
+          iif(
```
andrewazores, line 182: Nice use of `iif` here, very clever.

If the user has not selected any target (ie it's null/undefined/`NO_TARGET`), do we need to show anything at all in the template selection dropdown? I think we could just leave the dropdown empty instead.

maxcao13: Right, I was gonna ask about that. If we do that, then we force the user to use the `Match Expression Evaluator` to select a target to create a rule. 

> Now, new question: what templates do we display when the user has not selected any target? Perhaps they do not currently have any target to select, but they want to create an Automated Rule in advance before deploying their application.
> 
> In that case perhaps we should query ourselves and filter out the one with `Provider: Cryostat`

Going off what you said in the related issue, it seems this would entail different implementation and functionality than what you said before, unless I'm misinterpreting.

maxcao13:
> > I'm not sure why I even get these errors since I'm running
> 
> The `HTTP 427` you're seeing is about Cryostat failing to connect to the selected target JVM.
> 
> > CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true CRYOSTAT_CORS_ORIGIN=http://localhost:9000 sh smoketest.sh
> 
> These environment variables are telling Cryostat to start itself up with its own webserver running in HTTP mode rather than HTTPS (`DISABLE_SSL`), to not require JMX credentials when JMX clients attempt to connect to it (including itself as the connecting client - `DISABLE_JMX_AUTH`), and to allow cross-origin connections (needed for the `cryostat-web` devserver to properly interact, `CORS_ORIGIN`). If the target JVM you're connecting to requires JMX credentials then that's that, they're required, you can't disable credentials on a different target by asking Cryostat not to use credentials ;)

Oh gotcha, my bad didn't know that, thanks!

> 
> > It works but if a user clicks on a certain target and if there's any errors with SSL authentication or JMX authentication e.g. HTTP 427: JMX Authentication Failure, the ServiceApi will stop working and get stuck on the error, and hence subsequent clicks on the dropdown targets won't do anything
> 
> Is this problem reproducible on `main` without any changes applied? If that's an existing bug then we need to file and address it separately.

Yeah, so I just tried it with `cryostat -> HEAD` and `cryostat/web-client -> HEAD` 
running `CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true CRYOSTAT_CORS_ORIGIN=http://localhost:9000/ sh smoketest.sh` and `npm run start:dev`

When I select `es.andrewazor.demo.Main` on the dashboard, it gives a `Authentication Required` modal
 and requires a Username and Password, and in the terminal, it gives the 
```
WARNING: HTTP 427: JMX Authentication Failure
io.vertx.ext.web.handler.HttpException: Client Error (427)
Caused by: org.openjdk.jmc.rjmx.ConnectionException caused by java.lang.SecurityException: Authentication failed! Invalid username or password
```
before I even enter anything, and when I try to save the credentials that are autofilled, I get `405 Method Not Allowed` and the logger says its for 
```
Jul 18, 2022 5:44:09 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:45488): OPTIONS /api/v2.2/credentials 204 0ms
Jul 18, 2022 5:44:09 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:45494): POST /api/v2.2/credentials 405 0ms
```

and sort of the same thing will happen in the `Automated Rule Creation` since we try to get a connection to the target before query it's templates.

src/app/Rules/CreateRule.tsx
andrewazores: Ah fair, good point. Okay, what you have now makes sense.

andrewazores:
Weird, `405 Method Not Allowed` on the brand-new credentials endpoint shouldn't be happening. Are you your backend and frontend branches up to date with the upstream mains?

maxcao13:
Yup, I believe so.
image

maxcao13:
This is how I can reproduce it.

https://user-images.githubusercontent.com/69827283/179609257-92170ec6-4ac8-4c54-9c2e-13d3bd5b6ae7.mov

maxcao13:
Okay, I finally figured it out, I need to run `mvn -Dheadless=true clean package` to update my code change when running `smoketest.sh` or `run.sh`. I usually never do since I always use `devserver.sh` which does it for me, so I just forgot. :(

I realized after I got HTTP 406 from viewing reports on main. Sorry about that, however my previous error is still an issue
> if a user clicks on a certain target and if there's any errors with SSL authentication or JMX authentication e.g. HTTP 427: JMX Authentication Failure, the ServiceApi will stop working and get stuck on the error, and hence subsequent clicks on the dropdown targets won't do anything

maxcao13:
Not sure why, but `Observables<EventTemplate[]>` that are subscribed from `refreshTemplateList` that run into `Error 427: Client Error` never complete until a a different valid target is queried, then both are completed (can see from console). But Observables that run into `Error 502: Bad Gateway` are caught by the `catchError` that is implemented and hence get completed correctly.

andrewazores:
I see. That might just be a browser implementation thing. The browser itself does some level of processing of requests before we ever see them in application code, especially for `4xx` error codes, so that might just explain the difference in behaviour. If there's something we can do to work around it that would be nice, but that's an issue for another PR.

Could you file a new issue with a description of the behaviour you've found? We probably won't get to it for some time but it'll be good to have it on the back burner.

maxcao13:
Should we wait for #485, or can this be merged as it is? If there are no errors it works as intended, but if there is JMX Auth errors by selecting on an unauth'd target, the dropdown will show stale templates from the previous target, but it can be solved by just simply selecting a different target.

andrewazores:
We can proceed without #485.

andrewazores (approved):

andrewazores (changes requested):
Before:
image

After:
image

This little blue empty button thing under the breadcrumb shouldn't be there, should it?

src/app/Rules/CreateRule.tsx
```diff
@@ -185,6 +208,7 @@ const Comp = () => {
 
   return (
     <BreadcrumbPage pageTitle='Create' breadcrumbs={breadcrumbs} >
+      <Button onClick={() => {context.api.doGet<EventTemplate[]>(`targets/localhost:0/templates`).subscribe()}}/>
```
andrewazores, line 211: This looks like an addition for manual testing or something, not something actually needed for the feature, right?

maxcao13: Ahh yea, I really forgot about that.

andrewazores (approved):
Great work, thanks.

PR #484, opened 2022-07-18
fix(templates): delete custom event template no longer deletes default templates with similar name

### LINKED ISSUE #472: [Task] Deleting Custom Event Template with similar Name to Default Event Template deletes both
It seems that when I delete a custom event template with the same name as a Default Target Event Templates e.g. (Continuous/Profiling), it will delete both. 

1. Start Cryostat, navigate to Events, and upload unzipped [profile.jfc](https://github.com/cryostatio/cryostat-web/files/8960774/profile.zip)
2. Delete the custom Event template.

Seems that the front-end just sends one DELETE request to the v1 templates/{encoded_templateName} endpoint so it should only be deleting 1 file on the back-end, so I'm assuming something about how the notifications maintain state on the front-end is causing this.

maxcao13:
Fixes #472

The only thing that I'm thinking is if it's better to test every field of a `Template` object instead of just filtering `name` and `type`
i.e. `if (o.name != v.message.template.name) || type != type || provider != provider || description != description ` or make some helper for that.

I'm thinking not, since users can only create `CUSTOM` templates for now and cannot create new custom event templates with the same name, because of this `Event template "Continuous" already exists caused by MutableTemplateService.InvalidEventTemplateException: Event template "Continuous" already exists`

### DIFF
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -136,7 +136,7 @@ export const EventTemplates = () => {
   React.useEffect(() => {
     addSubscription(
       context.notificationChannel.messages(NotificationCategory.TemplateDeleted)
-        .subscribe(v => setTemplates(old => old.filter(o => o.name != v.message.template.name)))
+        .subscribe(v => setTemplates(old => old.filter(o => (o.name != v.message.template.name || o.type != v.message.template.type))))
     )
   }, [addSubscription, context, context.notificationChannel, setTemplates]);
 


andrewazores:
> I'm thinking not, since users can only create CUSTOM templates for now and cannot create new custom event templates with the same name, because of this Event template "Continuous" already exists caused by MutableTemplateService.InvalidEventTemplateException: Event template "Continuous" already exists

Yep, the only two kinds of template will be `TARGET` (predefined by the target JVM) and `CUSTOM` (uploaded to and retained by Cryostat). Ones that come with the JVM (ex `ls /usr/lib/jvm/java/lib/jfr` for the ones that come with your system default JDK) are always `TARGET`, and ones created via Cryostat API are always `CUSTOM`. So just testing the `name` and `type` is enough for uniqueness, and those are the two "dimensions" of the template definition that we use to uniquely ID them in the API implementation anyway.

andrewazores (approved):

PR #486, opened 2022-07-21
feat(recordings): Add search filters to target recording tables

### LINKED ISSUE #414: [Task] Search/filters in Recordings tables

jan-law:
Fixes #414 

Filter recordings by name, state, labels, duration, and start time. When a filter is selected, only the recordings matching all the filter criteria will be visible in the active or target-specific archived recordings table. If multiple queries for the same filter category are specified, eg `state: RUNNING, state: STOPPED`, the results will show recordings matching either of the queries, eg any recording that is either `RUNNING` or `STOPPED`.

### DIFF
--- a/src/app/Archives/AllArchivedRecordingsTable.tsx
+++ b/src/app/Archives/AllArchivedRecordingsTable.tsx
@@ -301,8 +301,6 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
         onHeaderCheck={handleHeaderCheck}
         isLoading={isLoading}
         isEmpty={!recordings.length}
-        isEmptyFilterResult={false}
-        clearFilters={() => {/** TODO */}}
         errorMessage=''
     >
       {recordingRows}
--- a/src/app/Archives/AllArchivedRecordingsTable.tsx
+++ b/src/app/Archives/AllArchivedRecordingsTable.tsx
@@ -301,6 +301,8 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
         onHeaderCheck={handleHeaderCheck}
         isLoading={isLoading}
         isEmpty={!recordings.length}
+        isEmptyFilterResult={false}
+        clearFilters={() => {/** TODO */}}
         errorMessage=''
     >
       {recordingRows}
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -345,16 +345,16 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
       filtered = filtered.filter((rec) => {
         const start = rec.startTime;
         const stop = rec.state === RecordingState.RUNNING ? new Date().getTime() : rec.startTime + rec.duration;
-        return !!filters.DateRange.filter((range) => {
+        return filters.DateRange.filter((range) => {
           const window = range.split(' to ');
           const beginning = new Date(window[0]).getTime();
           const end = new Date(window[1]).getTime();
           const isInDateRange =
-            (start >= beginning && start <= end) || // starts in date range
-            (stop >= beginning && stop <= end) || // stops in date range
+            (start >= beginning && start <= end) || // starts within date range
+            (stop >= beginning && stop <= end) || // stops within date range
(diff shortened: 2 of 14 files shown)


jan-law:
I had a few questions about the filter design:

1. Are there any plans to merge the Active recordings table and Archived recordings table for a selected target? As of now, I've programmed the recording `Name` filter to auto-complete with both active and archived recording names for the selected target and reusing the same filter component for both tables.

>  Before/after a datetime would also be nice.
2. Did you mean 'show me all recordings started within the specified datetime range' or 'show me all recordings that were running during the specified datetime range'? If it's the latter, I'd need to determine if a manually stopped continuous recording was running at a certain time, because the `duration` field is `0` and there's no `stopTime` field.

3. For the date/time range, hours and minutes selectors add precision but also add more clutter to the UI. How precise of a time range would users find most helpful?

andrewazores:
> Are there any plans to merge the Active recordings table and Archived recordings table for a selected target? As of now, I've programmed the recording Name filter to auto-complete with both active and archived recording names for the selected target and reusing the same filter component for both tables.

Those two tables might get merged eventually, but I wouldn't say there are any "plans" to do so.

> Did you mean 'show me all recordings started within the specified datetime range' or 'show me all recordings that were running during the specified datetime range'?

I didn't necessarily mean a `[start, end]` timestamp range provided by the user, though that would eventually be nice too. I was just thinking a "start time before X" filter and a "start time after X" filter.

> If it's the latter, I'd need to determine if a manually stopped continuous recording was running at a certain time, because the duration field is 0 and there's no stopTime field.

Don't worry about that for now. That's the kind of thing we should be tracking in a proper database for recordings, so we can have a model for the recording other than the metadata supplied to us by the target JVM. Then we can update it ourselves when we know that we explicitly stopped it.

> For the date/time range, hours and minutes selectors add precision but also add more clutter to the UI. How precise of a time range would users find most helpful?

I think what I see in the branch now is fine. The calendar and clock selectors that appear let me select from dropdowns with reasonable precision, and for extra precision I can type in a `hh:mm` string. Being able to type in `hh:mm:ss` might be good too, but no need for there to be any to-the-second selections in the dropdown.

jan-law:
> Being able to type in hh:mm:ss might be good too

If you want to include seconds in the `TimePicker`, we'd need to upgrade patternfly to the [2022-02-22 release](https://www.patternfly.org/v4/developer-resources/release-notes/#2022.02-release-notes-2022-02-22) to use the `includeSeconds` prop. I'll leave the filter as hh:mm for now.

jan-law:
This PR is ready for review. If I have enough time leftover before the end of my internship, I'll add recording filter unit tests in a separate PR.

andrewazores:
One seeming bug found in manual testing. If I choose the StartedAfterDate filter, select a date from the calendar picker, and then click the search magnifying glass icon, it actually applies a StartedBeforeDate filter with the chosen date.

src/app/Archives/AllArchivedRecordingsTable.tsx
```diff
@@ -301,6 +301,8 @@ export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedReco
         onHeaderCheck={handleHeaderCheck}
         isLoading={isLoading}
         isEmpty={!recordings.length}
+        isEmptyFilterResult={false}
+        clearFilters={() => {/** TODO */}}
```
andrewazores, line 305: Is there still something TODO here?

andrewazores:
Implementation looks good.

> If I have enough time leftover before the end of my internship

:-(

> I'll add recording filter unit tests in a separate PR.

Please file a follow-up issue after this is merged so it isn't forgotten. If you have time to take it on then great, otherwise somebody else can pick it up later.

src/app/Archives/AllArchivedRecordingsTable.tsx
jan-law: Oh yeah, originally I was going to add the search filter for the All Archives View, then we decided to wait until Hareet merged his PR to prevent conflicts. It looks like his PR already has a search feature as of now - I can revert my changes on the `AllArchivedRecordingsTable`.

andrewazores (approved):

PR #487, opened 2022-07-26
feat(recording): include helperText for invalid recording names when creating new recordings with form

### LINKED ISSUE #402: Improve recording name validation
When users enter a recording name containing invalid characters (eg hyphens or whitespace) on the Custom Flight Recording form, the validation error `!` appears in the field without any other text. It would be nice to somehow indicate what caused the validation error.

It looks like the code intended to send an error notification after someone presses the `Create` button, except the `Create` button is disabled while the name field is invalid.

https://github.com/cryostatio/cryostat-web/blob/main/src/app/CreateRecording/CustomRecordingForm.tsx#L154
```
const handleSubmit = () => {
    const notificationMessages: string[] = [];
    if (nameValid !== ValidatedOptions.success) {
      notificationMessages.push(`Recording name ${recordingName} is invalid`);
    }
```

https://github.com/cryostatio/cryostat-web/blob/main/src/app/CreateRecording/CustomRecordingForm.tsx#L327
`<Button variant="primary" onClick={handleSubmit} isDisabled={nameValid !== ValidatedOptions.success || !template || !templateType}>Create</Button>`

image

maxcao13:
Fixes #402

Thoughts or anything I missed? 

One thing I was wondering about was that, if users put 0 in the `Duration` input, the recording will be set to `Continuous` and I'm wondering if there's any indication that that is the case on the UI somewhere, and if not, if that should be included as well.

### DIFF
--- a/src/app/CreateRecording/CustomRecordingForm.tsx
+++ b/src/app/CreateRecording/CustomRecordingForm.tsx
@@ -48,6 +48,7 @@ import { DurationPicker } from '@app/DurationPicker/DurationPicker';
 import { FormSelectTemplateSelector } from '../TemplateSelector/FormSelectTemplateSelector';
 import { RecordingLabel } from '@app/RecordingMetadata/RecordingLabel';
 import { RecordingLabelFields } from '@app/RecordingMetadata/RecordingLabelFields';
+import { useSubscriptions } from '@app/utils/useSubscriptions';
 
 export interface CustomRecordingFormProps {
   onSubmit: (recordingAttributes: RecordingAttributes) => void;
@@ -59,6 +60,7 @@ export const CustomRecordingForm = (props) => {
   const context = React.useContext(ServiceContext);
   const notifications = React.useContext(NotificationsContext);
   const history = useHistory();
+  const addSubscription = useSubscriptions();
 
   const [recordingName, setRecordingName] = React.useState(props.recordingName || props?.location?.state?.recordingName || '');
   const [nameValid, setNameValid] = React.useState(ValidatedOptions.default);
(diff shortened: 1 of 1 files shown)


andrewazores:
Looks good.

> One thing I was wondering about was that, if users put 0 in the Duration input, the recording will be set to Continuous and I'm wondering if there's any indication that that is the case on the UI somewhere, and if not, if that should be included as well.

I don't think there is any other indication of that fact in the UI. I think it might be a good enough fix to not allow the number input to accept values less than 1. That combined with the fact that the input control becomes disabled when the Continuous box is checked seems pretty clear. Perhaps checking the Continuous box should also set the duration value to 0.

jan-law (commented):
Name validation looks good!

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -224,8 +233,10 @@ export const CustomRecordingForm = (props) => {
       <FormGroup
         label="Duration"
         isRequired
-        fieldId="recording-duration"
+        validated={durationValid}
         helperText="Time before the recording is automatically stopped"
+        helperTextInvalid="A recording may only have a non-zero integer duration"
```
andrewazores, line 238: `non-zero integer` would include negatives :-)

Just say `positive` here, I think, and maybe add a bit of description about Continuous recordings never being (automatically) stopped.

Is it possible to have the number input prevent 0 from being entered? It already won't let me put in a negative number or non-numeric value.

maxcao13: > Is it possible to have the number input prevent 0 from being entered? It already won't let me put in a negative number or non-numeric value.

I think this is a feature of patternfly's `TextInput` `type` prop which is set to `"number"`, which only allows positive numbers and decimals internally but maybe I can figure something out to override it.

maxcao13: So, there is not really an easy way to do this that I can see. The patternfly `TextInput` prop `type='number'` forces the input to always contain a number so if a user deletes everything in the input, the input while never be empty. I'm tried
```
  const handleDurationChange = React.useCallback((evt) => {
    if (Number(evt) == 0) {
      return;
    }
    setDuration(Number(evt));
    setDurationValid(DurationPattern.test(evt) ? ValidatedOptions.success : ValidatedOptions.error);
  }, [setDurationValid, setDuration]);
```
so that the input just doesn't allow the full number 0 to be in the input, but it feels annoying if a user tries to clear the field to type a new number since it wouldn't allow it to be cleared at all.

So one way to get past this is to just not use the `type` prop and do manual `onChange` checking so that users can actually clear the field, but still manually only allow numbers to be input. 

Thoughts or clarifications?

andrewazores: Okay, don't worry about it then. Removing the `type` prop would make things even worse I think, since we would need to handle non-numeric inputs and things too.

andrewazores (approved):

PR #496, opened 2022-08-05
fix(targets): select TargetSelect to none when TargetJvmNotification LOST if that target is currently selected

### LINKED ISSUE #294: [Task] CreateRecording form does not recognize when targets reappear
If a target becomes briefly unreachable while filling out
the “Create recording” form, clicking the "Create" button will trigger
an expected HTTP 500 error. When the target becomes available again,
the Create button sometimes stays greyed out and the template dropdown
remains empty. Clicking the “refresh targets” button doesn’t update
the template dropdown. If you exit and re-enter the Create Recording
form then you can successfully create a recording.

Steps to reproduce:
1) Install cryostat operator in an Openshift cluster
2) Install a sample app to monitor:
```
oc new-app --docker-image=quay.io/andrewazores/quarkus-test:0.0.2
oc patch svc/quarkus-test -p '{"spec":{"$setElementOrder/ports":[{"port":9096},{"port":9999}],"ports":[{"name":"jfr-jmx","port":9096}]}}'
```
3) Select your target quarkus-test and click Create Recording
4) While viewing the form, restart the pod:
```
oc scale --replicas=0 deployment/quarkus-test
oc scale --replicas=1 deployment/quarkus-test
```
5) Notice the 'target appeared' notification was received. The templates dropdown should still be empty. Alternatively, if you filled out the entire Create Recording form before restarting the pod, the "Create" button might still be blue. Clicking on the blue button doesn't do anything.

maxcao13:
Related to https://github.com/cryostatio/cryostat-web/issues/294#issuecomment-1204281360
Depends on https://github.com/cryostatio/cryostat/pull/1037
Fixes #294

The fix stops the `TargetSelect` from selecting a target if it was selected when a `TargetJvmDiscovery` LOST notification is emitted. So it is entirely dependent on LOST notifications getting through somewhat quickly.

### DIFF
--- a/src/app/TargetSelect/TargetSelect.tsx
+++ b/src/app/TargetSelect/TargetSelect.tsx
@@ -148,7 +148,7 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props) 
       }
     }
     setExpanded(false);
-  }, [context, context.target, selected, notifications, setExpanded, setCachedTargetSelection, removeCachedTargetSelection]);
+  }, [context, context.target, selected, notifications, setExpanded]);
 
   const selectNone = React.useCallback(() => {
     onSelect(undefined, undefined, true);
--- a/src/app/TargetSelect/TargetSelect.tsx
+++ b/src/app/TargetSelect/TargetSelect.tsx
@@ -47,6 +47,7 @@ import { catchError, first } from 'rxjs/operators';
 
 import { CreateTargetModal } from './CreateTargetModal';
 import _ from 'lodash';
+import { NotificationCategory } from '@app/Shared/Services/NotificationChannel.service';
 
 export interface TargetSelectProps {
 }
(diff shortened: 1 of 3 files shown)


src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -63,13 +64,34 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props)
...
+    addSubscription(
+      context.notificationChannel.messages(NotificationCategory.TargetJvmDiscovery)
+        .subscribe(v => {
+          const evt: TargetDiscoveryEvent = v.message.event;
+          if (evt.kind === 'LOST') {
+            const target: Target = {
+              connectUrl: evt.serviceRef.connectUrl,
+              alias: evt.serviceRef.alias,
+            }
+            context.target.target().subscribe((currentTarget) => {
```
andrewazores, line 85: This subscription is left open. It should either be added to a nested `addSubscription` call, or you can use something like `.pipe(first())` so that it automatically closes after the first value emission.

andrewazores (approved):

PR #499, opened 2022-08-18
fix(archives): use target-specific archive API endpoints

### LINKED ISSUE #445: [Task] Deleting archived recordings can delete wrong file for identically-named but distinct files
This problem of `/api/v/recordings/:recordingName/etc` endpoints being unable to differentiate between per-target archives and the uploaded archives is actually affecting more than just labels, too. Deleting an archived recording also runs into problems since the file is searched for by name, but can be found in two locations.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-web/issues/442#issuecomment-1143942713_

hareetd:
Fixes #445 
Related https://github.com/cryostatio/cryostat/pull/1047

### DIFF
--- a/src/app/Archives/Archives.tsx
+++ b/src/app/Archives/Archives.tsx
@@ -44,6 +44,7 @@ import { BreadcrumbPage } from '@app/BreadcrumbPage/BreadcrumbPage';
 import { ArchivedRecordingsTable } from '@app/Recordings/ArchivedRecordingsTable';
 import { Target } from '@app/Shared/Services/Target.service';
 import { of } from 'rxjs';
+import { UPLOADS_SUBDIRECTORY } from '@app/Shared/Services/Api.service';
 
 export const Archives = () => {
   const context = React.useContext(ServiceContext);
@@ -63,7 +64,7 @@ export const Archives = () => {
   notification handling in the ArchivedRecordingsTable. 
   */ 
   const target: Target = { 
-    connectUrl: '',
+    connectUrl: UPLOADS_SUBDIRECTORY,
     alias: '',
   }
 
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -338,7 +338,7 @@ export class ApiService {
     ));
   }
 
-  uploadArchivedRecordingToGrafana(sourceTarget: Subject<Target>, recordingName: string): Observable<boolean> {
+  uploadArchivedRecordingToGrafana(sourceTarget: Observable<Target>, recordingName: string): Observable<boolean> {
     return sourceTarget.pipe(concatMap(target => 
       this.sendRequest(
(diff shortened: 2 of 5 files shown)


src/app/Archives/Archives.tsx
```diff
@@ -63,7 +64,7 @@ export const Archives = () => {
   notification handling in the ArchivedRecordingsTable. 
   */ 
   const target: Target = { 
-    connectUrl: '',
+    connectUrl: UPLOADS_SUBDIRECTORY,
```
tthvo, line 67: I am just a bit concerned that within `ArchivedRecordingsTable`, we call to `context.target` for selected target at some hooks:

https://github.com/cryostatio/cryostat-web/blob/bb8b4db69ada370dd654ca2ddc54bdc44d8448ee/src/app/Recordings/ArchivedRecordingsTable.tsx#L247

but use `props.target` at others:

https://github.com/cryostatio/cryostat-web/blob/bb8b4db69ada370dd654ca2ddc54bdc44d8448ee/src/app/Recordings/ArchivedRecordingsTable.tsx#L228

Also, this could complicate more as now recording table saves a targetConnectURl as a local state.

tthvo: I think the `props.target` passed to ArchivedRecordingTable mostly for UploadsTable as it does not have target. I think we could either:

1. Create a separate component for UploadsTable
2. Rename props and add checks if we should use `context.target` or fixed `props.target`).

src/app/Shared/Services/Api.service.tsx
```diff
@@ -808,3 +820,5 @@ export interface MatchedCredential {
   matchExpression: string;
   targets: Target[];
 }
+
+export const UPLOADS_SUBDIRECTORY: string = 'uploads';
```
tthvo, line 824: > The WebServer now points to the beta RecordingGetHandler and ReportGetHandler endpoints when the WebServer#getArchivedDownloadUrl and WebServer#getArchivedReportUrl methods are called, respectively. This means archived recording-related notifications will now contain target-specific download URLs. Notifications have also been updated so that the target field is equal to "uploads" rather than the empty string for uploaded recordings. This is because for the new target-specific handlers the sourceTarget parameter cannot be set to the empty string in place of a connectUrl for uploaded recordings. Since the "uploads" string is the logical alternative to use for this parameter, in order to maintain consistency, notifications also explicitly identify a recording as having originated as an upload directly into Cryostat's archives.

I am thinking we could summarize Hareet's comment above and add it here for reference.

src/app/Archives/Archives.tsx
maxcao13: I think this is problem is a bit big to be part of this issue, I see what you mean about wanting the UploadsTable to be another component as the layout right now is a bit messy.

tthvo (approved):
Looks good and worked well :D I think #522 can be done as follow-up later to avoid confusion and potential bugs.

andrewazores (approved):

Issue #468, opened 2022-06-17
[Epic] Target Dashboard


andrewazores:
Currently the main/default "Dashboard" view is completely empty, other than a `TargetSelect` widget. This should be populated with some information about the target itself (JMX URL, alias, labels, annotations), as well as some widgets showing some high-level information about some select metrics. This could include some charts of timeseries-like JFR event data, or automated analysis reports scores. Ideally, the user can configure the layout and content of these widgets within the Dashboard, and this layout is reused for all targets and the layout information persisted for the user's next session.

Issue #429, opened 2022-05-06
[Task] Reimplement TargetSelect dropdown as Patternfly Context Selector


andrewazores:
https://www.patternfly.org/v4/components/context-selector

PR #501, opened 2022-08-23
feat(targets): target select component can be expanded to reveal details

andrewazores:
Related to #429 
Related to #468


src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -251,11 +263,24 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props)
 
   return (
     <>
-      <Card>
-        <CardHeader>
-          <CardHeaderMain>
+      <Card isExpanded={props.simple || selected === NO_TARGET ? undefined : isExpanded}>
```
tthvo, line 266: looks like `props.simple || selected === NO_TARGET` being used a lot. Maybe, we should extract into a React memo?

src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -251,11 +263,24 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props)
...
+        <CardHeader
+          onExpand={props.simple || selected === NO_TARGET ? undefined : onExpand}
```
tthvo, line 268: Since these props are `undefined` if `props.simple || selected === NO_TARGET`, maybe we could just defined a memo object that contains these props and spread its content out if not `props.simple || selected === NO_TARGET`?

src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -251,11 +263,24 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props)
...
+          isToggleRightAligned={props.simple || selected === NO_TARGET ? undefined : true}
+          toggleButtonProps={
+            props.simple || selected === NO_TARGET
+              ? undefined
+              : {
+                  id: 'toggle-button1',
```
tthvo, line 274: Maybe a more specific id and aria-label for this `targetSelect`? Just in case another component uses the same.

src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -251,11 +263,24 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props)
...
+                  'aria-label': 'Details',
+                  'aria-labelledby': 'expandable-card-title toggle-button1',
+                  'aria-expanded': isExpanded,
+                }
+          }
+        >
+          <CardTitle>
             <Text component={TextVariants.h4}>Target JVM</Text>
```
tthvo, line 282: Not sure why `h4` style is not being applied here. But using this fixes the issue:

https://www.patternfly.org/v4/components/card/#with-html-heading-element

```typescript
<CardTitle component="h4">
     Target JVM
</CardTitle>
```

src/app/Dashboard/Dashboard.tsx
```diff
@@ -41,7 +41,7 @@ import { TargetView } from '@app/TargetView/TargetView';
 export interface DashboardProps {}
 
 export const Dashboard: React.FunctionComponent<DashboardProps> = (props) => {
-  return <TargetView pageTitle="Dashboard" compactSelect={true} />;
+  return <TargetView pageTitle="Dashboard" compactSelect={false} />;
```
tthvo, line 44: Some reasons to expand the targetSelect out here :D ??

src/test/TargetSelect/TargetSelect.test.tsx
```diff
@@ -55,14 +55,15 @@ const cryostatAnnotation = {
...
+  jvmId: 'abcd',
   connectUrl: mockFooConnectUrl,
   alias: 'fooTarget',
   annotations: {
     cryostat: cryostatAnnotation,
     platform: {},
   },
 };
-const mockBarTarget: Target = { ...mockFooTarget, connectUrl: mockBarConnectUrl, alias: 'barTarget' };
+const mockBarTarget: Target = { ...mockFooTarget, jvmId: 'efgh', connectUrl: mockBarConnectUrl, alias: 'barTarget' };
```
tthvo, line 66: Maybe a test to check if the expanded codeblock is actually shown?

maxcao13 (commented):
Looks really good, just a general question for the future and not content for this PR. 

When we start to populate the Dashboard with various things listed in the Epic here https://github.com/cryostatio/cryostat/issues/945, does it make sense that `SerializedTargets` will start to include some data around things like CPU, memory, network, most recent archived recordings, Automated Analysis reports, etc. ? Or maybe have another wrapper around the `SerializedTarget` which includes that extra info, and keeps the Targets modular as they are?

src/test/TargetSelect/TargetSelect.test.tsx
andrewazores: Added

src/app/Dashboard/Dashboard.tsx
andrewazores: It was a bit redundant, but in the last set of commits I've removed the empty state from the dashboard itself and put an empty state into the `TargetSelect`. The expansion button is now always visible (unless in `simple` style) and the empty state is shown if nothing is selected. This cleans up the code a bit too, there's no more `props.simple || selected === NO_TARGET ? undefined : whatever` everywhere.

andrewazores:
> Looks really good, just a general question for the future and not content for this PR.
> 
> When we start to populate the Dashboard with various things listed in the Epic here [cryostatio/cryostat#945](https://github.com/cryostatio/cryostat/issues/945), does it make sense that `SerializedTargets` will start to include some data around things like CPU, memory, network, most recent archived recordings, Automated Analysis reports, etc. ? Or maybe have another wrapper around the `SerializedTarget` which includes that extra info, and keeps the Targets modular as they are?

I don't think that info would go inside the `SerializedTarget` component. Perhaps there could be a modular/dynamic card on the Dashboard that also wraps a `SerializedTarget` so the user can have that more prominently and permanently displayed, but those other bits of information like summaries of recordings on the target or metrics pull from cryostat-reports/jfr-datasource would go on their own dynamic dashboard cards, not within the SerializedTarget.

src/app/TargetSelect/TargetSelect.tsx
andrewazores: I'm not sure why any of our card titles were done the way they were at this point, either. `CardTitle` was just not being used, everything was `<CardHeaderMain><Text>foo</Text></CardHeaderMain>`. I've replaced those all with `CardTitle` so it's simpler and more idiomatic. This does mean that the card titles everywhere are now bolder, but that's just Patternfly default styling that we're now conformant to, so it's probably a good thing.

tthvo: Nice thanks :D

src/app/TargetSelect/TargetSelect.tsx
tthvo: Might also need to change `'aria-labelledby': 'expandable-card-title toggle-button1',` with the new id too?

src/app/Dashboard/Dashboard.tsx
tthvo: Nicee very much cleaner :D

src/app/TargetSelect/TargetSelect.tsx
andrewazores: Oh right, I had that before and then accidentally dropped the change, then partially re-added it... d'oh. Thanks.

src/app/TargetSelect/TargetSelect.tsx
tthvo: On this note, how about this? Probably easier to look at all the needed props.

```typescript
const cardHeaderProps = React.useMemo(() => props.simple? {}: (
    {
      onExpand: onExpand,
      isToggleRightAligned: true,
      toggleButtonProps: {
        id: 'target-select-expand-button',
        'aria-label': 'Details',
        'aria-labelledby': 'expandable-card-title target-select-expand-button',
        'aria-expanded': isExpanded,
      }
    }
  ), [onExpand, props.simple, isExpanded]);

  return (
    <>
      <Card isExpanded={props.simple ? undefined : isExpanded}>
        <CardHeader
          {...cardHeaderProps}
        >
[...to be continued]
```

What do you think?

src/test/TargetSelect/TargetSelect.test.tsx
tthvo: Actually, just a tiny suggestion here: maybe we should also check things visually alongside with tags so its closer to what users see.

src/test/TargetSelect/TargetSelect.test.tsx
```diff
@@ -138,6 +145,44 @@ describe('<TargetSelect />', () => {
...
+
+    expect(screen.getByText('Select target...')).toBeInTheDocument();
+
+    const expandButton = screen.getByLabelText('Details');
+    userEvent.click(expandButton);
+
+    const articleElement = container.querySelector('article');
+    expect(articleElement).toBeInTheDocument();
+    expect(articleElement).toBeVisible();
+    expect(screen.getByText(`No target selected`)).toBeInTheDocument();
```
tthvo, line 163: Probably also check if they are visible too?

tthvo (approved):
Looks great to me! Very nicee :D

PR #503, opened 2022-09-02
feat(recordings): more tools for filtering by labels

### LINKED ISSUE #79: [react] State management
As detailed in some recent PR comments:

https://github.com/rh-jmc-team/container-jfr-web/pull/73#issuecomment-629559576

We need to investigate client-side state management to simplify how different views can share the same application state, rather than needing to reinitialize their own local view of the state when mounted, and also to help prevent spurious data refreshes when things update that are not really local client state (ex. button clicks to expand/collapse menus etc.) but rather simple view state.

### LINKED ISSUE #440: [Story] Recordings search/sort/select enhancements
Tasks:
- [x] #414 
- [x] #441
- [x] #488

### LINKED ISSUE #441: [Task] Labels in recording tables are clickable and select similarly-labelled recordings
Parent: #440 

The labels displayed on recording table rows should be clickable. When I click a label I expect the current row to be selected (checkbox ticked), and all other rows in the same table with the same label also selected. I should then be able to perform bulk actions across this selection set such as stopping, label-editing, or archiving all of them at once.

How this is implemented should be considered alongside the sibling task #414. #414 should allow some basic meanings like `label/A = foo AND label/B = BAR AND label/C = BAZ` to be specified by the user. Clicking on labels in the recording rows could add them to the set of `label/A = FOO, label/B = BAR, label/C = BAZ`, with some toolbar menu option that allows the user to choose the `AND` or `OR` operation. The #414 UI should also allow the user to search or manually enter a label value textually, not only graphically by clicking on the label elements as specified by this issue.

### LINKED ISSUE #488: [Task] Recording search/filter unit tests
Unit tests for the search/filters component in the Active and Archived recording tables are missing unit tests.

The implementation for this feature is in #414

As of #414, the `<RecordingFilters/>` component in the recordings table toolbar is mocked out with the following:
```
jest.mock('@app/Recordings/RecordingFilters', () => {
  return {
    ...jest.requireActual('@app/Recordings/RecordingFilters'),
    RecordingFilters: jest.fn(() => {
      return <div>
                RecordingFilters
             </div>
    })
  };
});
```
Since the filter component will only generate useful output if it's part of a recordings table, you could remove this mock and test the recording filter component by performing various `userEvent.click`s in the `ActiveRecordingsTable.test` and `ArchivedRecordingsTable.test`. If you can find a way to test the component in it's own unit test file, that's good too.

tthvo:
Fixes #79
Fixes #488
Fixes #441 
Related to #486 
Fixes #440 (closing parent)

**Improvements:** 
- [x] Refactor filters to their own source files.
- [x] Users should be able to delete all filters in the same group/category (More convenient?).
- [x] Labels should be clickable for users to select matching recordings. Clicking on selected labels will deselect them.
- [x] Filtered labels should be highlighted (i.e. blue).
- [x] Filter states (i.e. selected category and filters) should persist between:
    - [x] Route changes (i.e. Redux)
    - [x] Page reloads (local state unless disabled).
- [x] Clickable labels should be styled for hover action.
- [x] Recording state filter should allow toggle.
- [x] Timepicker needs to render correctly (i.e. spacings are misaligned). Appending the popever menu to component higher in the tree solved the issue.
- [x] Date and time selection must sync (currently changing date does not reflect currently selected time).
- [x] Row checkboxes should update with recording filters (i.e. find alternative to just list index).
- [x] Labels in filter should be sorted for easier selection (essentially grouped together based on key).
- [x] Selected names/labels should not appear in selection options.
- [x] Optimize subscriptions to message channels (i.e duplicate parent and child sub causes unnecessary rebuilds)
- [x] Toolbar rebuilding should not always cause DOM updates (i.e. extract as a component instead of inner func).
- [x] Optimize `context.settings` calls with React.useMemo. 

**Test:**
- [x] Added unit tests for recording search/filters.
  - [x] Separate tests for each filter.
  - [x] RecordingFilters as a whole (no snapshots, see [comment](https://github.com/cryostatio/cryostat-web/pull/503#issuecomment-1247290505))
  - [x] LabelCells
  - [x] BulkEditLabels
  - [x] Active/Archived Table updates (no snapshots, see [comment](https://github.com/cryostatio/cryostat-web/pull/503#issuecomment-1247290505))
- [x] Use historyMemory instead of mock.
- [x] Added missing clean up calls after each tests.
- [x] Added wrapper for React Testing Library `render` to render with Context, Route, redux.Provider.


tthvo:
Any thoughts? I will add unit tests based on these new behaviours :D

For the `Category of the currently added filter should remain.`, there is a case where the user selects the category but does not add the filter yet and a new notification comes, the dropdown menu will not reflect the currently selected category. A quick fix to trigger parent rebuilds every time the category is chosen but I am thinking it becomes too many unnecessary rebuilds.

andrewazores:
I haven't looked at the implementation but just by testing it out by hand, I'm really excited to see this. It looks and feels fantastic.

> For the Category of the currently added filter should remain., there is a case where the user selects the category but does not add the filter yet and a new notification comes, the dropdown menu will not reflect the currently selected category. A quick fix to trigger parent rebuilds every time the category is chosen but I am thinking it becomes too many unnecessary rebuilds.

To check my understanding, say I am using the `Name` filter. I select that and click the dropdown, which displays a selection of recording names. If a new recording appears now, perhaps due to an Automated Rule, then the dropdown contents are not updated, so I have unsynced information between the dropdown list and the table I can see behind it. Is that what you mean?

andrewazores:
> > To check my understanding, say I am using the Name filter. I select that and click the dropdown, which displays a selection of recording names. If a new recording appears now, perhaps due to an Automated Rule, then the dropdown contents are not updated, so I have unsynced information between the dropdown list and the table I can see behind it. Is that what you mean?
> 
> Oh sorry actually the issue is, as in the situation above, instead we use `Label` filter (any other than `Name`) for example, then when a recording appears now, the dropdown actually changes back to `Name` abruptly.

Ah, I see. I suppose that's because of how the component relationships and lifecycles are set up - updating the content of the table causes the toolbar to also be re-rendered and we lose that selection state? That sounds like it matches what you said with `A quick fix to trigger parent rebuilds every time the category is chosen but I am thinking it becomes too many unnecessary rebuilds.`

tthvo:
Yess exactly the parent table now holds the state for currently selected category of the filters. I will try to find a better way to resolve this as rebuilding with large number of recordings for only a change in category dropdown is not ideal.

tthvo:
I played around with Redux and finally got some new refinements to our recording filters. Since last time, I added:

- Recording filters for each target are persistent across route changes and page reload.
- Filter for recording state now allows toggling as it uses checkboxes.
- Datetime filter layout is fixed (no longer misaligned and cut off due to overlapping).
- A nicer hover styling for clickable labels.

Tests are still failing everywhere with addition of Redux. I will fix them after. Any thoughts @andrewazores @maxcao13?

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -64,19 +67,14 @@ export interface ActiveRecordingsTableProps {
...
-    StartedAfterDate?: string[],
-    DurationSeconds?: string[],
-}
-
 export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTableProps> = (props) => {
   const context = React.useContext(ServiceContext);
   const routerHistory = useHistory();
+  const { url } = useRouteMatch();
+  const addSubscription = useSubscriptions();
+  const disPatch = useDispatch();
```
andrewazores, line 75: rename to `dispatch`

andrewazores:
This looks great. I've looked into Redux before but never truly used it, so my review of some parts of the implementation is very superficial, but things look good. Testing it out manually this also looks and feels excellent.

tthvo:
Great :D I will fix and add tests for unit tests and mark it ready when done.

maxcao13 (commented):
https://user-images.githubusercontent.com/69827283/189714290-b0ae9f46-73f7-45a3-9358-da462c89f442.mp4

Small bug with the way that the checkboxes are using state with the rowIndexes.

tthvo:
> It may be redundant to have a "Clear all filters" button as well as the `x` which does it as well. Not sure.

`Clear all filters` will clear all filters while the `x` in chip group just clear filters in a specific group :D I think we need both for flexibility.

tthvo:
> 2022-09-12.13-03-27.mp4
> 
> Small bug with the way that the checkboxes are using state with the rowIndexes.

Ahh right! I think the selectAll checkbox needs updating along with filters. I will look into and fix it.

maxcao13:
The bug is fixed and works well!

Just a couple of small things:
1. Maybe it might be better for the user if the dropdown for Label filters greys out any filters that are already applied?
2. This should probably be a future feature in a future PR, but something that would be cool is "strict" filters? For example, the recording MUST have these labels/template otherwise it does not show. Right now I think the filters are just a union function, implementing an intersection would be cool as well!

andrewazores:
The original #441 issue does describe the union/intersection filter application :-) (as `OR` and `AND`)

tthvo:
> Maybe it might be better for the user if the dropdown for Label filters greys out/disables any filters that are already applied? 

I think we could remove any "already selected" options for Name and Label Filters then.

> This should probably be a future feature in a future PR, but something that would be cool is "strict" filters? For example, the recording MUST have these labels/template otherwise it does not show. Right now I think the filters are just a union function, implementing an intersection would be cool as well!

That would be super nice :D Probably a little tweak from Patternfly guideline tho as said here:

https://www.patternfly.org/v4/guidelines/filters/#filter-chip
> Filters between attribute categories should be combined with a boolean “AND” operator. Filters within a category are shown grouped together and should be combined with a boolean “OR” operator.

tthvo:
> The original https://github.com/cryostatio/cryostat-web/issues/441 issue does describe the union/intersection filter application :-) (as OR and AND)

I guess we follow the guideline for now? https://www.patternfly.org/v4/guidelines/filters/#filter-chip

andrewazores:
Sure, that sounds good.

maxcao13:
Works well! I don't see any other problems, so I'll approve when tests are fixed. Good work!

tthvo:
I am trying to work around this issue with umounted components:
```bash
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    at Select (/home/thvo/workspace/cryostat-web/node_modules/@patternfly/react-core/dist/js/components/Select/Select.js:33:5)
    at NameFilter (/home/thvo/workspace/cryostat-web/src/app/Recordings/Filters/NameFilter.tsx:50:37)
```

I think the issue is:

- Patternfly `Select` component when calling `onToggle` also calls `setState` for its own component: https://github.com/patternfly/patternfly-react/blob/a61551a4c549a595f39839ce3ce8cdabd70717dd/packages/react-core/src/components/Select/Select.tsx#L371
- That passed toggle callback calls `setIsOpen` to update our filters `NameFilter` and `LabelFilter`: https://github.com/cryostatio/cryostat-web/blob/b355296753914ce8799748dc63f2513c659a2af1/src/app/Recordings/NameFilter.tsx#L79
- This causes updates on unmounted child `Select`[0].
- Also, the onToggle is called twice, which again causes updates on unmounted parent `NameFilter` or `LabelFilter` (unmounted due to previous call) [1].

For [1], I did add a ref to hold `mounted` state and avoid updates if unmounted. However, [0] is internal to patternfly. Any guide on this @andrewazores?

tthvo:
Interestingly, the issue occurs  only at the second time I select an option, not any subsequent selections.

andrewazores:
Sorry, you've already caught up to my knowledge of React/Patternfly, so I don't have much insight to offer here :-)

Would it help if the components communicated by passing each other ex. `Subject`/`Observable` that could be used to pub/sub events, rather than passing direct callback functions? This is a technique I used a few years back on a different project using Angular but IIRC there were similar reasons for it.

tthvo:
> Sorry, you've already caught up to my knowledge of React/Patternfly, so I don't have much insight to offer here :-)
> 
> Would it help if the components communicated by passing each other ex. `Subject`/`Observable` that could be used to pub/sub events, rather than passing direct callback functions? This is a technique I used a few years back on a different project using Angular but IIRC there were similar reasons for it.

Oh right! Thanks for the tips! I will read and give it a try. Since its just a warning, I guess I will finish up the tests first.

tthvo:
Unfortunately, for the datetime selection menu not to be cut off, we need to rely on DOM to find parent `Drawer`. This means we cannot run snapshot tests. Still, even without depending on DOM, we still cannot run snapshot tests with patternfly `DatePicker` and `TimePicker` as they relies on React ref.

andrewazores:
That's unfortunate. I don't know much about this, either. You've probably already seen some of these resources, but it may be useful to look at:

https://stackoverflow.com/questions/40852131/refs-are-null-in-jest-snapshot-tests-with-react-test-renderer

If there is no way around then that's fine, standard unit tests are enough.

tthvo:
Right! Thanks for pointing me to that :D

Just unfortunately, I just tried and it seemed Patternfly component also calls funcs on those refs, which make it hard to know which func to stub and crashes if not.

andrewazores:
> Right! Thanks for pointing me to that :D
> 
> Just unfortunately, I just tried and it seemed Patternfly component also calls funcs on those refs, which make it hard to know which func to stub and crashes if not.

Okay, I wouldn't worry too much about it right now then. Something to keep any eye out for to improve code quality in the future but out of the scope of this PR.

tthvo:
Ready for reviews!

I think the issue with `Warning: Can't perform a React state update on an unmounted component` needs a bit of investigating since it happens for components that use `Select` Patternfly component.

src/app/Recordings/ArchivedRecordingsTable.tsx
```diff
@@ -457,3 +505,12 @@ export const ArchivedRecordingsTable: React.FunctionComponent<ArchivedRecordings
     </Drawer>
   );
 };
+
+export const convertToNumber = (recordingName: string): number => {
```
tthvo, line 509: Recording rows require a numeric index. 

I used `id` for ActiveRecording. However, ArchivedRecording does not so I used its unique and calculate its hash. Kinda a hack but it should do a good job avoiding collision. Any better suggestions?

tthvo:
> I think the issue with Warning: Can't perform a React state update on an unmounted component needs a bit of investigating since it happens for components that use Select Patternfly component

I suspect a setState call Select's `onToggle` causes this issue. 

I could not figure out how to fix it with observable/sub. But a quick hack is to `useRef` keeping a flag to indicate whether the component is mounted. Then, `useEffect` to update it to true and clean-up to update it to false. Any callback passed to `onToggle` should wrap in this check and ignore if unmounted.

tthvo:
> > I think the issue with Warning: Can't perform a React state update on an unmounted component needs a bit of investigating since it happens for components that use Select Patternfly component
> 
> I suspect a setState call Select's `onToggle` causes this issue as `Select`'s source. Though interestingly, this happens only once at first-second renders.
> 
> ```js
> onToggle = (isExpanded: boolean, e: React.MouseEvent | React.ChangeEvent | React.KeyboardEvent | Event) => {
>     const { isInputValuePersisted, onSelect, onToggle, hasInlineFilter } = this.props;
>     if (!isExpanded && isInputValuePersisted && onSelect) {
>       onSelect(undefined, this.inputRef.current ? this.inputRef.current.value : '');
>     }
>     if (isExpanded && hasInlineFilter) {
>       this.setState({         //  Calling setState itself here but the parent is already unmounted
>         focusFirstOption: true
>       });
>     }
>     onToggle(isExpanded, e);
>   };
> ```
> 
> I could not figure out how to fix it with observable/sub. But a quick hack is to `useRef` keeping a flag to indicate whether the component is mounted. Then, `useEffect` to update it to true and clean-up to update it to false. Any callback passed to `onToggle` should wrap in this check and ignore if unmounted.

After some investigating, I think I found the root issue: 

Our `RecordingToolbar` is defined a inner function, which is computed on every render => essentially creates a new type of component. That's why the toolbar is always updating the DOM and all its children rebuild from fresh (lost states). I think because of this behaviour, UI events are not properly handled (perhaps callbacks to unmmounted components are cached and called). 

To fix this, I move the `RecordingToolbar` outside as a separate component and the issue `Warning: Can't perform a React state update on an unmounted component ` no longer occurs. 

However, tests still throw errors. Interestingly, I rearranged the test orders and was able to fix for `DatetimePicker.test.tsx`. The same might not be true for other filters. I think it's related to issue of `React Testing Library` or my test setup, which is very odd.

src/app/Recordings/ArchivedRecordingsTable.tsx
andrewazores: Makes sense to me. The archived recording name is unique in the context of an individual target, so hashing the name seems like a good solution.

andrewazores:
Speaking of toolbars, I just noticed something about the `Duration` filter in the toolbar.

The two "started before/after" filters have a magnifying glass search button to perform the filter after selecting the date/time parameters. The other filters have a dropdown menu (`Select`) and selecting an item immediately starts the filtering with that item. All good so far.

For the Duration filter, clicking the `Continuous` checkbox starts filtering for recordings with a `continuous` duration. Unchecking the box removes the filter. There doesn't seem to be a way to filter for recordings that are *not* continuous, other than by using the numeric duration filter.

The numeric duration filter only allows for matching exact recording durations, not greater-than or less-than, which I think is okay (for now), but to actually add this filter the user has to click the numeric input field and then press Enter.

Here's my idea for how to make this filter control a little more intuitive, at least to me:

1. Add a magnifying glass button like the "started before/after" filter which actually adds a filter based on the selected criteria
2. Remove the static "Continuous" checkbox
3. Add a `Select` with values `Continuous`, `Greater Than`, `Less Than`
4. Add a secondary control that varies according to the value selected above:
4a. If `Continuous`, dynamically display a "Continuous" checkbox. The state of this checkbox becomes a "continuous == state" filter when the magnifying glass is clicked. This way the user can filter by only continuous recordings, or only fixed durations
4b. If `Greater Than` or `Less Than`, dynamically display a duration picker. I think this should match the duration field on the recording creation form, with a number input and a separate units input. When the magnifying glass is clicked then apply a filter that looks for recordings fitting the specific criteria, ex. "recordings with fixed duration less than 5 minutes".

This is best suited for a separate PR, in any case.

src/app/Recordings/ArchivedRecordingsTable.tsx
andrewazores: Could you extract this into somewhere within `src/app/utils` and rename it something like `hashcode`?

src/app/RecordingMetadata/LabelCell.tsx
```diff
@@ -36,21 +36,56 @@
...
 
+import { getLabelDisplay } from '@app/Recordings/Filters/LabelFilter';
+import { UpdateFilterOptions } from '@app/Shared/Redux/RecordingFilterReducer';
 import { Label, Text } from '@patternfly/react-core';
 import React from 'react';
+import { ClickableLabel } from './ClickableLabel';
 import { RecordingLabel } from './RecordingLabel';
 
 export interface LabelCellProps {
+  target: string,
```
andrewazores, line 47: s/,/;/

src/app/RecordingMetadata/ClickableLabel.tsx
```diff
@@ -0,0 +1,88 @@
...
+    () => props.onLabelClick(props.label), 
+    [props.label, props.onLabelClick, getLabelDisplay]
+  );
+
+  return <>
+    <Label
+      aria-label={`${props.label.key}: ${props.label.value}`}
+      style={style}
+      onMouseEnter={() => handleHoveredOrFocused()}
+      onMouseLeave={handleNonHoveredOrFocused}
```
andrewazores, line 79: Is there a particular reason that this one is passed as a reference to a callback and the line above passes a lambda that calls the callback?

src/app/RecordingMetadata/LabelCell.tsx
```diff
@@ -36,21 +36,56 @@
...
   labels: RecordingLabel[];
+  // Must be specified along with updateFilters.
```
andrewazores, line 49: Are these two props expected to always both be undefined or both be defined? If so, would it make sense to wrap them both into an optional config object, where the config object is optional but both of these fields within are required? This way the type system helps ensure the programmer doesn't make a mistake and only define one of the two.

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -64,38 +69,28 @@ export interface ActiveRecordingsTableProps {
...
-}
-
 export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTableProps> = (props) => {
   const context = React.useContext(ServiceContext);
   const routerHistory = useHistory();
+  const { url } = useRouteMatch();
+  const addSubscription = useSubscriptions();
+  const dispatch = useDispatch<StateDispatch>();
 
+  const [target, setTarget] = React.useState(""); // connectURL of the target
```
andrewazores, line 79: How about removing the comment and just renaming the variable to suit? :-)

src/app/Shared/Redux/RecordingFilterReducer.tsx
```diff
@@ -0,0 +1,253 @@
...
+  return {
+    ...targetRecordingFilter, 
+    active: {
+      selectedCategory: targetRecordingFilter.active.selectedCategory,
+      filters: emptyActiveRecordingFilters,
+    }
+  }
+}
+
+// Initial states are loaded from local storage if there are any (TODO)
```
andrewazores, line 129: Is this still a TODO?

src/app/RecordingMetadata/ClickableLabel.tsx
tthvo: Ahh no actually. I think I forgot to remove it. Fixed now.

src/app/Shared/Redux/RecordingFilterReducer.tsx
tthvo: Right I will remove this!

src/app/RecordingMetadata/LabelCell.tsx
tthvo: Right make sense!

tthvo:
> Speaking of toolbars, I just noticed something about the `Duration` filter in the toolbar.
> 
> The two "started before/after" filters have a magnifying glass search button to perform the filter after selecting the date/time parameters. The other filters have a dropdown menu (`Select`) and selecting an item immediately starts the filtering with that item. All good so far.
> 
> For the Duration filter, clicking the `Continuous` checkbox starts filtering for recordings with a `continuous` duration. Unchecking the box removes the filter. There doesn't seem to be a way to filter for recordings that are _not_ continuous, other than by using the numeric duration filter.
> 
> The numeric duration filter only allows for matching exact recording durations, not greater-than or less-than, which I think is okay (for now), but to actually add this filter the user has to click the numeric input field and then press Enter.
> 
> Here's my idea for how to make this filter control a little more intuitive, at least to me:
> 
>     1. Add a magnifying glass button like the "started before/after" filter which actually adds a filter based on the selected criteria
> 
>     2. Remove the static "Continuous" checkbox
> 
>     3. Add a `Select` with values `Continuous`, `Greater Than`, `Less Than`
> 
>     4. Add a secondary control that varies according to the value selected above:
>        4a. If `Continuous`, dynamically display a "Continuous" checkbox. The state of this checkbox becomes a "continuous == state" filter when the magnifying glass is clicked. This way the user can filter by only continuous recordings, or only fixed durations
>        4b. If `Greater Than` or `Less Than`, dynamically display a duration picker. I think this should match the duration field on the recording creation form, with a number input and a separate units input. When the magnifying glass is clicked then apply a filter that looks for recordings fitting the specific criteria, ex. "recordings with fixed duration less than 5 minutes".
> 
> 
> This is best suited for a separate PR, in any case.

Great idea! I think I could look into this for this PR too as it seems within scope :D

andrewazores:
> > Speaking of toolbars, I just noticed something about the `Duration` filter in the toolbar.
> > The two "started before/after" filters have a magnifying glass search button to perform the filter after selecting the date/time parameters. The other filters have a dropdown menu (`Select`) and selecting an item immediately starts the filtering with that item. All good so far.
> > For the Duration filter, clicking the `Continuous` checkbox starts filtering for recordings with a `continuous` duration. Unchecking the box removes the filter. There doesn't seem to be a way to filter for recordings that are _not_ continuous, other than by using the numeric duration filter.
> > The numeric duration filter only allows for matching exact recording durations, not greater-than or less-than, which I think is okay (for now), but to actually add this filter the user has to click the numeric input field and then press Enter.
> > Here's my idea for how to make this filter control a little more intuitive, at least to me:
> > ```
> > 1. Add a magnifying glass button like the "started before/after" filter which actually adds a filter based on the selected criteria
> > 
> > 2. Remove the static "Continuous" checkbox
> > 
> > 3. Add a `Select` with values `Continuous`, `Greater Than`, `Less Than`
> > 
> > 4. Add a secondary control that varies according to the value selected above:
> >    4a. If `Continuous`, dynamically display a "Continuous" checkbox. The state of this checkbox becomes a "continuous == state" filter when the magnifying glass is clicked. This way the user can filter by only continuous recordings, or only fixed durations
> >    4b. If `Greater Than` or `Less Than`, dynamically display a duration picker. I think this should match the duration field on the recording creation form, with a number input and a separate units input. When the magnifying glass is clicked then apply a filter that looks for recordings fitting the specific criteria, ex. "recordings with fixed duration less than 5 minutes".
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is best suited for a separate PR, in any case.
> 
> Great idea! I think I could look into this for this PR too as it seems within scope :D

I'd rather it as a separate PR, just so it's a little easier to review. This one's already pretty big :-)

tthvo:
> > > Speaking of toolbars, I just noticed something about the `Duration` filter in the toolbar.
> > > The two "started before/after" filters have a magnifying glass search button to perform the filter after selecting the date/time parameters. The other filters have a dropdown menu (`Select`) and selecting an item immediately starts the filtering with that item. All good so far.
> > > For the Duration filter, clicking the `Continuous` checkbox starts filtering for recordings with a `continuous` duration. Unchecking the box removes the filter. There doesn't seem to be a way to filter for recordings that are _not_ continuous, other than by using the numeric duration filter.
> > > The numeric duration filter only allows for matching exact recording durations, not greater-than or less-than, which I think is okay (for now), but to actually add this filter the user has to click the numeric input field and then press Enter.
> > > Here's my idea for how to make this filter control a little more intuitive, at least to me:
> > > ```
> > > 1. Add a magnifying glass button like the "started before/after" filter which actually adds a filter based on the selected criteria
> > > 
> > > 2. Remove the static "Continuous" checkbox
> > > 
> > > 3. Add a `Select` with values `Continuous`, `Greater Than`, `Less Than`
> > > 
> > > 4. Add a secondary control that varies according to the value selected above:
> > >    4a. If `Continuous`, dynamically display a "Continuous" checkbox. The state of this checkbox becomes a "continuous == state" filter when the magnifying glass is clicked. This way the user can filter by only continuous recordings, or only fixed durations
> > >    4b. If `Greater Than` or `Less Than`, dynamically display a duration picker. I think this should match the duration field on the recording creation form, with a number input and a separate units input. When the magnifying glass is clicked then apply a filter that looks for recordings fitting the specific criteria, ex. "recordings with fixed duration less than 5 minutes".
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This is best suited for a separate PR, in any case.
> > 
> > 
> > Great idea! I think I could look into this for this PR too as it seems within scope :D
> 
> I'd rather it as a separate PR, just so it's a little easier to review. This one's already pretty big :-)

Ahh right! Okayy I will leave it as now.

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -462,115 +472,68 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
...
+    setExpandedRows(expandedRows => {
+      const idx = expandedRows.indexOf(id);
+      return idx >= 0 ? [...expandedRows.slice(0, idx), ...expandedRows.slice(idx + 1, expandedRows.length)] : [...expandedRows, id]
+    });
+  }, [expandedRows, setExpandedRows]);
 
-  const handleWarningModalClose = React.useCallback(() => {
-    setWarningModalOpen(false);
-  }, [setWarningModalOpen]);
+  const isDeleteDialogEnabled = React.useMemo(
```
maxcao13, line 482: Since the deletion warning thing is in the toolbar now, this check might be better off in there as well.

src/app/RecordingMetadata/LabelCell.tsx
```diff
@@ -36,21 +36,60 @@
...
 
 export const LabelCell: React.FunctionComponent<LabelCellProps> = (props) => {
-  // TODO make labels clickable to select multiple recordings with the same label
+  const isLabelSelected = React.useCallback((label: RecordingLabel) => {
+    if (props.clickableOptions) {
+      const labelFilterSet = new Set(props.clickableOptions.labelFilters);
+      return labelFilterSet.has(getLabelDisplay(label))
+    }
+    return false;
+  }, [getLabelDisplay]);
```
maxcao13, line 62: should props.clickableOptions be in deps?

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -155,9 +148,13 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
...
     addSubscription(
-      context.target.target().subscribe(refreshRecordingList)
+      context.target.target().subscribe((target) => {
+        setTargetConnectURL(target.connectUrl);
+        dispatch(addTargetIntent(target.connectUrl));
+        refreshRecordingList();
+      })
     );
-  }, [addSubscription, context, context.target, refreshRecordingList]);
+  }, [addSubscription, context, context.target, refreshRecordingList, setTargetConnectURL]);
```
maxcao13, line 157: dispatch and addTargetIntent in deps?

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -312,57 +319,48 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
...
+      } else {
+        dispatch(deleteFilterIntent(target, filterKey, filterValue, false));
+      }
+    } else {
+      dispatch(addFilterIntent(target, filterKey, filterValue, false));
     }
-    const id = window.setInterval(() => refreshRecordingList(), context.settings.autoRefreshPeriod() * context.settings.autoRefreshUnits());
-    return () => window.clearInterval(id);
-  }, [refreshRecordingList, context, context.settings]);
+  }, [dispatch]);
```
maxcao13, line 336: intents in deps?

src/app/Recordings/ActiveRecordingsTable.tsx
tthvo: Right makes sense :D I will fix that!

tthvo:
Just noticed: I did change a field name in filter categories so local storage is not updated unless changing the filters. This will cause error in rendering (i.e. blank page). To fix, clear cache or use incognito :D

maxcao13 (dismissed):
Looks good to me, great work on this! About the error, would we have to tell users to clear cache when they update to the new release?

tthvo:
> Looks good to me, great work on this! About the error, would we have to tell users to clear cache when they update to the new release?

Hmm local storage is only implemented for filters since this PR. So, I think probably not :D

tthvo:
> Timepicker needs to render correctly (i.e. spacings are misaligned). Appending the popever menu to component higher in the tree solved the issue.

For this point, I initially wanted to mount the menu to the drawer to avoid accessibility issue, using `getElementById`. However, this approach is not safe as DOM element might not be available or already removed. To fix, just mount to `document.body` in risk of accessibility.

andrewazores:
Found a bug.

1. Start an active recording manually on some sample application
2. Wait a moment, then archive it
3. Wait another moment and archive it again
4. Switch to the Archived Recordings tab for that application
5. Click the `template.name: Continuous` label chip.

image

A `Labels` (plural) filter type appears and with a `State` dropdown for `RUNNING | STARTING | STOPPING | STOPPED` states.

tthvo:
The only left

> Found a bug.
> 
>     1. Start an active recording manually on some sample application
> 
>     2. Wait a moment, then archive it
> 
>     3. Wait another moment and archive it again
> 
>     4. Switch to the Archived Recordings tab for that application
> 
>     5. Click the `template.name: Continuous` label chip.
> 
> 
> image
> 
> A `Labels` (plural) filter type appears and with a `State` dropdown for `RUNNING | STARTING | STOPPING | STOPPED` states.

Ahh sorry. It was meant `LabelCell` didn't update its category to singular. Fixed it now :D

andrewazores (approved):

tthvo:
Sorry, just a final touch to make sure the filter is rendered correctly. The category dropdown now does not depend on the passed recording filters but rather the static array to ensure the above (state filter pops up in archived view) issue does not appear.

andrewazores (approved):

maxcao13 (approved):
LGTM!

PR #511, opened 2022-09-15
fix(recordings): labels should be applied to created recordings

### LINKED ISSUE #500: [Task] Cannot add labels to new recording using CreateRecording form
1. Create a recording using the CreateRecording form on any target and set metadata
2. The metadata labels are not updated with the new label.

Editing the recording afterwards with new labels is completely fine however.

tthvo:
Fixes #500 

Labels should be added to newly created recording.


maxcao13 (approved):
Git blame is on me, nice spotting :)

andrewazores (approved):

PR #515, opened 2022-09-20
fix(about): correct logo height using \!important override

### LINKED ISSUE #514: [Task] Cryostat logo should be responsive and size-constrained
Our logo seems to be render pretty oddly (i.e. quite large). I think we might need a set of responsive logos and add some size constraints to them.

Screenshot from 2022-09-20 12-03-27

andrewazores:
override some Patternfly styling update that broke this view

Fixes #514

### DIFF
--- a/src/app/app.css
+++ b/src/app/app.css
@@ -58,7 +58,7 @@ html, body, #root {
   padding-right: 10rem;
 }
 .cryostat-logo {
-  height: 2em;
+  height: 2em !important;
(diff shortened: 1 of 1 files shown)


tthvo (approved):
Ahh right! Make sense! 

Patternfly's updated [docs](https://www.patternfly.org/v4/components/brand) on `Brand` show a large horizontal logo in samples too.

PR #524, opened 2022-09-27
feat(jmx): re-implement enhanced transient JMX credentials

### LINKED ISSUE #523: [Task] Re-implement transient JMX credentials
Currently, JMX credentials are always stored in the Cryostat backend, whether the user adds them through the Security panel or by the Auth modal prompt that appears when trying to ex. view active recordings for a target and the backend has no matching credentials.

When the auth modal prompt appears for an interactive client trying to do something like view active recordings, those credentials should not automatically be stored in backend storage. The user should have the option of allowing these credentials to persist only for the duration of their interactive browser session. This option should probably be the default.

Related https://github.com/cryostatio/cryostat/pull/1082 / https://github.com/cryostatio/cryostat/pull/1083

Specifically, see the comment at the bottom of https://github.com/cryostatio/cryostat/pull/1083 . It used to be that the Cryostat backend could not store any JMX credentials, and clients were required to always pass them through the `X-JMX-Authorization` header with requests to targets requiring credentials. This was more secure, but incompatible with the concept of an Automated Rule where Cryostat can open connections to targets "unattended", where there is no client supplying the `X-JMX-Authorization` header to pass through. Once credentials storage for rules was implemented the client-side transient credentials storage and `X-JMX-Authorization` support was dropped, though the Cryostat backend still does support that header and passthrough functionality. The web-client support for it should be brought back and improved so that users are not /required/ to use the backend credentials storage for interactive use cases, and can continue to use the client-side transient storage and JMX credential passthrough if preferred. This will not work for Automated Rules, but the user can choose not to use those and therefore to never store credentials on the backend.

andrewazores:
Fixes #523 . See there for a description of why this change is useful/necessary and the historical context.

TODO later, maybe: implement a client-side keyring. ie store the JMX credentials in local storage with a user-provided encryption key.


src/app/Settings/CredentialsStorage.tsx
```diff
@@ -0,0 +1,121 @@
...
+      setSelection(location);
+      setExpanded(false);
+      saveToLocalStorage('JMX_CREDENTIAL_LOCATION', selection);
+    },
+    [getLocation, setSelection, setExpanded, saveToLocalStorage]
+  );
+
+  React.useEffect(() => {
+    handleSelect(undefined, getFromLocalStorage('JMX_CREDENTIAL_LOCATION', Locations.BROWSER_SESSION));
+  }, [handleSelect, getLocation, getFromLocalStorage]);
```
maxcao13, line 91: Does getLocation need to be a dep here? From my understanding handleSelect already has getLocation as a dep so the handleSelect function would change anyways, but I'm not completely sure.

src/app/Settings/CredentialsStorage.tsx
```diff
@@ -0,0 +1,121 @@
...
+      </Select>
+    </>
+  );
+};
+
+export const CredentialsStorage: UserSetting = {
+  title: 'JMX Credentials Storage',
+  description: (
+    <Text>
+      When you attempt to connect to a target application which requires authentication you will see a prompt for
```
maxcao13, line 114: Comma should be here I think.
`requires authentication, you will see a prompt for`

maxcao13 (commented):
Everything else looks good, the ability to choose whether credentials are stored in backend or per session works well, there is something somewhat unrelated but there seems to be a memory-leak with the JMXAuthModal even though it wasn't touched in this PR. 

This gets printed to the console 
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
JmxAuthForm@webpack-internal:///./src/app/AppLayout/JmxAuthForm.tsx:53:54
// etc... etc...
```
I think it's just that this subscription on line 61 isn't closed.
```
  const handleSave = React.useCallback(() => {
    props.onSave(username, password).then(() => {
      clear();
      context.target.setAuthRetry(); // <--------------
    });
  }, [context, context.target, clear, props.onSave, username, password]);

```

src/app/Settings/CredentialsStorage.tsx
andrewazores: Oh yea, thanks. This used to actually call `getLocation` before I cleaned things up and opened the PR. You're right that it isn't needed here anymore.

maxcao13:
Oh, whoops my bad, then it's something else internal then I guess, hm...

tthvo:
> > I think it's just that this subscription on line 61 isn't closed.
> 
> Hmm, I don't think there's a subscription there... that method just returns `void`.
> 
> @tthvo has seen and investigated these "state update on an unmounted component" messages lately. Any insights here?

Ohh yesss, here:
https://github.com/cryostatio/cryostat-web/blob/ec23856cc0dfeeb5b61beab39a94225003d5ff56/src/app/AppLayout/JmxAuthForm.tsx#L60

There is a call to `setState` in `JMXAuthForm` to clear input. However, when saving, the modal is already closed (perhaps unmount its body and remove from tree internally ), this call will be invalid since calling on unmounted components.

To fix, just remove it as it is not needed.

src/app/SecurityPanel/Credentials/StoreJmxCredentials.tsx
```diff
@@ -457,6 +459,14 @@ export const StoreJmxCredentials = () => {
...
 export const StoreJmxCredentialsCard: SecurityCard = {
   title: 'Store JMX Credentials',
-  description: `Credentials that Cryostat uses to connect to target JVMs over JMX are stored here.`,
+  description: (
+    <Text>
+      Credentials that Cryostat uses to connect to target JVMs over JMX are stored here. These are stored in encrypted
+      storage managed by the Cryostat backend. These credentials may be used for manually managing recordings and event
+      templates on target JVMs, as well as for Automated Rules which run in the background and open unattended target
+      connections. Any locally-stored client credentials held by your browser session are not displayed here. See{' '}
+      <Link to="/settings">Settings</Link> to configure locally-stored credentials.
```
tthvo, line 468: Maybe, `to configure how/where to store credentials` or something similar because the setting toggles the behaviour?

src/app/Settings/Settings.tsx
```diff
@@ -40,18 +40,20 @@ import * as React from 'react';
...
+  const settings = [NotificationControl, CredentialsStorage, DeletionDialogControl, WebSocketDebounce, AutoRefresh].map(
+    (c) => ({
+      title: c.title,
+      description: c.description,
+      element: React.createElement(c.content, null),
+    })
+  );
   return (
     <>
       <BreadcrumbPage pageTitle="Settings">
```
tthvo, line 59: Each setting cards should be assigned a key (probably title?):

```bash
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `Settings`. See https://reactjs.org/link/warning-keys for more information.
Settings
Route@webpack-internal:///./node_modules/react-router/esm/react-router.js:448:29
RouteWithTitleUpdates@webpack-internal:///./src/app/routes.tsx:223:21
Switch@webpack-internal:///./node_modules/react-router/esm/react-router.js:650:29
LastLocationProvider@webpack-internal:///./node_modules/react-router-last-location/dist/index.js:161:47
C@webpack-internal:///./node_modules/react-router/esm/react-router.js:705:31
AppRoutes@webpack-internal:///./src/app/routes.tsx:237:54
main
div
DrawerContentBody@webpack-internal:///./node_modules/@patternfly/react-core/dist/esm/components/Drawer/DrawerContentBody.js:16:54
div
div
DrawerMain@webpack-internal:///./node_modules/@patternfly/react-core/dist/esm/components/Drawer/DrawerMain.js:16:34
DrawerContent@webpack-internal:///./node_modules/@patternfly/react-core/dist/esm/components/Drawer/DrawerContent.js:20:131
div
Drawer@webpack-internal:///./node_modules/@patternfly/react-core/dist/esm/components/Drawer/Drawer.js:30:138
div
div
Page@webpack-internal:///./node_modules/@patternfly/react-core/dist/esm/components/Page/Page.js:49:9
AppLayout@webpack-internal:///./src/app/AppLayout/AppLayout.tsx:101:20
Router@webpack-internal:///./node_modules/react-router/esm/react-router.js:79:30
BrowserRouter@webpack-internal:///./node_modules/react-router-dom/esm/react-router-dom.js:57:35
Provider@webpack-internal:///./node_modules/react-redux/es/components/Provider.js:14:18
App
```

src/app/Shared/Services/JmxCredentials.service.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export class JmxCredentials {
+  // TODO replace with Redux?
+  private readonly store = new Map<string, Credential>();
+
+  constructor(private readonly api: () => ApiService) {}
+
+  setCredential(targetId: string, username: string, password: string): Observable<boolean> {
+    let location = getFromLocalStorage('JMX_CREDENTIAL_LOCATION', Locations.BROWSER_SESSION);
+    switch (location) {
```
tthvo, line 56: I think we should be comparing `location.key`?

src/app/Shared/Services/JmxCredentials.service.tsx
```diff
@@ -0,0 +1,81 @@
...
+      case Locations.BACKEND.key:
+        return this.api().postCredentials(`target.connectUrl == "${targetId}"`, username, password);
+      case Locations.BROWSER_SESSION.key:
+        this.store.set(targetId, { username, password });
+        return of(true);
+      default:
+        console.warn('Unknown storage location', location);
+        return of(true);
```
tthvo, line 64: Should return `false`?

src/app/Shared/Services/Services.tsx
```diff
@@ -44,19 +44,22 @@ import { NotificationChannel } from './NotificationChannel.service';
...
+  jmxCredentials: JmxCredentials;
   notificationChannel: NotificationChannel;
   reports: ReportService;
   settings: SettingsService;
   login: LoginService;
 }
 
 const settings = new SettingsService();
-const login = new LoginService(TargetInstance, settings);
+const jmxCredentials = new JmxCredentials(() => api);
```
tthvo, line 61: Should it be defined after defining `api` (just to avoid surprises)? Interestingly, JS/TS allows this.

src/app/SecurityPanel/Credentials/StoreJmxCredentials.tsx
andrewazores: The toggle doesn't disable backend-stored credentials though, it only controls whether the JMX auth modal persists the credentials in the backend or holds them only in browser memory, so I would prefer to keep the "locally-stored" phrasing for this case. The note doesn't really apply to this credentials table, it's just there to explain to the user why they may not see credentials that they entered in the dialog and which they can observe are still present somewhere and usable since they can re-connect to the same target without re-entering credentials within the same browser session.

src/app/Settings/CredentialsStorage.tsx
```diff
@@ -0,0 +1,121 @@
...
+  static readonly BACKEND: Location = {
+    key: 'Backend',
+    description:
+      'Keep credentials in encrypted Cryostat backend storage. These credentials will be available to other users and will be used for Automated Rules.',
+  };
+}
+
+const locations = [Locations.BROWSER_SESSION, Locations.BACKEND];
+
+function getLocation(key: string): Location {
```
tthvo, line 66: We could use arrow-func here for consistency?

tthvo (commented):
Looks good! Just a couple comments. Also, it seems there are no current unit tests for this `Setting` view. I think we should add them :D

src/app/SecurityPanel/Credentials/StoreJmxCredentials.tsx
tthvo: Right! Makes sense! Thanks for explaining that!

src/app/Shared/Services/Services.tsx
andrewazores: It can't be, because the `login` has a direct reference to `jmxCredentials`, and `api` has a direct reference to `login`. I used `() => api` as a forward-reference to avoid the circular dependency problem, similar to what you would see with `Lazy` in Dagger terms on the backend.

tthvo: Wow, interesting! sorry didn't notice the below lines. Thanks for explaining again :D

maxcao13: > It can't be, because the `login` has a direct reference to `jmxCredentials`, and `api` has a direct reference to `login`. I used `() => api` as a forward-reference to avoid the circular dependency problem, similar to what you would see with `Lazy` in Dagger terms on the backend.

To solve the circular dependency problem on the backend I've used Provider<>, are there merits for using either?

andrewazores: Some details on the difference here: https://dagger.dev/api/2.0/dagger/Lazy.html

In short, for the use cases you've used it on the backend it probably works out the same. If you're injecting something using `Provider` then that something gets instantiated right away, before `provider.get()` is called. `Lazy` will only perform the instantiation when `lazy.get()` is called (hence the name), so if nothing ever calls `lazy.get()` then the "something" is never instantiated. But on the backend, anything you're injecting with either a Provider or a Lazy is also a `Singleton`, and `.get()` will be called, so either one should work essentially equivalently. If the `new Something()` threw an exception then I suppose you'd get different stack traces for a Provider vs Lazy.

All that said, this example of `() => api` is in fact more akin to a Dagger Provider than a Lazy :-)

src/test/Settings/CredentialsStorage.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  it('sets value to local storage when dropdown is clicked', async () => {
+    render(React.createElement(CredentialsStorage.content, null));
+
+    expect(getFromLocalStorage).toHaveBeenCalledTimes(1);
+    expect(saveToLocalStorage).toHaveBeenCalledTimes(1);
+    expect(saveToLocalStorage).lastCalledWith(storageKey, sessionStorageValue);
+
+    userEvent.click(screen.getByRole('button'));
```
tthvo, line 85: we should check if `Session (Browser Memory)` is displayed by default?

src/test/Settings/CredentialsStorage.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    await screen.findByRole('listbox');
+    const ul = screen.getByRole('listbox');
```
tthvo, line 88: I think this line could be merged with above as `const ul = await screen.findByRole('listbox')`

src/test/Settings/CredentialsStorage.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const session = getByText(ul, sessionStorageValue);
```
tthvo, line 89: Probably, use `screen.getByText` for consistency?

src/test/Settings/CredentialsStorage.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const backend = getByText(ul, backendStorageValue);
+    userEvent.click(backend);
+    await waitFor(() => expect(session).not.toBeVisible());
+
```
tthvo, line 93: Would need to check if `Backend` is displayed?

src/test/Settings/CredentialsStorage.test.tsx
andrewazores: I'm querying for children under the `<ul>` because it seems that the Patternfly Select component actually makes a few different child elements with that same inner text:

```
$ yarn test src/test/Settings/CredentialsStorage.test.tsx 
yarn run v1.22.17
$ jest --maxWorkers=50% --coverage=true src/test/Settings/CredentialsStorage.test.tsx
 FAIL  src/test/Settings/CredentialsStorage.test.tsx
  <CredentialsStorage/>
    ✓ renders correctly (19 ms)
    ✓ defaults to Session storage (62 ms)
    ✕ sets value to local storage when dropdown is clicked (51 ms)

  ● <CredentialsStorage/> › sets value to local storage when dropdown is clicked

    TestingLibraryElementError: Found multiple elements with the text: Session (Browser Memory)

    Here are the matching elements:

    Ignored nodes: comments, <script />, <style />
    <span
      class="pf-c-select__toggle-text"
    >
      Session (Browser Memory)
    </span>

    Ignored nodes: comments, <script />, <style />
    <span
      class="pf-c-select__menu-item-main"
    >
      Session (Browser Memory)
      <span
        class="pf-c-select__menu-item-icon"
      >
        <svg
          aria-hidden="true"
          fill="currentColor"
          height="1em"
          role="img"
          style="vertical-align: -0.125em;"
          viewBox="0 0 512 512"
          width="1em"
        >
          <path
            d="M173.898 439.404l-166.4-166.4c-9.997-9.997-9.997-26.206 0-36.204l36.203-36.204c9.997-9.998 26.207-9.998 36.204 0L192 312.69 432.095 72.596c9.997-9.997 26.207-9.997 36.204 0l36.203 36.204c9.997 9.997 9.997 26.206 0 36.204l-294.4 294.401c-9.998 9.997-26.207 9.997-36.204-.001z"
          />
        </svg>
      </span>
    </span>

    (If this is intentional, then use the `*AllBy*` variant of the query (like `queryAllByText`, `getAllByText`, or `findAllByText`)).

    Ignored nodes: comments, <script />, <style />
    <body>
      <div>
        <div
          class="pf-c-select pf-m-expanded"
          data-ouia-component-id="OUIA-Generated-Select-single-3"
          data-ouia-component-type="PF4/Select"
          data-ouia-safe="true"
        >
          <button
            aria-expanded="true"
            aria-haspopup="listbox"
            aria-label="Options menu"
            aria-labelledby=" pf-select-toggle-id-3"
            class="pf-c-select__toggle"
            id="pf-select-toggle-id-3"
            type="button"
          >
            <div
              class="pf-c-select__toggle-wrapper"
            >
              <span
                class="pf-c-select__toggle-text"
              >
                Session (Browser Memory)
              </span>
            </div>
            <span
              class="pf-c-select__toggle-arrow"
            >
              <svg
                aria-hidden="true"
                fill="currentColor"
                height="1em"
                role="img"
                style="vertical-align: -0.125em;"
                viewBox="0 0 320 512"
                width="1em"
              >
                <path
                  d="M31.3 192h257.3c17.8 0 26.7 21.5 14.1 34.1L174.1 354.8c-7.8 7.8-20.5 7.8-28.3 0L17.2 226.1C4.6 213.5 13.5 192 31.3 192z"
                />
              </svg>
            </span>
          </button>
          <ul
            aria-label=""
            class="pf-c-select__menu"
            role="listbox"
          >
            <li
              class="pf-c-select__menu-wrapper"
              id="select-option-1664471309403h7v5xg5ychb"
              role="presentation"
            >
              <button
                aria-selected="true"
                class="pf-c-select__menu-item pf-m-selected pf-m-description"
                role="option"
                type="button"
              >
                <span
                  class="pf-c-select__menu-item-main"
                >
                  Session (Browser Memory)
                  <span
                    class="pf-c-select__menu-item-icon"
                  >
                    <svg
                      aria-hidden="true"
                      fill="currentColor"
                      height="1em"
                      role="img"
                      style="vertical-align: -0.125em;"
                      viewBox="0 0 512 512"
                      width="1em"
                    >
                      <path
                        d="M173.898 439.404l-166.4-166.4c-9.997-9.997-9.997-26.206 0-36.204l36.203-36.204c9.997-9.998 26.207-9.998 36.204 0L192 312.69 432.095 72.596c9.997-9.997 26.207-9.997 36.204 0l36.203 36.204c9.997 9.997 9.997 26.206 0 36.204l-294.4 294.401c-9.998 9.997-26.207 9.997-36.204-.001z"
                      />
                    </svg>
                  </span>
                </span>
                <span
                  class="pf-c-select__menu-item-description"
                >
                  Keep credentials in browser memory for the current session only. When you close this browser tab the credentials will be forgotten.
                </span>
              </button>
            </li>
            <li
              class="pf-c-select__menu-wrapper"
              id="select-option-1664471309405o8wzrm48rcs"
              role="presentation"
            >
              <button
                class="pf-c-select__menu-item pf-m-description"
                role="option"
                type="button"
              >
                <span
                  class="pf-c-select__menu-item-main"
                >
                  Backend
                </span>
                <span
                  class="pf-c-select__menu-item-description"
                >
                  Keep credentials in encrypted Cryostat backend storage. These credentials will be available to other users and will be used for Automated Rules.
                </span>
              </button>
            </li>
          </ul>
        </div>
      </div>
    </body>

       97 |
       98 |     const ul = await screen.findByRole('listbox');
    >  99 |     const session = screen.getByText(sessionStorageValue);
          |                            ^
      100 |     const backend = getByText(ul, backendStorageValue);
      101 |     userEvent.click(backend);
      102 |     await waitFor(() => expect(session).not.toBeVisible());

      at Object.getElementError (node_modules/@testing-library/dom/dist/config.js:38:19)
      at getElementError (node_modules/@testing-library/dom/dist/query-helpers.js:25:35)
      at getMultipleElementsFoundError (node_modules/@testing-library/dom/dist/query-helpers.js:29:10)
      at node_modules/@testing-library/dom/dist/query-helpers.js:66:13
      at node_modules/@testing-library/dom/dist/query-helpers.js:111:19
      at src/test/Settings/CredentialsStorage.test.tsx:99:28
      at step (src/test/Settings/CredentialsStorage.test.tsx:52:23)
      at Object.next (src/test/Settings/CredentialsStorage.test.tsx:33:53)
      at fulfilled (src/test/Settings/CredentialsStorage.test.tsx:24:58)
```

tthvo: Ahh wow right! Then, maybe we can use:

`within(ul).getByText("sessionStorageValue")`

Just a bit clearer what is being done :D

src/test/Settings/CredentialsStorage.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+    expect(saveToLocalStorage).toHaveBeenCalledTimes(1);
+    expect(saveToLocalStorage).lastCalledWith(storageKey, sessionStorageValue);
+
+    userEvent.click(screen.getByRole('button'));
+
+    const ul = await screen.findByRole('listbox');
+    const session = getByText(ul, sessionStorageValue);
+    const backend = getByText(ul, backendStorageValue);
+    userEvent.click(backend);
+    await waitFor(() => expect(session).not.toBeVisible());
```
tthvo, line 102: I am thinking this should wait for the menu to close as it is closer to what the user would see and expect?

src/test/Shared/Services/JmxCredentials.service.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+describe('JmxCredentials.service', () => {
+  let svc: JmxCredentials;
+  beforeEach(() => {
+    saveToLocalStorage('JMX_CREDENTIAL_LOCATION', undefined);
+    postCredentialsSpy.mockReset();
+    svc = new JmxCredentials(() => mockApi);
+  });
+
+  it('should not check storage on instantiation', () => {
+    expect(getFromLocalStorage as jest.Mock).not.toHaveBeenCalled();
```
tthvo, line 74: we can leave the casting out here right?

tthvo (approved):
Looks great to me!

PR #526, opened 2022-09-29
chore(deps): update to React 18

### LINKED ISSUE #482: [Task] Refactor React.FunctionalComponent/React.FC using codemod-replace-react-fc-typescript
> I haven't seen that before. I don't really have any thoughts on the matter or opinion one way or the other.
> 
> https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/
> 
> I tend to like the explicit/static types where possible, but sometimes it isn't really necessary - and if the type signature is actually lying to you, like it is about the implicit children, then it does more harm than good.
> 
> If you'd like to go through and refactor to remove all those type declarations later then I wouldn't be opposed.

Maybe I will do a refactor in the future ([the docs you linked has something for that anyways](https://github.com/gndelia/codemod-replace-react-fc-typescript)), but for now I think it may be best to stop using React.FC in `-web` in the future and stick with the usual 
```
const Foo = ( {prop1, prop2}: FooProps ): JSX.Element => {
   ...
   return ...
}
``` 
so that everything is using the same and preferred style. 

I'll open an issue for it, if that's okay with you anyways. This isn't a big deal by any means ^^. I'm not sure what is the best way to let everyone know though, haha.

_Originally posted by @maxcao13 in https://github.com/cryostatio/cryostat-web/issues/481#issuecomment-1183747147_

maxcao13:
Fixes #482 
Depends on https://github.com/patternfly/patternfly-react/issues/7142

What's been updated:
``` 
"react": "^17.0.2" => "^18.2.0",
"react-dom": "^17.0.2" => "react-dom": "^18.2.0",

@testing-library/dom: 8.11.3 => 8.18.1
@testing-library/jest-dom: 5.16.2 => 5.16.5
@testing-library/react: 12.1.4 => 13.4.0
@testing-library/user-event: 13.5.0 => 14.4.3
```

What's been added:
```
"@types/react": "^18.0.21", // typescript react types
"@types/react-dom": "^18.0.6",

"react-test-renderer": "^18.2.0",
```

Noteworthy: the update of `@testing-library/user-event` has made all their api calls to return promises, await and async should be used for those calls in our tests now. https://github.com/testing-library/user-event/releases/tag/v14.0.0

### DIFF
--- a/src/test/Recordings/ActiveRecordingsTable.test.tsx
+++ b/src/test/Recordings/ActiveRecordingsTable.test.tsx
@@ -116,6 +116,7 @@ jest.spyOn(defaultServices.api, 'doGet').mockReturnValue(of([mockRecording]));
 jest.spyOn(defaultServices.api, 'downloadRecording').mockReturnValue();
 jest.spyOn(defaultServices.api, 'downloadReport').mockReturnValue();
 jest.spyOn(defaultServices.api, 'grafanaDatasourceUrl').mockReturnValue(of('/grafanaUrl'));
+jest.spyOn(defaultServices.api, 'grafanaDashboardUrl').mockReturnValue(of('/grafanaUrl'));
 jest.spyOn(defaultServices.api, 'stopRecording').mockReturnValue(of(true));
 jest.spyOn(defaultServices.api, 'uploadActiveRecordingToGrafana').mockReturnValue(of(true));
 
@@ -267,18 +268,18 @@ describe('<ActiveRecordingsTable />', () => {
     expect(screen.getByText('Delete')).toBeInTheDocument();
   });
 
-  it('routes to the Create Flight Recording form when Create is clicked', () => {
+  it('routes to the Create Flight Recording form when Create is clicked', async () => {
     renderWithServiceContextAndReduxStoreWithRoute(<ActiveRecordingsTable archiveEnabled={true} />, {
       preloadState: preloadedState,
       history: history,
     });
 
-    userEvent.click(screen.getByText('Create'));
+    await userEvent.click(screen.getByText('Create'));
 
     expect(history.entries.map((entry) => entry.pathname)).toStrictEqual(['/recordings', '/recordings/create']);
   });
 
-  it('archives the selected recording when Archive is clicked', () => {
+  it('archives the selected recording when Archive is clicked', async () => {
     renderWithServiceContextAndReduxStoreWithRoute(<ActiveRecordingsTable archiveEnabled={true} />, {
       preloadState: preloadedState,
       history: history,
@@ -286,8 +287,8 @@ describe('<ActiveRecordingsTable />', () => {
 
     const checkboxes = screen.getAllByRole('checkbox');
     const selectAllCheck = checkboxes[0];
-    userEvent.click(selectAllCheck);
-    userEvent.click(screen.getByText('Archive'));
+    await userEvent.click(selectAllCheck);
+    await userEvent.click(screen.getByText('Archive'));
 
     const archiveRequestSpy = jest.spyOn(defaultServices.api, 'archiveRecording');
 
@@ -295,7 +296,7 @@ describe('<ActiveRecordingsTable />', () => {
     expect(archiveRequestSpy).toBeCalledWith('someRecording');
   });
 
-  it('stops the selected recording when Stop is clicked', () => {
+  it('stops the selected recording when Stop is clicked', async () => {
     renderWithServiceContextAndReduxStoreWithRoute(<ActiveRecordingsTable archiveEnabled={true} />, {
       preloadState: preloadedState,
       history: history,
@@ -303,8 +304,8 @@ describe('<ActiveRecordingsTable />', () => {
 
     const checkboxes = screen.getAllByRole('checkbox');
     const selectAllCheck = checkboxes[0];
-    userEvent.click(selectAllCheck);
-    userEvent.click(screen.getByText('Stop'));
+    await userEvent.click(selectAllCheck);
+    await userEvent.click(screen.getByText('Stop'));
 
     const stopRequestSpy = jest.spyOn(defaultServices.api, 'stopRecording');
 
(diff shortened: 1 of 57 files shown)


andrewazores:
`yarn format:apply` should fix the CI failure, or at least let it advance to the next step.

maxcao13:
Should `es-lint` formatting be applied too? I noticed it's not include in the format command.

andrewazores:
> Should `es-lint` formatting be applied too? I noticed it's not included in the format command.

The last time I tried it, which was quite a while ago, the auto-formatting it applied actually broke things and needed manual cleanup by hand, and it was quite a significant amount of busywork. I just tried it again now and it looks like it applies some safe fixes fine, though there are many other linting warnings still present after.

I think you can update the `format:apply` script to also run `eslint:apply` since it seems safe. Don't add `eslint:check` to `format:check` yet because there will still be a bunch of lint warnings emitted right now, and adding it to that script would cause builds to fail until we clean up the lint issues.

src/test/Recordings/Filters/DateTimePicker.test.tsx
```diff
@@ -46,6 +46,11 @@ const onDateTimeSelect = jest.fn((date) => {
 });
 const currentDate = new Date('14 Sep 2022 00:00:00 UTC');
 
+// See https://github.com/testing-library/user-event/issues/833 with new testing-library versions
```
andrewazores, line 49: Going by https://github.com/testing-library/user-event/issues/833#issuecomment-1171452841 , would it be better to use a mock test-controlled timer rather than disabling the delays altogether?

https://jestjs.io/docs/timer-mocks

src/test/Recordings/ActiveRecordingsTable.test.tsx
```diff
@@ -219,45 +222,55 @@ describe('<ActiveRecordingsTable />', () => {
...
     expect(screen.getByText('STOPPED')).toBeInTheDocument();
     expect(screen.queryByText('RUNNING')).not.toBeInTheDocument();
   });
 
-  it('removes a recording after receiving a notification', () => {
-    renderWithServiceContextAndReduxStoreWithRoute(<ActiveRecordingsTable archiveEnabled={true} />, {
-      preloadState: preloadedState,
-      history: history,
+  it('removes a recording after receiving a notification', async () => {
+    await act(async () => {
```
tthvo, line 259: Some reasons behind wrapping these calls in `act`?

src/test/Recordings/ActiveRecordingsTable.test.tsx
```diff
@@ -175,6 +176,8 @@ jest
   .mockReturnValueOnce(of(mockDeleteNotification))
   .mockReturnValue(of()); // all other tests
 
+jest.spyOn(window, 'open').mockReturnValue(null);
```
tthvo, line 179: Oh what happened that we need this mocked?

maxcao13:
There are some rendering problems when actually running the web-client, it just seems a lot more clunky and slower to use. Currently investigating the reason.

src/test/Recordings/ActiveRecordingsTable.test.tsx
maxcao13: The RecordingActions component calls grafanaUpload which calls context.api.grafanaDashboardUrl which wasn't mocked before and for some reason wasn't called either. But now it does, and that has a callback to window.open. In the ArchivedRecordingsTable test there is a similar jest.spyOn so I just reused it.

src/test/Recordings/ActiveRecordingsTable.test.tsx
maxcao13: There was a lot of console warnings with the tests before adding that, but other than that not actually sure of the underlying reason why that fixes it.

src/test/Recordings/ActiveRecordingsTable.test.tsx
tthvo: Make sense thanks!!

src/test/Recordings/ActiveRecordingsTable.test.tsx
tthvo: Pretty odd! This only happens with `*RecordingsTable` components :D

maxcao13:
> There are some rendering problems when actually running the web-client with React 18, it just seems a lot more clunky and slower to use. Currently investigating the reason.

After a lot of digging and testing, there doesn't seem to be many problems from users online with React 18 in general with what is going on here - Thuan's PR does seems to speed up rendering but the main problem is that intermediate and default states seems to flicker in between rendering. The most obvious replication of this is if you click on the Recordings tab with an Active Recording and click to a different tab and keep clicking back and forth. So I'm beginning to wonder if it is a Patternfly issue or not since they do not fully support React 18 yet, and won't do so until their next release in early 2023. https://github.com/patternfly/patternfly-react/issues/7142#issuecomment-1261128196

src/test/Recordings/ActiveRecordingsTable.test.tsx
tthvo: Seems like using `iframe` for ReportView causes this issue. It uses `onLoad` callback that updates React states outside React callstack, which throws a warning in tests. Surprisingly, this didn't get caught in old versions of React testing.

https://github.com/cryostatio/cryostat-web/blob/19f353f30d8783dae8e7501f75c1a4fbe1cf3311/src/app/Recordings/ReportFrame.tsx#L83

andrewazores:
> This PR/issue depends on:

Not actually resolved, just autoclosed due to inactivity.

maxcao13:
Patternfly 5 has been released with full support of React 18 https://medium.com/patternfly/patternflys-major-release-everything-you-need-to-know-62f7c890ff57

Here's a full upgrade article: https://www.patternfly.org/get-started/upgrade/

maxcao13:
Probably be better to start from scratch here.

PR #529, opened 2022-09-30
chore(hooks): update useSubscriptions to reduce re-rendering

### LINKED ISSUE #528: [Task] Update useSubscriptions hook to reduce re-rendering
Right now, we define useSubscriptions:

```typescript
export function useSubscriptions() {
  const subsRef = React.useRef([] as Subscription[]);
  const [subscriptions, setSubscriptions] = React.useState(subsRef.current);

  React.useEffect(() => () => subsRef.current.forEach((s: Subscription): void => s.unsubscribe()), []);

  const addSubscription = (sub: Subscription): void => {
    setSubscriptions((subs: Subscription[]): Subscription[] => {
      const result = subs.concat([sub]);
      subsRef.current = result;
      return result;
    });
  };

  return React.useCallback(addSubscription, [setSubscriptions]);
}
```

which holds a local state of subcriptions. However, we already have the `React.ref` to store such subs. And, components does not need to re-render if the list of subscriptions update as its sub's result will update the component anyway.

My thought now is that we just need to use `React.ref`:

```typescript
export function useSubscriptions() {
  const subsRef = React.useRef([] as Subscription[]);

  React.useEffect(() => () => subsRef.current.forEach((s: Subscription): void => s.unsubscribe()), []);

  const addSubscription = (sub: Subscription): void => {
    subsRef.current = subsRef.current.concat([sub]);
  };

  return React.useCallback(addSubscription, [subsRef]);
}
```

tthvo:
Fixes #528 

Cleaned up useSubscriptions hook to remove local states and reduce re-rendering.

### DIFF
--- a/src/app/utils/useSubscriptions.ts
+++ b/src/app/utils/useSubscriptions.ts
@@ -40,17 +40,12 @@ import { Subscription } from 'rxjs';
 
 export function useSubscriptions() {
   const subsRef = React.useRef([] as Subscription[]);
-  const [subscriptions, setSubscriptions] = React.useState(subsRef.current);
 
   React.useEffect(() => () => subsRef.current.forEach((s: Subscription): void => s.unsubscribe()), []);
 
   const addSubscription = (sub: Subscription): void => {
-    setSubscriptions((subs: Subscription[]): Subscription[] => {
-      const result = subs.concat([sub]);
-      subsRef.current = result;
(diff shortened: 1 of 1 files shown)


maxcao13 (approved):
Does seem to speed up the rendering process, very nice!

andrewazores (approved):

PR #531, opened 2022-10-03
fix(jest): fix mock systemtime setup

### LINKED ISSUE #530: [Task] Jest mock system timer is not properly set up
It seems like Jest mock system time was not properly set up as real current date was set instead of the mock date (Sep 14, 2022).

tthvo:
Fixes #530 

Properly set up Jest mock system timer and added callback to reset to mock date to real after `DatetimePicker` tests.

The issue is that `DateTimePicker` should be imported after systemtime mock to receive the mocks.

### DIFF
--- a/src/test/Recordings/Filters/DateTimePicker.test.tsx
+++ b/src/test/Recordings/Filters/DateTimePicker.test.tsx
@@ -36,21 +36,22 @@
  * SOFTWARE.
  */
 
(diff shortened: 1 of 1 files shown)


tthvo:
Sorry I didn't notice this requirements while testing in September :((

andrewazores (approved):

PR #533, opened 2022-10-03
feat(rules): rules can be enabled/disabled

### LINKED ISSUE #525: [Task] Rules can be enabled/disabled
Implement a UI for https://github.com/cryostatio/cryostat/issues/773

andrewazores:
Fixes #525

### DIFF
--- a/src/app/Rules/Rules.tsx
+++ b/src/app/Rules/Rules.tsx
@@ -45,6 +45,7 @@ import {
   CardHeaderMain,
   EmptyState,
   EmptyStateIcon,
+  Switch,
   Text,
   TextVariants,
   Title,
@@ -83,6 +84,7 @@ export interface Rule {
   name: string;
   description: string;
   matchExpression: string;
+  enabled: boolean;
   eventSpecifier: string;
   archivalPeriodSeconds: number;
   initialDelaySeconds: number;
@@ -106,6 +108,7 @@ export const Rules = () => {
   const [cleanRuleEnabled, setCleanRuleEnabled] = React.useState(true);
 
   const tableColumns = [
+    { title: 'Enabled', },
     {
       title: 'Name',
       transforms: [sortable],
@@ -231,6 +234,12 @@ export const Rules = () => {
     setIsUploadModalOpen(true);
   }, [setIsUploadModalOpen]);
 
+  const handleToggle = React.useCallback((rule: Rule, enabled: boolean): void => {
+    addSubscription(
+      context.api.updateRule({ ...rule, enabled }).subscribe(() => refreshRules())
+    );
+  }, [context, context.api, addSubscription, refreshRules]);
+
   const displayRules = React.useMemo(() => {
     const { index, direction } = sortBy;
     let sorted = [...rules];
@@ -241,6 +250,14 @@ export const Rules = () => {
       sorted = direction === SortByDirection.asc ? sorted : sorted.reverse();
     }
     return sorted.map((r: Rule) => [
+      <>
+        <Switch
+          label=''
+          aria-label={`${r.name} is enabled`}
+          isChecked={r.enabled}
+          onChange={state => handleToggle(r, state)}
+        />
+      </>,
       r.name,
       r.description,
       r.matchExpression,
@@ -251,13 +268,13 @@ export const Rules = () => {
       r.maxAgeSeconds,
       r.maxSizeBytes,
     ]);
-  }, [rules, sortBy]);
+  }, [rules, sortBy, handleToggle]);
 
   const handleDelete = React.useCallback(
     (rowData: IRowData, clean: boolean = true) => {
       addSubscription(
         context.api
-          .deleteRule(rowData[0], clean)
+          .deleteRule(rowData[1], clean)
           .pipe(first())
           .subscribe(() => {} /* do nothing - notification will handle updating state */)
       );
@@ -267,7 +284,7 @@ export const Rules = () => {
 
   const handleDownload = React.useCallback(
     (rowData: IRowData) => {
-      context.api.downloadRule(rowData[0]);
+      context.api.downloadRule(rowData[1]);
     },
     [context, context.api]
   );
--- a/src/app/Rules/Rules.tsx
+++ b/src/app/Rules/Rules.tsx
@@ -209,6 +209,21 @@ export const Rules = () => {
   }, [addSubscription, context, context.notificationChannel, setRules]);
 
   React.useEffect(() => {
+    addSubscription(
+      context.notificationChannel.messages(NotificationCategory.RuleUpdated).subscribe((msg) => {
+        setRules((old) => {
+          for (const r of old) {
+            if (r.name === msg.message.name) {
+              r.enabled = msg.message.enabled;
+            }
+          }
+          return [...old];
+        });
+      })
+    );
+  }, [addSubscription, context, context.notificationChannel, setRules]);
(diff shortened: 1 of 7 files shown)


tthvo (approved):
Look great to me!

lkonno (dismissed):
It looks good to me as well.

andrewazores:
Rebased and added a small commit that uses the WebSocket notification channel to update the rules table to reflect the rules' updated state, rather than refreshing the whole table with a GET query.

src/app/Rules/Rules.tsx
```diff
@@ -231,6 +234,28 @@ export const Rules = () => {
     setIsUploadModalOpen(true);
   }, [setIsUploadModalOpen]);
 
+  const handleToggle = React.useCallback(
+    (rule: Rule, enabled: boolean): void => {
+      addSubscription(context.api.updateRule({ ...rule, enabled }).subscribe());
+    },
+    [context, context.api, addSubscription, refreshRules]
```
tthvo, line 241: I think we can remove this deps `refreshRules` here?

src/app/Rules/Rules.tsx
```diff
@@ -241,6 +266,14 @@ export const Rules = () => {
       sorted = direction === SortByDirection.asc ? sorted : sorted.reverse();
     }
     return sorted.map((r: Rule) => [
+      <>
+        <Switch
+          label=""
+          aria-label={`${r.name} is enabled`}
```
tthvo, line 272: I think this aria-label should be given `enabled/disabled` based on the `r.enabled` field?

tthvo (commented):
Looks good :D Probably, we should a test to check the UI when receiving the new `RuleUpdated` notifications?

src/app/Rules/Rules.tsx
andrewazores: Hmm, maybe. The docs say this:

```
  /** Adds accessible text to the switch, and should describe the isChecked="true" state. When label is defined, aria-label should be set to the text string that is visible when isChecked is true. */
  'aria-label'?: string;
```

I'm not sure exactly how the `aria-label` ends up getting used but I figured it was something like this string is used as a key and the switch state is the corresponding value for the screen reader or other assistive tech.

tthvo: Right! Just reading the usage of the label here: https://www.patternfly.org/v4/components/switch/#without-label

Looks like its used just in `checked` state for accessibility.

andrewazores:
I'm having a hard time getting that test to work. Within the testing context it seems like the switch's styling isn't rendering how I expect it to, and in the DOM both the "on" and "off" states of the switch appear to be visible at the same time...

tthvo:
> I'm having a hard time getting that test to work. Within the testing context it seems like the switch's styling isn't rendering how I expect it to, and in the DOM both the "on" and "off" states of the switch appear to be visible at the same time...

Screenshot from 2022-10-04 13-16-01

Oh yess interesting! I guess we could find the span element with class `pf-m-on/off` and check `toBeVisible`?

maxcao13 (dismissed):
Feels and looks good!

andrewazores:
> > I'm having a hard time getting that test to work. Within the testing context it seems like the switch's styling isn't rendering how I expect it to, and in the DOM both the "on" and "off" states of the switch appear to be visible at the same time...
> 
> Screenshot from 2022-10-04 13-16-01
> 
> Oh yess interesting! I guess we could find the span element with class `pf-m-on/off` and check `toBeVisible`?

That's what I've been trying. I committed the current state of my test, but skipped so that it doesn't fail the build. At the end of `'update a rule when receiving a notification'` I have these queries and expectations:

```typescript
    const labels = container.querySelectorAll('label');
    expect(labels).toHaveLength(1);
    const label = labels[0];
    const spans = label.querySelectorAll(':scope > .pf-c-switch__label');
    expect(spans).toHaveLength(2);

    const onState = label.querySelector(':scope > .pf-c-switch__label.pf-m-on');
    expect(onState).toBeInTheDocument();
    expect(onState).not.toBeVisible();

    const offState = label.querySelector(':scope > .pf-c-switch__label.pf-m-off');
    expect(offState).toBeInTheDocument();
    expect(offState).toBeVisible();
```

but, both `onState` and `offState` are visible...

andrewazores:
I guess those elements are hidden by the Patternfly css applying `display: none` or `display: block` to them. I don't think that stylesheet gets loaded in these tests though, does it? So as far as the test renderer can tell these things both are actually visible. Not sure what to do about that.

tthvo:
> > > I'm having a hard time getting that test to work. Within the testing context it seems like the switch's styling isn't rendering how I expect it to, and in the DOM both the "on" and "off" states of the switch appear to be visible at the same time...
> > 
> > 
> > Screenshot from 2022-10-04 13-16-01
> > Oh yess interesting! I guess we could find the span element with class `pf-m-on/off` and check `toBeVisible`?
> 
> That's what I've been trying. I committed the current state of my test, but skipped so that it doesn't fail the build. At the end of `'update a rule when receiving a notification'` I have these queries and expectations:
> 
> ```ts
>     const labels = container.querySelectorAll('label');
>     expect(labels).toHaveLength(1);
>     const label = labels[0];
>     const spans = label.querySelectorAll(':scope > .pf-c-switch__label');
>     expect(spans).toHaveLength(2);
> 
>     const onState = label.querySelector(':scope > .pf-c-switch__label.pf-m-on');
>     expect(onState).toBeInTheDocument();
>     expect(onState).not.toBeVisible();
> 
>     const offState = label.querySelector(':scope > .pf-c-switch__label.pf-m-off');
>     expect(offState).toBeInTheDocument();
>     expect(offState).toBeVisible();
> ```
> 
> but, both `onState` and `offState` are visible...

Hmm, just looking back at the docs. How about we remove the label props to `Switch` (instead of setting it as "" - its treated as labeled so 2 spans are created)? This way we only have a single element within `<label>` to check?

andrewazores:
Heh. Yea, the reason I set the label explicitly to the empty string is that if it isn't specified then the component falls back to displaying the `aria-label` visually. I don't want that label appearing inside a table row, though. I could do that if I remove the separate `Name` table column and use the first column to display both the Switch and the rule's name, but then the rule deleting/updating logic is broken because we access the rules by the `rowData`, and now we would not have any index in that array that contains just the unique rule name that we need for API calls - it would be a Switch component with the name as part of a label instead.

tthvo:
Right! Make sense! But I think probably there is an icon in-place if label is not set https://www.patternfly.org/v4/components/switch/#without-label? 

Though, this would not solve the issue. I checked the element in `js-dom`, but they don't have `block: none` as in the real browsers. Do you think it would be some `js-dom` limitation when loading css?

andrewazores:
I think so. I don't think any CSS even gets loaded by our test setup. The real application hooks the css in in `index.tsx`, but the test harness just loads each component directly from its declaration and renders it, and there is no apparent link from that component definition to any kind of stylesheets at all.

tthvo:
Just went throught some js-dom issue and saw this: https://github.com/testing-library/jest-dom/issues/209#issuecomment-587827304. Could be the root issue?

andrewazores:
That issue sounds related and perhaps would be a blocker for any fix on our end where we do try to load the PF CSS.

Anyway, hacking it by adding a specific `className` sounds like a good enough fix for now.

tthvo (approved):
Looks good :D

PR #542, opened 2022-10-06
feat(recording): add form control for archiveOnStop

### LINKED ISSUE #541: [Task] Add "Archive on Stop" control to manual recording creation UI
Depends on cryostatio/cryostat#1100

andrewazores:
Fixes #541

### DIFF
--- a/src/app/CreateRecording/CustomRecordingForm.tsx
+++ b/src/app/CreateRecording/CustomRecordingForm.tsx
@@ -84,6 +84,7 @@ export const CustomRecordingForm = (props) => {
   );
   const [nameValid, setNameValid] = React.useState(ValidatedOptions.default);
   const [continuous, setContinuous] = React.useState(false);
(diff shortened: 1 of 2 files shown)


tthvo (dismissed):
Looks good!

lkonno:
- If  `Archive on Stop` is unchecked and I select the `Continuous` checkbox, the `Archive on Stop` option is disabled unchecked. It seems OK. 
- But when I select the `Archive on Stop` checkbox, the `Continuous` checkbox is still enabled. If I select the `Continuous` checkbox now, the `Archive on Stop` option is disabled, but it is still checked. 
I think that it can cause a wrong expectation to the user as the recording will not be archived on a manual stop. 
IMO the expected behavior would be to only allow one of the options to be selected at a time. It does not need to disable the checkbox, but when one option is selected, uncheck the other option.  

test1

lkonno (approved):
It looks good.

PR #543, opened 2022-10-06
fix(rules): show error view and modal when template retrieving fails

### LINKED ISSUE #485: Different HTTP error types [4xx vs. 5xx] may be handled differently by the browser or by rxjs error handling
Not sure why, but `Observables<EventTemplate[]>` that are subscribed from `refreshTemplateList` that run into `Error 427: Client Error (JMX Authentication)` never complete until a a different valid target is queried, then both are completed (can see from console). But Observables that run into `Error 502: Bad Gateway (Target SSL Untrusted)`are caught by the `catchError` that is implemented and hence get completed correctly.

_Originally posted by @maxcao13 in https://github.com/cryostatio/cryostat-web/pull/483#issuecomment-1190676076_

### LINKED ISSUE #527: [Task] JMX Auth Modal prompt does not appear on Automated Rules creation form
When viewing the Automated Rules creation form we query the backend for the event templates available for the selected target JVM (if any is selected). This opens a JMX connection and therefore may require JMX credentials to be passed or stored so that Cryostat can check the TARGET templates. A similar data flow occurs on the Recordings and Events views, and a modal dialog appears prompting the user to enter JMX credentials. On the Automated Rule creation view the connection and HTTP request simply fail if credentials are required and an error notification appears. The standard JMX credentials auth modal should appear instead and, just like for Recordings/Events, the request should automatically re-fire if credentials are saved so that the event template selector can populate.

tthvo:
Fixes #527 
Fixes #485 
Depends on https://github.com/cryostatio/cryostat/issues/1104

**Fixes**
- Added Auth form or SSL warning if failing to retrieve a template list from selected target.
- Editted ErrorView for better layout and visuals :D

**Chore**
- Use `useSubscriptions` for subs and `React.useCallback` for callbacks in `AppLayout`.

### DIFF
--- a/src/app/Rules/CreateRule.tsx
+++ b/src/app/Rules/CreateRule.tsx
@@ -58,7 +58,7 @@ import {
   ValidatedOptions,
 } from '@patternfly/react-core';
 import { useHistory, withRouter } from 'react-router-dom';
-import { catchError, filter, first, mergeMap, toArray } from 'rxjs/operators';
+import { filter, first, mergeMap, toArray } from 'rxjs/operators';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { NotificationsContext } from '@app/Notifications/Notifications';
 import { BreadcrumbPage, BreadcrumbTrail } from '@app/BreadcrumbPage/BreadcrumbPage';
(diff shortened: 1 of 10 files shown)


src/app/Rules/CreateRule.tsx
```diff
@@ -243,6 +211,14 @@ const Comp = () => {
     maxSizeUnits,
   ]);
 
+  const handleTemlateList = React.useCallback(
```
andrewazores, line 214: typo: "Temlate"

tthvo: oh sorry haha :D

andrewazores (approved):

tthvo:
Oh sorry I just added a new issue. I will fix it in a new PR :P

PR #540, opened 2022-10-05
feat(archive): display file size for archived recordings

tthvo:
Fixes #516
Fixes #491
Related to #532

**Features**
- Added recording size column for archived table.
- All recording tables are now scrollable (previously only All-archives view). Default height size is `500px`.

**Fixes**:
- Appended action menus to higher element - `document.body` in DOM tree (avoiding cut-off). Had to create manual dropdown as patternfly `ActionColumn` is not configurable for this issue. 
- Snapshot tests for recording tables are dropped as rendering requires DOM, and replaced with tests that checks if elements are visible as expected with `js-dom`.
- `ActiveRecording` and `ArchivedRecording` are now different types (i.e. no inheritance) with `Recording` as their representing Union type.

PR #544, opened 2022-10-06
feat(auth): add retry button on auth failure

### LINKED ISSUE #485: Different HTTP error types [4xx vs. 5xx] may be handled differently by the browser or by rxjs error handling

tthvo:
Fixes #485
Related to #540 

The issue is that the observable for `427` error code is hanged waiting for retry.

So, I added a `Retry` button for auth failure, which picks up the hanged `Observable` to retry. If the user does not click cancel, it will still be there until moving away to another page or reloads.

Also, just noticed the total size for archived recording is displayed when there are no recordings. This is removed now.

### DIFF
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -374,19 +374,17 @@ export const EventTemplates = () => {
     </>
   );
 
+  const authRetry = React.useCallback(() => {
+    context.target.setAuthRetry();
+  }, [context.target, context.target.setAuthRetry]);
+
   if (errorMessage != '') {
-    const isAuthError = React.useMemo(() => errorMessage === authFailMessage, [errorMessage, authFailMessage]);
-
-    const authRetry = React.useCallback(() => {
-      context.target.setAuthRetry();
-    }, [context.target, context.target.setAuthRetry]);
-
     return (
       <ErrorView
         message={
           <>
             <Text>{errorMessage}</Text>
-            {isAuthError && (
+            {errorMessage === authFailMessage && (
               <Button variant="link" onClick={authRetry}>
                 Retry
               </Button>
--- a/src/app/Events/EventTemplates.tsx
+++ b/src/app/Events/EventTemplates.tsx
@@ -54,6 +54,7 @@ import {
   ToolbarGroup,
   ToolbarItem,
   TextInput,
+  Text,
 } from '@patternfly/react-core';
 import { UploadIcon } from '@patternfly/react-icons';
 import {
@@ -71,7 +72,7 @@ import {
 import { useHistory } from 'react-router-dom';
 import { concatMap, filter, first } from 'rxjs/operators';
 import { LoadingView } from '@app/LoadingView/LoadingView';
-import { ErrorView } from '@app/ErrorView/ErrorView';
+import { authFailMessage, ErrorView } from '@app/ErrorView/ErrorView';
 import { DeleteWarningType } from '@app/Modal/DeleteWarningUtils';
 import { DeleteWarningModal } from '@app/Modal/DeleteWarningModal';
 
@@ -374,7 +375,27 @@ export const EventTemplates = () => {
   );
 
   if (errorMessage != '') {
-    return <ErrorView message={errorMessage} title={'Fail to retrieve event templates'} />;
+    const isAuthError = React.useMemo(() => errorMessage === authFailMessage, [errorMessage, authFailMessage]);
+
+    const authRetry = React.useCallback(() => {
+      context.target.setAuthRetry();
+    }, [context.target, context.target.setAuthRetry]);
+
+    return (
+      <ErrorView
+        message={
+          <>
+            <Text>{errorMessage}</Text>
+            {isAuthError && (
+              <Button variant="link" onClick={authRetry}>
+                Retry
+              </Button>
+            )}
+          </>
+        }
+        title={'Fail to retrieve event templates'}
+      />
+    );
   } else if (isLoading) {
     return (
       <>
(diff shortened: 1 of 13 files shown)


tthvo:
Also, I feel like the target selection menu should allow re-selection of the same target. This way, we avoid "surprise" from users that selecting does nothing (i.e. in case of target needing jmx credentials, previously open warning modal, but next time none happens). What do you think?

andrewazores:
Re-selection is currently ignored so that selecting the same target you're already looking at doesn't force the content view to refresh. It isn't a big deal either way but IMO it isn't surprising that re-selecting the same thing doesn't do anything, including not causing the dialog to re-appear after you choose to cancel it.

tthvo:
> Re-selection is currently ignored so that selecting the same target you're already looking at doesn't force the content view to refresh. It isn't a big deal either way but IMO it isn't surprising that re-selecting the same thing doesn't do anything, including not causing the dialog to re-appear after you choose to cancel it.

Oh rightt :D That make sense. I removed those changes :d

src/app/Rules/CreateRule.tsx
```diff
@@ -264,12 +265,28 @@ const Comp = () => {
...
+                <>
+                  <Text>{errorMessage}</Text>
+                  {errorMessage === authFailMessage && (
+                    <Button variant="link" onClick={authRetry}>
+                      Retry
+                    </Button>
+                  )}
+                </>
+              }
+              title={'Fail to retrieve event templates'}
```
andrewazores, line 288: This message snippet is mostly copy-pasted across different components. Can it be extracted as a reuseable component? Maybe extract the whole `ErrorView` into something like `ErrorRetryView`?

andrewazores:
Along with this change, maybe it's finally time to remove the refresh button on the `TargetSelect` widget? That button just re-queries the Cryostat backend for the targets list, but it should be queried at web-client startup and then updated by target discovery notifications afterward. The explicit view refresh buttons were removed from every other content view a long time ago, ever since the websocket notification channel got fleshed out, but that one stuck around partially because it did also work as a "ask me again about JMX auth" button - but this was never clear and doesn't even really make sense.

src/app/Rules/CreateRule.tsx
tthvo: Right, instead of a passing the entire message component, I refactor it to allow passing `retry` callback. If passed, the `ErrorView ` becomes `ErrorRetryView`.

tthvo:
> Along with this change, maybe it's finally time to remove the refresh button on the `TargetSelect` widget? That button just re-queries the Cryostat backend for the targets list, but it should be queried at web-client startup and then updated by target discovery notifications afterward. The explicit view refresh buttons were removed from every other content view a long time ago, ever since the websocket notification channel got fleshed out, but that one stuck around partially because it did also work as a "ask me again about JMX auth" button - but this was never clear and doesn't even really make sense.

Right! That makes sense! I removed the refresh button from the targetSelect now.

src/app/Events/EventTemplates.tsx
```diff
@@ -373,8 +374,12 @@ export const EventTemplates = () => {
     </>
   );
 
+  const authRetry = React.useCallback(() => {
+    context.target.setAuthRetry();
+  }, [context.target, context.target.setAuthRetry]);
+
   if (errorMessage != '') {
-    return <ErrorView message={errorMessage} title={'Fail to retrieve event templates'} />;
+    return <ErrorView title={'Fail to retrieve event templates'} message={errorMessage} retry={authRetry} />;
```
maxcao13, line 382: Maybe change to either "Failed to retrieve event templates" and change the other one to match with it, or change to "Error retrieving event templates" so that this one matches with the other error message.

src/app/Events/EventTypes.tsx
```diff
@@ -219,9 +221,13 @@ export const EventTypes = () => {
...
   };
 
+  const authRetry = React.useCallback(() => {
+    context.target.setAuthRetry();
+  }, [context.target, context.target.setAuthRetry]);
+
   // TODO replace table with data list so collapsed event options can be custom formatted
   if (errorMessage != '') {
-    return <ErrorView message={errorMessage} title={'Fail to retrieve event types'} />;
+    return <ErrorView title={'Fail to retrieve event types'} message={errorMessage} retry={authRetry} />;
```
maxcao13, line 230: Same idea here.

src/app/Rules/CreateRule.tsx
```diff
@@ -264,12 +265,16 @@ const Comp = () => {
...
+    context.target.setAuthRetry();
+  }, [context.target, context.target.setAuthRetry]);
+
   return (
     <BreadcrumbPage pageTitle="Create" breadcrumbs={breadcrumbs}>
       <Grid hasGutter>
         <GridItem xl={7}>
           {errorMessage ? (
-            <ErrorView message={errorMessage} title={'Fail to retrieve event templates'}></ErrorView>
+            <ErrorView title={'Fail to retrieve event templates'} message={errorMessage} retry={authRetry}></ErrorView>
```
maxcao13, line 277: Here as well.

src/app/Events/EventTemplates.tsx
tthvo: Good idea! I will fix that!

maxcao13 (approved):
Looks good, thanks for doing this!

andrewazores (approved):

PR #547, opened 2022-10-11
build(config): add build properties JSON file

### LINKED ISSUE #548: [Task] Add build properties file
This makes downstream build easier to configure web-client information.

andrewazores:
Fixes #548

Easy build-time overriding of distribution URLs for bug trackers, mailing list, product name, etc.

### DIFF
--- a/src/app/About/About.tsx
+++ b/src/app/About/About.tsx
@@ -37,6 +37,7 @@
  */
 
 import React from 'react';
+import build from '@app/build.json';
 import cryostatLogoHorizontal from '@app/assets/logo-cryostat-3-horizontal.svg';
 import { BreadcrumbPage } from '@app/BreadcrumbPage/BreadcrumbPage';
 import { AboutDescription, CRYOSTAT_TRADEMARK } from './AboutDescription';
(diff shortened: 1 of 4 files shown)


tthvo (approved):
Looks good! Worked for me :D I just file an issue and reference it here.

ebaron (commented):
Looks good! How easy would it be to customize the version link as well? For downstream builds I patch out the link since the commit hash does not refer to an upstream commit. Is it possible to make it a URL parameter in the JSON file and if empty/unset the `href` is omitted?

andrewazores:
> Looks good! How easy would it be to customize the version link as well? For downstream builds I patch out the link since the commit hash does not refer to an upstream commit. Is it possible to make it a URL parameter in the JSON file and if empty/unset the `href` is omitted?

If I understood correctly, the last commit makes this easy now. Just patch the `build.json` and set `"commitHashUrl": ""` and the text component will simply render the version string reported by the backend, with no linking to the upstream repo.

ebaron (approved):
Thanks for doing that as well, that's exactly what I had in mind.

PR #550, opened 2022-10-11
feat(archives): add non-target-specific "all-archives" view

### LINKED ISSUE #546: [Task] Add a view for users to find all archives without target-specificity
> Somewhat off-topic: While working on this I had the realization that in the important deployment scenarios, ex. k8s/OpenShift, the JMX Service URL is generally going to contain a service IP address rather than some resolvable DNS hostname. This means that the migration/transfer work we do is unlikely to be triggered to begin with, because a restarted instance may not appear to be "the same" target, as both its JMX Service URL and its JVM ID will have changed. In that kind of case I think it's OK that we don't associate the old archives to the new target instance, however, we DO need to provide the user a way to access those archives still. Just because the target has gone offline and wasn't replaced doesn't mean the user doesn't still need those archives. That's mainly a frontend problem, but we might need to tweak some of our queries to suit.

> Anyway, I think adding a query akin to the old-style `GET /api/v1/recordings` is what we need to ensure the user can find those recordings from old targets that we don't see any replacement for. IMO this should probably be the new underlying query for the frontend's All-Archives view. Rather than getting the list of known targets and then grabbing the corresponding archived recording information for those targets only, it needs to be essentially just a view of the archives and the subdirectories, with no consideration for which of those correspond to currently-known targets.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/1105#issuecomment-1272148324_

maxcao13:
Fixes #546
Related https://github.com/cryostatio/cryostat/pull/1110
Depends on https://github.com/cryostatio/cryostat/pull/1110

Must be tested with the above back-end PR to work.

### DIFF
--- a/src/app/Archives/AllArchivedRecordingsTable.tsx
+++ b/src/app/Archives/AllArchivedRecordingsTable.tsx
@@ -1,0 +1,546 @@
+/*
+ * Copyright The Cryostat Authors
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ * The above copyright notice and either this complete permission notice or at
+ * a minimum a reference to the UPL must be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import * as React from 'react';
+import { ArchivedRecording, UPLOADS_SUBDIRECTORY } from '@app/Shared/Services/Api.service';
+import { ServiceContext } from '@app/Shared/Services/Services';
+import { NotificationCategory } from '@app/Shared/Services/NotificationChannel.service';
+import { useSubscriptions } from '@app/utils/useSubscriptions';
+import {
+  Button,
+  Checkbox,
+  Drawer,
+  DrawerContent,
+  DrawerContentBody,
+  Toolbar,
+  ToolbarContent,
+  ToolbarGroup,
+  ToolbarItem,
+  Text,
+  TextVariants,
+  FlexItem,
+  Flex,
+  Chip,
+  Badge,
+} from '@patternfly/react-core';
+import { Tbody, Tr, Td, ExpandableRowContent, TableComposable } from '@patternfly/react-table';
+import { PlusIcon } from '@patternfly/react-icons';
+
+import { Observable, forkJoin, merge, combineLatest, of } from 'rxjs';
+import { concatMap, filter, first, map } from 'rxjs/operators';
+import { NO_TARGET, Target } from '@app/Shared/Services/Target.service';
+import { parseLabels } from '@app/RecordingMetadata/RecordingLabel';
+import { LabelCell } from '../RecordingMetadata/LabelCell';
+
+import { DeleteWarningModal } from '@app/Modal/DeleteWarningModal';
+import { DeleteWarningType } from '@app/Modal/DeleteWarningUtils';
+
+import { ArchiveUploadModal } from '@app/Archives/ArchiveUploadModal';
+import { useDispatch, useSelector } from 'react-redux';
+import { TargetRecordingFilters, UpdateFilterOptions } from '@app/Shared/Redux/RecordingFilterReducer';
+import {
+  addFilterIntent,
+  addTargetIntent,
+  deleteAllFiltersIntent,
+  deleteCategoryFiltersIntent,
+  deleteFilterIntent,
+} from '@app/Shared/Redux/RecordingFilterActions';
+import { RootState, StateDispatch } from '@app/Shared/Redux/ReduxStore';
+import { formatBytes, hashCode } from '@app/utils/utils';
+import { RecordingActions } from '@app/Recordings/RecordingActions';
+import { ReportFrame } from '@app/Recordings/ReportFrame';
+import { RecordingLabelsPanel } from '@app/Recordings/RecordingLabelsPanel';
+import { RecordingsTable } from '@app/Recordings/RecordingsTable';
+import { emptyArchivedRecordingFilters, filterRecordings, RecordingFilters, RecordingFiltersCategories } from '@app/Recordings/RecordingFilters';
+
+export interface AllArchivedRecordingsTableProps {
+  isUploadsTable: boolean;
+  isNestedTable: boolean;
+}
+
+export const AllArchivedRecordingsTable: React.FunctionComponent<AllArchivedRecordingsTableProps> = (props) => {
+  const context = React.useContext(ServiceContext);
+  const addSubscription = useSubscriptions();
+  const dispatch = useDispatch<StateDispatch>();
+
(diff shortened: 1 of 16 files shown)


maxcao13:
Thoughts before tests?

The new view works differently in that notifications refresh the entire page. This is done because the querying to the backend actually happens in the parent component `AllArchivedRecordingTable` which contains multiple nested `ArchivedRecordingsTable`s <- these normally update state by themselves but the new view has a entirely different set of api calls that don't depend on jvmIds and such so I would have to create a new ArchivedTableRecording component to properly handle notifications instead of just refreshing the page, which I haven't done. It is entirely possible to do, but I'm not sure if this the current state is okay enough since tomorrow is the deadline.

andrewazores:
Sorry, I won't have time to review this fully tonight, but I will get it done tomorrow before making the upstream branches.

The full-view refresh on notification sounds fine for now, I don't expect this to be a popularly-used view in most scenarios, so it's OK if we leave a performance optimization on the table for later.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -867,6 +919,7 @@ export interface RecordingAttributes {
 
 export interface Metadata {
   labels: Object;
+  targetId?: string;
```
tthvo, line 922: Probably remove this from metadata.

src/app/Recordings/ReportFrame.tsx
```diff
@@ -57,23 +59,24 @@ export const ReportFrame: React.FunctionComponent<ReportFrameProps> = React.memo
...
+              setReport(JSON.stringify(err));
+            }
           }
-        }
-      );
-    return () => sub.unsubscribe();
-  }, [context, context.reports, recording, isExpanded, setReport, props, props.isExpanded, props.recording]);
+        )
+    );
+  }, [addSubscription, context.reports, recording, isExpanded, setReport, props, props.isExpanded, props.recording]);
```
tthvo, line 79: I think we can remove `props.isExpanded` and `props.recording` from this list as they are the same as `recording` and `isExpanded`.

src/app/Recordings/ArchivedRecordingsTable.tsx
```diff
@@ -554,6 +579,16 @@ export const ArchivedRecordingsTable: React.FunctionComponent<ArchivedRecordings
     </Drawer>
   );
 };
+
+export interface RecordingRowProps {
+  key: string;
+  recording: any;
```
tthvo, line 585: This should be `ArchivedRecording` type. Also, on this issue, could you also change this field type to `Recording` instead? I must have missed it from the previous PR.

https://github.com/cryostatio/cryostat-web/blob/770ace96fd8654e1c56c4c2f84dc05b9d888d99c/src/app/Recordings/RecordingActions.tsx#L57

src/app/Recordings/ArchivedRecordingsTable.tsx
```diff
@@ -406,14 +421,21 @@ export const ArchivedRecordingsTable: React.FunctionComponent<ArchivedRecordings
           <RecordingActions
             recording={props.recording}
             index={props.index}
-            uploadFn={() => context.api.uploadArchivedRecordingToGrafana(props.sourceTarget, props.recording.name)}
+            uploadFn={
+              props.directory
+                ? () =>
+                    context.api.uploadArchivedRecordingToGrafanaFromPath(props.directory!!.jvmId, props.recording.name)
```
tthvo, line 427: Maybe only one `!` here?

src/app/Recordings/ArchivedRecordingsTable.tsx
```diff
@@ -320,18 +325,28 @@ export const ArchivedRecordingsTable: React.FunctionComponent<ArchivedRecordings
...
+          filteredRecordings.forEach((r: ArchivedRecording) => {
+            if (checkedIndices.includes(hashCode(r.name))) {
+              context.reports.delete(r);
+              tasks.push(context.api.deleteArchivedRecording(t.connectUrl, r.name).pipe(first()));
+            }
+          });
+        })
+      );
+    }
     addSubscription(forkJoin(tasks).subscribe());
```
tthvo, line 348: Should this call be after pushing tasks into the task array (i.e. inside the first addSubscriptions)?

src/app/Archives/AllArchivedRecordingsTable.tsx
```diff
@@ -0,0 +1,328 @@
...
+      context.notificationChannel.messages(NotificationCategory.ArchivedRecordingDeleted).subscribe((v) => {
+        refreshDirectoriesAndCounts();
+      })
+    );
+  }, [addSubscription, context.notificationChannel, refreshDirectoriesAndCounts]);
+
+  const toggleExpanded = React.useCallback(
+    (dir) => {
+      const idx = indexOfDirectory(expandedDirectories, dir);
+      setExpandedDirectories((expandedTargets) =>
```
tthvo, line 169: Should be named `expandedDirectory` for the parameter of the func :D

src/app/Archives/AllArchivedRecordingsTable.tsx
```diff
@@ -0,0 +1,328 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import * as React from 'react';
+import { ServiceContext } from '@app/Shared/Services/Services';
+import { includesTarget, indexOfTarget, isEqualTarget, Target } from '@app/Shared/Services/Target.service';
```
tthvo, line 40: Maybe clean up some imports here?

src/app/Archives/AllArchivedRecordingsTable.tsx
```diff
@@ -0,0 +1,328 @@
...
+    [expandedDirectories, setExpandedDirectories]
+  );
+
+  const includesDirectory = (arr: RecordingDirectory[], dir: RecordingDirectory): boolean => {
+    return arr.some((t) => t.connectUrl === dir.connectUrl);
+  };
+
+  const indexOfDirectory = (arr: RecordingDirectory[], dir: RecordingDirectory): number => {
+    let index = -1;
+    arr.forEach((t, idx) => {
```
tthvo, line 184: should use `dir` instead of `t` for clarity?

tthvo (commented):
Looks good overall to me. Just some questions :D

tthvo:
> Thoughts before tests?
> 
> The new view works differently in that notifications refresh the entire page. This is done because the querying to the backend actually happens in the parent component `AllArchivedRecordingsTable` which contains multiple nested `ArchivedRecordingsTable`s <- these normally update state by themselves but the new view has a entirely different set of api calls that don't depend on jvmIds and such so I would have to create a new `ArchivedRecordingsTable`-like component to properly handle notifications instead of just refreshing the page, which I haven't done. It is entirely possible to do, but I'm not sure if this the current state is okay enough since tomorrow is the deadline.

One way to go around this issue (i.e. similar to AllTargetArchive) is that the parent `AllArchive` can query only the list of directories and counts. Then, pass in an optional `DirectoryOptions` props including the directory info to the ArchiveTable. Then, the archive can check if this option is available, and make the api calls to fetch recordings for that directory. I think we should do the same for uploads. But this seems to require some refactoring and maybe backend supports :D

src/app/Recordings/ArchivedRecordingsTable.tsx
tthvo: something like this to make sure that the task list is populated before calling `forkJoin`? Because the current order looks like the task list is done async (possibly after `forkJoin` returns)?

```typescript
addSubscription(
  props.target.subscribe((t) => {
    const tasks: Observable<any>[] = [];
    filteredRecordings.forEach((r: ArchivedRecording) => {
      if (checkedIndices.includes(hashCode(r.name))) {
        context.reports.delete(r);
        tasks.push(context.api.deleteArchivedRecording(t.connectUrl, r.name).pipe(first()));
      }
    });
    addSubscription(forkJoin(tasks).subscribe());
  })
);
```

src/app/Recordings/ArchivedRecordingsTable.tsx
tthvo: Ohh yeh sorry I meant only change the type in `RecordingActions` to `Recording`.  And the type in props should be `ArchivedRecording`. Should've reference the link first.

src/app/Archives/AllArchivedRecordingsTable.tsx
```diff
@@ -0,0 +1,308 @@
...
+      context.notificationChannel.messages(NotificationCategory.RecordingMetadataUpdated).subscribe((v) => {
+        refreshDirectoriesAndCounts();
+      })
+    );
+  }, [addSubscription, context.notificationChannel, refreshDirectoriesAndCounts]);
+
+  React.useEffect(() => {
+    addSubscription(
+      context.notificationChannel.messages(NotificationCategory.ActiveRecordingSaved).subscribe((v) => {
+        refreshDirectoriesAndCounts();
```
tthvo, line 151: Probably, when we refactor this, we can avoid refreshing everything :D

andrewazores:
@tthvo @maxcao13 please file an issue (or issues) to track these planned improvements to this view so we can put them on the backlog to be addressed after the release branchpoint.

andrewazores:
Just had one suggestion for the way the rows are rendered on the new table. PR opened against this feature branch here:

https://github.com/maxcao13/cryostat-web/pull/1

andrewazores:
@tthvo thanks for taking the code review earlier, I've given it a brief look-over after your more thorough review and things look OK. It works well, too. Other than my suggestion PR to change how the rows are displayed I have no further requests.

andrewazores:
Oh, I didn't update the snapshot test along with that PR I made. Could you update that yourself @maxcao13 ? Good to go after that, I think.

tthvo:
> @tthvo @maxcao13 please file an issue (or issues) to track these planned improvements to this view so we can put them on the backlog to be addressed after the release branchpoint.

Right, I filed the issue for refactoring at #552 :D

tthvo (approved):
Looks great now! Thanks for making this features :D

andrewazores (approved):

PR #551, opened 2022-10-12
fix(archive): all-archive should properly display when refreshing

### LINKED ISSUE #549: [Task] Refreshing causes incorrect display of All-Archives view
> There is a bug, probably not caused by this PR (and probably front-end related haven't checked main yet)but basically if you go to the archives view and have a target with archived recordings, wait for a autoRefresh, and the target and its archived recordings disappear. 

_Originally posted by @maxcao13 in https://github.com/cryostatio/cryostat/pull/1105#pullrequestreview-1137726916_

tthvo:
Fixes #549 

**Fixes**

All Archive view now correctly refresh view when auto-refresh is enabled. The issue was that there is a call to `Array.includes`, which only checks object reference.

**Chore**

Counts of archives for each target should not be an saved in an array (i.e. sorting might be supported). Alternative was `Map<string, number>` with key is `target.connectUrl`. This way, subscriptions to notification channel are only called once in `React.useEffect` for each type.

### DIFF
--- a/src/app/Archives/AllTargetsArchivedRecordingsTable.tsx
+++ b/src/app/Archives/AllTargetsArchivedRecordingsTable.tsx
@@ -37,7 +37,7 @@
  */
 import * as React from 'react';
 import { ServiceContext } from '@app/Shared/Services/Services';
-import { Target } from '@app/Shared/Services/Target.service';
+import { includesTarget, indexOfTarget, isEqualTarget, Target } from '@app/Shared/Services/Target.service';
 import { NotificationCategory } from '@app/Shared/Services/NotificationChannel.service';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
 import {
@@ -56,10 +56,9 @@ import { TableComposable, Th, Thead, Tbody, Tr, Td, ExpandableRowContent } from 
 import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
 import { ArchivedRecordingsTable } from '@app/Recordings/ArchivedRecordingsTable';
 import { of } from 'rxjs';
-import { concatMap, map } from 'rxjs/operators';
+import { map } from 'rxjs/operators';
 import { TargetDiscoveryEvent } from '@app/Shared/Services/Targets.service';
 import { LoadingView } from '@app/LoadingView/LoadingView';
-import _ from 'lodash';
 
 export interface AllTargetsArchivedRecordingsTableProps {}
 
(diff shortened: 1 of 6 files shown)


src/app/Archives/Archives.tsx
```diff
@@ -58,14 +59,16 @@ export const uploadAsTarget: Target = {
...
   const context = React.useContext(ServiceContext);
+  const addSubscription = useSubscriptions();
   const [activeTab, setActiveTab] = React.useState(0);
   const [archiveEnabled, setArchiveEnabled] = React.useState(false);
 
   React.useEffect(() => {
-    const sub = context.api.isArchiveEnabled().subscribe(setArchiveEnabled);
-    return () => sub.unsubscribe();
+    addSubscription(context.api.isArchiveEnabled().subscribe(setArchiveEnabled));
   }, [context.api]);
```
maxcao13, line 72: Do we need addSubscription in deps?

maxcao13: and setArchiveEnabled?

src/app/Archives/AllTargetsArchivedRecordingsTable.tsx
```diff
@@ -170,18 +167,14 @@ export const AllTargetsArchivedRecordingsTable: React.FunctionComponent<AllTarge
...
-          return old.filter((o) => !_.isEqual(o, target));
+          return old.filter((t) => !isEqualTarget(t, target));
         });
-        setExpandedTargets((old) => old.filter((o) => !_.isEqual(o, target)));
+        setExpandedTargets((old) => old.filter((t) => !isEqualTarget(t, target)));
         setCounts((old) => {
-          let updated = [...old];
-          updated.splice(idx, 1);
-          return updated;
+          const newMap = new Map<string, number>(old);
```
maxcao13, line 175: I'm wondering if the map should be a `React.useState()` state instead of being generated every render. Is there something that prevents that from working?

maxcao13: Oh, I see the state is a map, but you are creating a new one and using the old one as an argument.

src/app/Archives/Archives.tsx
tthvo: Yeh definitely! Sorry I missed that!

src/app/Archives/AllTargetsArchivedRecordingsTable.tsx
tthvo: yeh I have to create new map since we have deps on hooks, which check if this map changes (i.e. obj ref).

maxcao13 (approved):
LGTM, bug is fixed.

andrewazores (approved):

PR #557, opened 2022-10-17
fix(recordings): fix recording filters from being cutoff in nested archived recordings tables

### LINKED ISSUE #556: [Task] Recording filters are cutoff in nested Archived recording tables
Filters are cutoff in the All-Targets and All-Archives view.

Screenshot from 2022-10-17 20-39-41

maxcao13:
Fixes #556 
Depends on #555

### DIFF
--- a/src/app/Recordings/Filters/LabelFilter.tsx
+++ b/src/app/Recordings/Filters/LabelFilter.tsx
@@ -44,7 +44,6 @@ import { parseLabels, RecordingLabel } from '@app/RecordingMetadata/RecordingLab
 export interface LabelFilterProps {
   recordings: Recording[];
   filteredLabels: string[];
-  menuAppendTo?: HTMLElement | (() => HTMLElement) | 'inline' | 'parent';
   onSubmit: (inputLabel: string) => void;
 }
 
@@ -83,7 +82,6 @@ export const LabelFilter: React.FunctionComponent<LabelFilterProps> = (props) =>
       aria-label="Filter by label"
       typeAheadAriaLabel="Filter by label..."
       placeholderText="Filter by label..."
-      menuAppendTo={props.menuAppendTo}
     >
       {labels.map((option, index) => (
         <SelectOption key={index} value={option}>
--- a/src/app/Recordings/Filters/LabelFilter.tsx
+++ b/src/app/Recordings/Filters/LabelFilter.tsx
@@ -82,6 +82,7 @@ export const LabelFilter: React.FunctionComponent<LabelFilterProps> = (props) =>
       aria-label="Filter by label"
       typeAheadAriaLabel="Filter by label..."
       placeholderText="Filter by label..."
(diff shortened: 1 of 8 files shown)


maxcao13:
Currently investigating how to fix the tests since I added a prop `menuAppendTo` which appends the select dropdowns to the `"parent"` instead of `"inline"` to the Select component inside the `[]Filter` components. But the test renderer for the snapshot tests doesn't seem to have a parent.

tthvo:
Nicee catch! Could you also change the `menuAppendTo` of `DatetimePicker` to `parent` too? Look like some refactors from previous commits now allow `parent` instead of `document.body` (not sure which change makes this possible)?

https://github.com/cryostatio/cryostat-web/blob/b201606d565dcc10e23391506b28ab9fb7ee1982/src/app/Recordings/Filters/DateTimePicker.tsx#L83

This way we can avoid accessibility issue: https://www.patternfly.org/v4/components/select/#appending-document-body-vs-parent

tthvo:
Also, on this topic, this issue seems to be caused by the table having `sticky header`. Should we make all recording tables "sticky" too?

https://github.com/cryostatio/cryostat-web/blob/b201606d565dcc10e23391506b28ab9fb7ee1982/src/app/Recordings/RecordingsTable.tsx#L125

Then, all filters must be checked to avoid being cut-off.

src/app/Recordings/Filters/LabelFilter.tsx
```diff
@@ -82,6 +83,7 @@ export const LabelFilter: React.FunctionComponent<LabelFilterProps> = (props) =>
       aria-label="Filter by label"
       typeAheadAriaLabel="Filter by label..."
       placeholderText="Filter by label..."
+      menuAppendTo={props.menuAppendTo}
```
tthvo, line 86: Should probably add the same props field for other filter components?

tthvo (commented):
Looks good! Just another notice about the filter selects:

Screenshot from 2022-10-18 03-23-44

Looks like the select menu is not scrollable and could go as long as it needs. I think we should allow scrolling. This could be done by setting `maxHeight` on Select:

```typescript
<Select maxHeight={"300px"}>
  {options}
</Select>
```

What do you think?

andrewazores:
If that `maxHeight` prop supports general CSS size units then it would be better to use a relative size (like `em`) rather than absolute (like `px`) so that this better supports displays with different pixel densities.

andrewazores:
While you're at that, please update these two `px` uses as well:

```rg
src/app/SecurityPanel/Credentials/MatchedTargetsTable.tsx
126:        <InnerScrollContainer style={{ height: '300px' }}>

src/app/app.css
81:  height: 512px; 
```

tthvo (approved):
Looks nicee! Fixed the bug :D

andrewazores (approved):

PR #558, opened 2022-10-18
feat(jmc-agent): add JMC Agent probes UI

Josh-Matsuoka:
This PR implements the frontend of the Agent Plugin, adding a page with a table displaying currently available probe templates and an option to upload more and insert them into the target, as well as a page displaying any currently active probes with options to deactivate them.

Currently resolving a problem with the tests, likely an issue with how I've set up the mocks since the mocked probe templates don't seem to be making it to the page (i.e. it's getting stuck at the loading screen in tests), will keep this in draft for the moment until that's resolved.

Depends on https://github.com/cryostatio/cryostat/pull/1126


andrewazores:
@maxcao13 or @tthvo if you have time, could you take a look at this and help Josh figure out the tests?

andrewazores:
I think the yarn update should be backed out from this patch set.

maxcao13:
I think the problem is that in `handleTemplates` on line 94 in the `AgentProbeTemplates`, JSON.parse is called but the subscription value that gets passed to the function is type of array (i.e. context.api.getProbeTemplates returns `Observable<ProbeTemplate>[]` , and for some reason it doesn't log an error. I confirmed this by putting print statements before and after the parse calling. And then the error propagates and `setIsLoading` is never called to set the state to false ever again when a notification comes.

tthvo:
Right, just also checked to confirm Max's solution should solve the loading error. Other tests failures:

- `Tests involving template deletion`: this checks if warning modal is open. However, deletion warning modal is not set up for this agent view, so we need to set up before checks.
- `Tests involving notifcations`: 
  - There should be only 2 mock returns for each tests and order matters. 
  - For create notification, there is an extra third mock call. 
  - For delete notifcation, the order is inverted (i.e. delete comes seconds). 
  - Currently, `jest.spyOn(defaultServices.api, 'getProbeTemplates')` only mock returns one template. This should adjust based on tests (i.e. create notifcation test should mock this method return 2 templates).
- `Others`: The search texts should be corrected to `someProbeTemplate` as in mocks. 

Hope it solved the issue :D

Josh-Matsuoka:
Managed to get the tests fixed, thanks for the help :) This is ready for review now

package.json
```diff
@@ -86,7 +86,7 @@
     "webpack-cli": "^4.9.2",
     "webpack-dev-server": "^4.9.2",
     "webpack-merge": "^5.8.0",
-    "yarn": "^1.22.13"
+    "yarn": "^1.22.18"
```
andrewazores, line 89: This change still needs backing out I think. It shouldn't be needed here, and undoing this should also fix the CI run.

andrewazores:
Following the setup steps and example probe template from here: https://github.com/cryostatio/cryostat/pull/1126#issuecomment-1285681149

The workflow seems to work as anticipated, but the notification that appears when inserting probes has `undefined` in it:

image

And also, the deletion prompt for a probe template is titled as "Event Template":

image

I think just these few strings need to be updated and then this looks OK. I would also like to rename the main menu item from "Agent" to something like "JMC Agent", or perhaps even move this view content under the "Event" main menu item rather than having it as a top-level view. I can imagine that a main menu item named "Agent" would be useful for when we have a Cryostat Agent as well.

andrewazores:
Here's my concept for just moving the new JMC agent-related views under the existing Events nav item:

image

I just removed the top-level nav item and moved the `Card` component inside `Events`, wrapping the now two cards together in `Stack` and `StackItem` components.

Josh-Matsuoka:
I've pushed some changes to relabel the menu item to JMC Agent, if you'd prefer to have it nested under events I can look into doing that as well. I've also fixed the notifications and the deletion warning

andrewazores:
Finally, this is what it looks like when you connect to a target that doesn't have the JMC Agent loaded and visit these new views (whether they're under Events or JMC Agent menu items):

image

The server response code should probably not be `500` here but rather one of `501`, `502`, or `503` - each seems justifiable in different ways. Then the frontend encounters a `501`/`502`/`503` on this request then rather than displaying an error notification it should use some empty content placeholder view, probably with a message explaining that the JMC Agent was not found within the selected target application.

Josh-Matsuoka:
Updated to add hints to the error page when probes fail to load, as well as moving the agent items to the events page and adding a description card outlining what it's for. I changed the TargetProbesGet handler to return a 503 when retrieving event probes fails.

As a TODO I'll look into improving the error reporting from the jmc agent itself when I get some time, it would make our lives easier determining causes and giving better guidance on error views. For now I've gone with what you suggested with a generic error view and giving a hint that the agent may not be running.

andrewazores:
Nice, this is already looking much better. Could the helper text you added be incorporated into the body of the new card itself? Or maybe it can be a separate card, if it only displays when the JMC Agent views card is expanded? I like that the cards "collapse" when the active selected tab is on a different card:

image

but I think in this case, the JMC Agent helper text should not be visible, since it's not relevant to the views I can actually see.

I notice that no notification seems to appear when I delete a probe template. The UI does update to reflect it being removed and I see in the server logs that the notification should have been emitted, though.

After this last round of comments I am all good with the UI design.

andrewazores:
There seem to be duplicate notifications.

After inserting probes:

image


After removing probes:

image

Josh-Matsuoka:
Removed the duplicate notification

Josh-Matsuoka:
The format check is passing for me locally with no problems, is the one done by the CI different?

andrewazores:
The CI one is "npm run format:check"

andrewazores:
@Josh-Matsuoka @maxcao13 @tthvo thoughts on the latest changes I made?

Josh-Matsuoka:
Looks good to me, only displaying it when it's supported makes sense

src/app/Shared/Services/Api.service.tsx
```diff
@@ -476,6 +476,81 @@ export class ApiService {
...
+          'v2',
+          `targets/${encodeURIComponent(target.connectUrl)}/probes/${encodeURIComponent(templateName)}`,
+          {
+            method: 'POST',
+          }
+        ).pipe(
+          tap((resp) => {
+            if (resp.status == 400) {
+              this.notifications.warning(
+                'Failed to Insert Probes',
```
tthvo, line 510: Maybe `Failed to insert Probes` to match failed remove message above?

src/app/Shared/Services/Api.service.tsx
```diff
@@ -476,6 +476,81 @@ export class ApiService {
...
+          first()
+        )
+      )
+    );
+  }
+
+  addCustomProbeTemplate(file: File): Observable<boolean> {
+    const body = new window.FormData();
+    body.append('probeTemplate', file);
+    return this.sendRequest('v2', `probes/` + file.name, {
```
tthvo, line 525: Not a web issue but just a question for this endpoint: Any reason that backend requires the file name to be in path?

src/app/Modal/DeleteWarningUtils.tsx
```diff
@@ -84,6 +85,14 @@ export const DeleteEventTemplates: DeleteWarning = {
   ariaLabel: 'Event template delete warning',
 };
 
+export const DeleteProbeTemplates: DeleteWarning = {
+  id: DeleteWarningType.DeleteProbeTemplates,
+  title: 'Permanently delete Probe Template?',
+  label: 'Delete Probe Template',
+  description: `Custom event template data will be lost.`,
```
tthvo, line 92: -> `Custom probe template data will be lost.` instead?

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -0,0 +1,442 @@
...
+    );
+  }, [addSubscription, context, context.notificationChannel, setTemplates]);
+
+  React.useEffect(() => {
+    addSubscription(
+      context.notificationChannel.messages(NotificationCategory.TemplateDeleted).subscribe((v) => refreshTemplates())
+    );
+  }, [addSubscription, context, context.notificationChannel, setTemplates]);
+
+  const handleUploadCancel = () => {
```
maxcao13, line 291: I think, from past experience, many of these functions can be React.useCallback functions like 
```
const handleUploadCancel = React.useCallback() => {
  setUploadFile(undefined);
  setUploadFilename('');
  setModalOpen(false);
}, [setUploadFile, setUploadFilename, setModalOpen])
```
with the correct dependencies array and such.

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -330,7 +330,10 @@ export const AgentProbeTemplates = () => {
     return (
       <ErrorView
         title={'Error retrieving probe templates'}
-        message={errorMessage}
+        message={
+          'This is commonly caused by the agent not being loaded/active, check that the target was started with the agent (-javaagent:/path/to/agent.jar) ' +
```
maxcao13, line 334: Would start a new sentence with `This is commonly caused by the agent not being loaded/active. Check that...`

maxcao13 (dismissed):

src/app/Agent/AgentProbeTemplates.tsx
maxcao13: Whoops i'm not meaning to approve.

src/app/Events/Events.tsx
```diff
@@ -36,33 +36,80 @@
...
+                  </Tab>
+                </Tabs>
+              </CardBody>
+            </Card>
+          </StackItem>
+          {enabled ? (
+            <StackItem>
+              <Card>
+                <CardBody>
+                  <Tabs activeKey={activeTab} onSelect={handleTabSelect}>
```
tthvo, line 100: We might need to separate active-tab states here. Right now, a single activeTab states control 2 cards.

tthvo: This will unexpectedly close the open tab in other card.

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+  const authRetry = React.useCallback(() => {
+    context.target.setAuthRetry();
+  }, [context.target, context.target.setAuthRetry]);
+
+  if (errorMessage != '') {
+    return (
+      <ErrorView
+        title={'Error retrieving events'}
+        message={
+          'This is commonly caused by the agent not being loaded/active, check that the target was started with the agent (-javaagent:/path/to/agent.jar) ' +
```
tthvo, line 211: Maybe a period or some other options instead of blank space.

Screenshot from 2022-10-21 16-33-13

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+          </CardHeader>
+          <CardBody>
+            The JMC Agent allows users to dynamically inject custom JFR events into running JVMs. In order to make use
+            of the JMC Agent, the agent jar must be present in the same container as the target, and the target must be
+            started with the agent (-javaagent:/path/to/agent.jar). Once these pre-requisites are met the user can
+            upload probe templates to Cryostat and insert/remove them from targets, as well as view currently active
+            probes.
+          </CardBody>
+        </Card>
+        <Toolbar id="event-templates-toolbar">
```
tthvo, line 236: should have a different id? `event-templates-toolbar` is already in use.

https://github.com/cryostatio/cryostat-web/blob/e41e2c8cb1d8bacb004721e10b2d8e5cf84603e7/src/app/Events/EventTemplates.tsx#L365

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -0,0 +1,442 @@
...
+        }
+        retry={isAuthFail(errorMessage) ? authRetry : undefined}
+      />
+    );
+  } else if (isLoading) {
+    return <LoadingView />;
+  } else {
+    return (
+      <>
+        <Card>
```
tthvo, line 345: Looks like the card lower part is being covered. Same for live probe view. Maybe we could add some spacing between helper text and table.

Screenshot from 2022-10-21 16-41-55

tthvo: Look like the card needs some sort of shadow or styling so it needs more spacing below.

src/app/Events/Events.tsx
```diff
@@ -36,33 +36,80 @@
...
-  const handleTabSelect = (evt, idx) => {
-    setActiveTab(idx);
-  };
+  React.useEffect(() => {
+    addSubscription(
+      context.target
+        .target()
+        .pipe(
+          filter((t) => t !== NO_TARGET),
+          concatMap((_) => context.api.getActiveProbes(true))
```
tthvo, line 62: `context.api.getActiveProbes` internally calls `context.target.target()`. I think we can remove it here.

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+                  placeholder="Filter..."
+                  aria-label="Probe template filter"
+                  onChange={setFilterText}
+                />
+              </ToolbarItem>
+            </ToolbarGroup>
+            <ToolbarGroup variant="icon-button-group">
+              <ToolbarItem>
+                <Button key="create" variant="primary" onClick={handleModalToggle}>
+                  Remove Probes
```
tthvo, line 253: I think we should color it as red. Does this button remove all probes? Then, it should be renamed to `Remove All Probes`?

Screenshot from 2022-10-21 17-27-15

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -0,0 +1,442 @@
...
+  const [sortBy, setSortBy] = React.useState({} as ISortBy);
+  const [isLoading, setIsLoading] = React.useState(false);
+  const [errorMessage, setErrorMessage] = React.useState('');
+  const [rowDeleteData, setRowDeleteData] = React.useState({} as IRowData);
+  const [warningModalOpen, setWarningModalOpen] = React.useState(false);
+  const addSubscription = useSubscriptions();
+
+  const tableColumns = React.useMemo(
+    () => [
+      { title: 'name', transforms: [sortable] },
```
tthvo, line 105: Title should be capitalized for consistency with other tables.

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+  const refreshTemplates = React.useCallback(() => {
+    setIsLoading(true);
+    addSubscription(
+      context.target
+        .target()
+        .pipe(
+          filter((target) => target !== NO_TARGET),
+          first(),
+          concatMap((target) => context.api.getActiveProbes())
```
tthvo, line 142: We can remove call to `context.target.target` here as `getActiveProbes` already calls it internally.

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+    }
+    const id = window.setInterval(
+      () => refreshTemplates(),
+      context.settings.autoRefreshPeriod() * context.settings.autoRefreshUnits()
+    );
+    return () => window.clearInterval(id);
+  }, []);
+
+  React.useEffect(() => {
+    const sub = context.target.authFailure().subscribe(() => {
```
tthvo, line 172: Wrap in `addSubscription`?

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+  }, [addSubscription, context, context.target, context.api, setIsLoading, handleTemplates, handleError]);
+
+  React.useEffect(() => {
+    addSubscription(
+      context.target.target().subscribe(() => {
+        setFilterText('');
+        refreshTemplates();
+      })
+    );
+  }, [context, context.target, addSubscription, refreshTemplates]);
```
tthvo, line 158: Missing `setFilterText`?

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+    });
+    return () => sub.unsubscribe();
+  }, [context.target]);
+
+  const displayTemplates = React.useMemo(
+    () => templates.map((t: EventProbe) => [t.id, t.name, t.clazz, t.description, t.methodName + t.methodDescriptor]),
+    [templates]
+  );
+
+  const handleModalToggle = () => {
```
tthvo, line 183: Should be `handleDeleteProbes` since there is no warning modal here?

tthvo: Or perhaps we should add the warning modal instead?

tthvo: Should also be wrapped in `React.useCallback`?

src/app/Shared/Services/Api.service.tsx
andrewazores: That endpoint uses the filename as the resource name.

```xml
<?xml version="1.0"?>
<jfragent>
  <!-- Global configuration options -->
  <config>
    <!-- This is the prefix to use when generating event class names. -->
    <classprefix>__JFREvent</classprefix>
    <!-- Will allow the recording of arrays and object parameters as Strings. This will cause toString to be called for array elements and objects other than strings, which in turn can cause trouble if the toString method is badly implemented. Use with care. -->
    <allowtostring>true</allowtostring>
    <!-- Allows converters to be used. See the org.openjdk.jmc.agent.converters package for more information. -->
    <allowconverter>true</allowconverter>
  </config>
  <events>
    <event id="demo.jfr.test1">
      <label>DemoEvent</label>
      <description>Event for the agent plugin demo</description>
      <path>demo</path>
      <stacktrace>true</stacktrace>
      <class>io.cryostat.net.web.http.api.v1.TargetTemplatesGetHandler</class>
      <method>
        <name>handleAuthenticated</name>
        <descriptor>(Lio/vertx/ext/web/RoutingContext;)V</descriptor>
      </method>
      <!-- location {ENTRY, EXIT, WRAP}-->
      <location>ENTRY</location>
    </event>
  </events>
</jfragent>
```

This is an example of what the XML resource looks like - it doesn't have a meaningful name element to extract, or anything like that. A future Cryostat bcakend version should probably register these in a database and assign them numeric IDs instead.

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+    }
+    const { index, direction } = sortBy;
+    if (typeof index === 'number') {
+      const keys = ['ID', 'Label', 'Description', 'Class', 'Method'];
+      const key = keys[index];
+      const sorted = filtered.sort((a, b) => (a[key] < b[key] ? -1 : a[key] > b[key] ? 1 : 0));
+      filtered = direction === SortByDirection.asc ? sorted : sorted.reverse();
+    }
+    setFilteredTemplates([...filtered]);
+  }, [filterText, templates, sortBy]);
```
tthvo, line 115: Missing deps on `setFilteredTemplates`.

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+            <ToolbarGroup variant="icon-button-group">
+              <ToolbarItem>
+                <Button key="create" variant="primary" onClick={handleModalToggle}>
+                  Remove Probes
+                </Button>
+              </ToolbarItem>
+            </ToolbarGroup>
+          </ToolbarContent>
+        </Toolbar>
+        <Table aria-label="Active Events" variant={TableVariant.compact} cells={tableColumns} rows={displayTemplates}>
```
tthvo, line 259: Table should have specify `onSort={handleSort}` to set the current `sortBy`. Right now, toggling sort does not work.

andrewazores:
Feedback so far all makes sense to me. I will leave it to the three of you to determine who performs the work and commits the fixes.

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+          t.name.toLowerCase().includes(ft) ||
+          t.description.toLowerCase().includes(ft) ||
+          t.clazz.toLowerCase().includes(ft) ||
+          t.methodDescriptor.toLowerCase().includes(ft) ||
+          t.methodName.toLowerCase().includes(ft)
+      );
+    }
+    const { index, direction } = sortBy;
+    if (typeof index === 'number') {
+      const keys = ['ID', 'Label', 'Description', 'Class', 'Method'];
```
tthvo, line 109: Keys should match case of the object fields.

https://github.com/Josh-Matsuoka/cryostat-web/blob/a74db433295f7edb2c310486b484e03e41954c0a/src/app/Shared/Services/Api.service.tsx#L1058

Looks like we are using `Label` column for name. Should be change to just `Name`?

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -0,0 +1,442 @@
...
+    });
+  };
+
+  const handleFileChange = (value, filename) => {
+    setFileRejected(false);
+    setUploadFile(value);
+    setUploadFilename(filename);
+  };
+
+  const handleUploadSubmit = () => {
```
tthvo, line 255: Same suggestion from Max's comment https://github.com/cryostatio/cryostat-web/pull/558/files#r1002156730

tthvo: I think this should apply to all callbacks defined here.

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -0,0 +1,442 @@
...
+import { concatMap, filter, first } from 'rxjs/operators';
+import { LoadingView } from '@app/LoadingView/LoadingView';
+import { authFailMessage, ErrorView, isAuthFail } from '@app/ErrorView/ErrorView';
+import { ProbeTemplate } from '@app/Shared/Services/Api.service';
+import { DeleteWarningType } from '@app/Modal/DeleteWarningUtils';
+import { DeleteWarningModal } from '@app/Modal/DeleteWarningModal';
+
+export const AgentProbeTemplates = () => {
+  const context = React.useContext(ServiceContext);
+  const history = useHistory();
```
tthvo, line 86: history not used :D

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -0,0 +1,442 @@
...
+    [setIsLoading, setErrorMessage]
+  );
+
+  const refreshTemplates = React.useCallback(() => {
+    setIsLoading(true);
+    addSubscription(
+      context.target
+        .target()
+        .pipe(
+          concatMap((target) => context.api.getProbeTemplates()),
```
tthvo, line 154: Seems like api call is not specific to a target. Should we remove `context.target.target`?

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -0,0 +1,442 @@
...
+      return;
+    }
+    const id = window.setInterval(
+      () => refreshTemplates(),
+      context.settings.autoRefreshPeriod() * context.settings.autoRefreshUnits()
+    );
+    return () => window.clearInterval(id);
+  }, []);
+
+  React.useEffect(() => {
```
tthvo, line 184: wrap in `addSubscription`?

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -0,0 +1,442 @@
...
+      context.api
+        .addCustomProbeTemplate(uploadFile)
+        .pipe(first())
+        .subscribe((success) => {
+          setUploading(false);
+          if (success) {
+            setUploadFile(undefined);
+            setUploadFilename('');
+            setModalOpen(false);
+            refreshTemplates();
```
tthvo, line 271: We don't need to refresh here as notification will trigger it?

src/test/Agent/Agent.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+      tree = renderer.create(
+        <ServiceContext.Provider value={defaultServices}>
+          <AgentProbeTemplates />
+        </ServiceContext.Provider>
+      );
+    });
+    expect(tree.toJSON()).toMatchSnapshot();
+  });
+
+  it('adds a recording after receiving a notification', () => {
```
tthvo, line 138: Test name should be from using `recording` to `probe templates`.

src/test/Agent/Agent.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'),
+  useRouteMatch: () => ({ url: '/baseUrl' }),
+  useHistory: () => ({
+    push: mockHistoryPush,
+  }),
+}));
+
+jest.spyOn(defaultServices.settings, 'deletionDialogsEnabledFor').mockReturnValue(false); // don't ask again
```
tthvo, line 95: Should have tests to check if deletion warning modal is show correctly :D

tthvo (commented):
> Feedback so far all makes sense to me. I will leave it to the three of you to determine who performs the work and commits the fixes.

Just couple suggestions. I think Max and I could help cover these React-related changes.

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,266 @@
...
+import { LoadingView } from '@app/LoadingView/LoadingView';
+import { ErrorView, isAuthFail } from '@app/ErrorView/ErrorView';
+import { EventProbe } from '@app/Shared/Services/Api.service';
+
+export const AgentLiveProbes = () => {
+  const context = React.useContext(ServiceContext);
+  const history = useHistory();
+
+  const [templates, setTemplates] = React.useState([] as EventProbe[]);
+  const [filteredTemplates, setFilteredTemplates] = React.useState([] as EventProbe[]);
```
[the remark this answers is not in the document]
tthvo: Should be used to display templates instead of raw `template` state.

src/app/Shared/Services/Api.service.tsx
tthvo: Right! That makes sense! Thanks for explaining :D

tthvo:
Looks like when you upload a probe template with the same file name,virtually it seems a bit unsure whether it is successfully accepted or not. Though, after reloading the page, the new duplicate template is not shown, so I guess it is not added. Should this return some more descriptive message/status code (i.e. 400)?

Screenshot from 2022-10-23 23-14-45

Cryostat's logs:

```log
SEVERE: HTTP 500: new_file.xml
io.vertx.ext.web.handler.HttpException: Internal Server Error
Caused by: io.cryostat.net.web.http.api.v2.ApiException: new_file.xml
	at io.cryostat.net.web.http.api.v2.ProbeTemplateUploadHandler.handle(ProbeTemplateUploadHandler.java:151)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:102)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:68)
	at io.vertx.ext.web.impl.BlockingHandlerDecorator.lambda$handle$0(BlockingHandlerDecorator.java:48)
	at io.vertx.core.impl.ContextImpl.lambda$null$0(ContextImpl.java:159)
	at io.vertx.core.impl.AbstractContext.dispatch(AbstractContext.java:100)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$1(ContextImpl.java:157)
	at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:76)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.nio.file.FileAlreadyExistsException: new_file.xml
	at io.cryostat.core.agent.LocalProbeTemplateService.addTemplate(LocalProbeTemplateService.java:98)
	at io.cryostat.net.web.http.api.v2.ProbeTemplateUploadHandler.handle(ProbeTemplateUploadHandler.java:141)
	... 11 more
```

src/app/Events/Events.tsx
tthvo: Edit: This is needed to listen for target change. Added back.

tthvo:
Ready for review now... Addressed all comments above and fixed tests. Just a question about uploading duplicate file name above (perhaps its some backend issue).

As for `AgentProbeTemplates`, I removed the listener for `authFailure` because this table is not target-specific resources. This also helps avoid the case where changing a target that results in `authFailture` does not render the error view in the probe template table but does render after clicking Retry on `EventTemplates` table.

andrewazores:
> Should this return some more descriptive message/status code (i.e. 400)?

Agreed, the backend should respond with `400` or maybe `409` in this case, and should not emit a WS notification.

andrewazores:
One minor bug. In the following scenarios I will use `9093` and `9097` to mean targets without the JMC Agent attached, and `9091` to mean Cryostat itself that does have the Agent attached. Also, for all scenarios, selecting the target from within Events or selecting it elsewhere and then navigating into Events doesn't matter unless otherwise described.

1. Go to Events and choose `9093`. JMC Agent card does not appear.
2. Go to Events and choose `9091`. JMC Agent card does appear.
3. Go to Events and choose `9093`. JMC Agent card does not appear. Choose `9097`. Card does not appear.
4. Go to Events and choose `9093`. Agent card does not appear. Choose`9091` and card does appear.

One slightly broken scenario:

5. Go to Events and choose `9091`. Agent card does appear. Choose `9093`. Card disappears, but a `501` error notification also appears. The notification should not appear.

Josh-Matsuoka:
I'll take a look and see whats going on with the scenario that Andrew just mentioned.

Thanks for fixing all of those things Thuan

andrewazores:
"Remove All Probes" shows a notification like this:

image

I think it should be reworded to talk about probes being removed rather than template being deleted, and the `undefined` should be either removed or replaced with something more meaningful.

Josh-Matsuoka:
the suppress notifications flag got missed in the refactor of refreshTemplates, I've readded it.

r.e. the duplicate filename issue, I think it makes sense to adjust the backend and add an overwrite flag to the api request to determine the behavior the user wants, then we can adjust the frontend accordingly. I'll make an issue for the backend

tthvo:
> "Remove All Probes" shows a notification like this:
> 
> image
> 
> I think it should be reworded to talk about probes being removed rather than template being deleted, and the `undefined` should be either removed or replaced with something more meaningful.

Also, look like we are not currently listening for this deletion event here. Trying to delete a probe from a second browser session does not update the first one. Maybe after adding a new notification category, we can start to listen.

andrewazores:
> r.e. the duplicate filename issue, I think it makes sense to adjust the backend and add an overwrite flag to the api request to determine the behavior the user wants, then we can adjust the frontend accordingly. I'll make an issue for the backend

I try to keep the API design CRUD, so in that sense the existing `POST` request should only be for creating new definitions and would fail if a uniqueness constraint is violated. Replacement would be done either by deleting the original resource and creating a new one that reuses the freed unique attribute, or by issuing an update request (`PATCH`) to the same resource and replacing the contents as needed.

Separate from that:

How does everyone feel about this change?

image

2022-10-24T15:30:34,582419661-04:00

This keeps the Probe Templates subview visible for the user even if they have no Agent-enabled targets around, and also exposes the helper text about the Agent in the same cases. The user can get to this view if they first go to a Target that has the Agent, select that tab, and then select a different Target, but they're locked out from re-entering that subview with that same Target afterward:

image

Josh-Matsuoka:
That change looks good, I have a patch for the backend nearly ready to go that fixes the issue with the notification on removing all probes, it's currently using the wrong notification category. I'll have it up shortly

andrewazores:
Ah perfect, I was just writing a comment to ask about that work when that commit got pushed.

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,304 @@
...
+    context.target.setAuthRetry();
+  }, [context.target, context.target.setAuthRetry]);
+
+  const handleDeleteAllProbes = React.useCallback(() => {
+    addSubscription(
+      context.api
+        .removeProbes()
+        .pipe(first())
+        .subscribe(() => {
+          refreshProbes();
```
tthvo, line 140: I think we can remove this call too. Notification will handle refresh :D

tthvo:
Should also add another test for `ProbeRemoves` notification?

Josh-Matsuoka:
ah I think I missed that in my commit, give me a sec

src/app/Agent/AgentLiveProbes.tsx
```diff
@@ -0,0 +1,304 @@
...
+        setErrorMessage(authFailMessage);
+      })
+    );
+  }, [addSubscription, context.target, setErrorMessage]);
+
+  React.useEffect(() => {
+    addSubscription(
+      context.notificationChannel.messages(NotificationCategory.ProbeTemplateApplied).subscribe((_) => refreshProbes())
+    );
+  }, [addSubscription, context, context.notificationChannel, setProbes]);
```
tthvo, line 191: Should be `refeshProbes` here instead of `setProbes` :D

src/test/Agent/AgentLiveProbes.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+jest
+  .spyOn(defaultServices.api, 'getActiveProbes')
+  .mockReturnValueOnce(of([mockProbe])) // renders correctly
+
+  .mockReturnValueOnce(of([mockProbe])) // should add a probe after receiving a notification
+  .mockReturnValueOnce(of([mockProbe, mockAnotherProbe]))
+
+  .mockReturnValue(of([mockProbe])); // All other tests
+
+jest
```
tthvo, line 119: Notification mock should be as below: 

```typescript
jest
  .spyOn(defaultServices.notificationChannel, 'messages')
  .mockReturnValueOnce(of()) // renders correctly
  .mockReturnValueOnce(of())

  .mockReturnValueOnce(of(mockApplyTemplateNotification)) // should add a probe after receiving a notification
  .mockReturnValueOnce(of())

  .mockReturnValueOnce(of())
  .mockReturnValueOnce(of(mockRemoveProbesNotification))
  
  .mockReturnValue(of()); // All other tests
```

src/test/Agent/AgentLiveProbes.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+    expect(addTemplateName).toBeVisible();
+  });
+
+  it('should remove a probe after receiving a notification', () => {
+    render(
+      <ServiceContext.Provider value={defaultServices}>
+        <AgentLiveProbes />
+      </ServiceContext.Provider>
+    );
+    expect(screen.queryByText('another_name')).not.toBeInTheDocument();
```
tthvo, line 161: We should check that the table is empty. Perhaps both `another_name` and `some_name`? Name of test should be `should remove all probes...`?

tthvo:
We might need to disable the remove button if there are no probes. Right now, users can click and make requests successfully.

Screenshot from 2022-10-24 17-56-10

andrewazores:
I made one more tiny enhancement in my last commit. Is everyone happy with this patchset?

Josh-Matsuoka:
Looks good. Thanks for the help with the cleanup and implementing changes from the review.

andrewazores (approved):

PR #566, opened 2022-10-26
fix(target): disable target creation if connectUrl is left empty

### LINKED ISSUE #565: [Task] Target creation allows empty connectUrl when users press Enter key
Visually, an empty string will disable the create button. However, if the user presses Enter, the target creation form is submitted. This should also be disabled as the Create button.

Screenshot from 2022-10-25 19-18-06

Pressing Enter:

Screenshot from 2022-10-25 19-22-40

tthvo:
Fixes #565 

**Fixes**

- Disable target creation if connectUrl is left empty in both cases (i.e. clicking Create button and pressing Enter).
- Target labels/annotations are changed to generic object since JSON is not mapped into this Map type and fields are added as general support for any objects. See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#setting_object_properties).

**Chores**

- Remove subscription to JVMDiscovery notification since `context.target.targets` will be called to return new array of targets.
- Refactor to use common local storage utilities.


src/app/TargetSelect/TargetSelect.tsx
```diff
@@ -133,55 +109,45 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props)
...
   );
 
   const selectNone = React.useCallback(() => {
-    onSelect(undefined, undefined, true);
+    onSelect(undefined, NO_TARGET, true);
   }, [onSelect]);
 
+  const selectTargetFromCache = React.useCallback(
+    (targets) => {
+      if (targets.length) {
```
andrewazores, line 121: I think this outer if/else can actually be removed. If the `targets` array is empty then the `targets.some()` call later will trivially return `false` and the same handling case runs either way.

```bash
$ node
> ['a'].some(v => true)
true
> [].some(v => true)
false
```

andrewazores:
Testing this out manually, it seems like something is causing a LOT of repeated API requests to the backend. I guess some component is getting re-rendered continuously and causing a request or series of requests each time. Just visiting the Recordings or Events views and selecting a target is enough to trigger it.

src/app/TargetSelect/TargetSelect.tsx
tthvo: Right! I will clean up this part too!

tthvo:
> Testing this out manually, it seems like something is causing a LOT of repeated API requests to the backend. I guess some component is getting re-rendered continuously and causing a request or series of requests each time. Just visiting the Recordings or Events views and selecting a target is enough to trigger it.

Ah yess I saw it now. Sorry didnt see it yesterday. I will have a look to see which commit is causing this.

tthvo:
Seemed like I ran into the classic indefinite effects due to ever-changing deps for effect hooks. Fixed it now :D but I am checking an odd warning when select another target from no target-selected state.

```log
Cannot update a component (`TargetView`) while rendering a different component (`TargetSelect`). To locate the bad setState() call inside `TargetSelect`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
```

tthvo:
It should be ready for reviews again now....

> Seemed like I ran into the classic indefinite effects due to ever-changing deps for effect hooks. Fixed it now :D but I am checking an odd warning when select another target from no target-selected state.
> 
> ```
> Cannot update a component (`TargetView`) while rendering a different component (`TargetSelect`). To locate the bad setState() call inside `TargetSelect`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
> ```

The issue for this is that i was trying to avoid depending on the current target state of `TargetSelect` as callback will change if state changes, so I used setState which returns the currently selected target. However, we are calling `context.target.setTarget()` from inside here, which triggers other components to render and React is pretty much against this.

So, I reverted back to use state and added a comment to warn about this. Perhaps after #429, we can resolve this....

andrewazores (approved):

PR #574, opened 2022-10-26
fix(filters): fix datetimepicker error text alignments

### LINKED ISSUE #570: [Task] Invalid datetime input causes misalignments in toolbar layout
We might need a custom way to show error instead of relying on the Patternfly components.

Screenshot from 2022-10-26 16-38-32

### LINKED ISSUE #571: [Task] Datetime picker should allow local datetime search
Right now, we only allow datetime search using `UTC-0`. I think we should support local datetime (e.g. `UTC-8`). Perhaps, a dropdown of UTC zones or maybe just `UTC` or `localtime`. Then, we can convert them to `UTC-0` in the background.

Screenshot from 2022-10-26 16-51-06

tthvo:
Fixes #570 

Correctly aligned toolbar items when there is an error text in Datetime Picker. A couple refactoring and quick hack with `align-self` and `InputGroup` solved the issue.

Screenshot from 2022-10-26 18-59-41

### DIFF
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -718,7 +718,9 @@ const ActiveRecordingsToolbar: React.FunctionComponent<ActiveRecordingsToolbarPr
           filters={props.targetRecordingFilters}
           updateFilters={props.updateFilters}
         />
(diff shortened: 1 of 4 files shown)


maxcao13 (approved):
Nice, good work!

andrewazores (approved):

PR #584, opened 2022-10-27
fix(agent): action menu should have flip enabled

### LINKED ISSUE #583: [Task] JMC Agent Event Template action menu goes out of viewport
image

If the viewport is not scrolled all the way down before clicking the three-dots action menu, then the dropdown appears and goes outside the viewport. The viewport cannot be scrolled any further and so part of the content is just not reachable.

### LINKED ISSUE #595: [Task] Refactor legacy Table to TableComposable
> I think we should convert all legacy tables to use this new component so that we can use new features later.

_Originally posted by @tthvo in https://github.com/cryostatio/cryostat-web/issues/584#issuecomment-1294435416_

tthvo:
Fixes #583 
Fixed #595 (this PR includes the final refactoring piece).

Refactored probe template table to use custom action menu that enables flipping when viewport boundary is reached.


tthvo:
This is more of a hack. We could do better to configure the menu to open up if there is not enough space below. Not sure if there is a way now with `IAction`. We might need to create a manual popover as recording table.

andrewazores:
I was hoping there was a nice way to do this with the dropdown menu opening upward when needed, but it doesn't seem possible. The only thing is this:

```react
      <Table
        aria-label="Actions table"
        cells={columns}
        rows={rows}
        {...(propToUse === 'actions' && { actions })}
        {...(propToUse === 'actionResolver' && { actionResolver })}
        areActionsDisabled={rowData => !!rowData.disableActions}
        dropdownPosition="left"
        dropdownDirection="down"
        actionsToggle={useCustomToggle ? customActionsToggle : undefined}
      >
```
(https://www.patternfly.org/v4/components/table/react-legacy)

Forcing this dropdown to always open upward seems a little weird. Making the margins larger or just adding some bottom padding to the card seems probably better.

tthvo:
Ahh that unfortunate :((. Are we okayy with `1em` for bottom margin of the card? Though if later we add another option to the action, we might need to add more or refactor to use custom action menu.

andrewazores:
I think a little more bottom margin would be good. Maybe leave the `margin-top` styling and add a `margin-bottom: 2em` or so.

Fixing the menu item to dynamically pop upward would be even better but that's a lot more work for a single small display issue that has an easy workaround aready.

tthvo:
> I think a little more bottom margin would be good. Maybe leave the `margin-top` styling and add a `margin-bottom: 2em` or so, or maybe `padding` to the card so that there's more space between the end of the table and the end of the card and the menu stays more on the card instead of hanging off the edge.
> 
> Fixing the menu item to dynamically pop upward would be even better but that's a lot more work for a single small display issue that has an easy workaround aready.


Right... okayy I refactored the table to use the new `TableComposable`, which allows custom action menu. The action menu is implemented as a `Dropdown` with `isFlipEnabled={true}` and `position={DropdownPosition.right}`.

The menu is flipped vertically as expected. However, the horizontal positioning to the right is not applied. It took a while to find out that Patternfly team actually filed a bug here: https://github.com/patternfly/patternfly-react/issues/8013 (`2022.14`). This does not look like we can upgrade in time for our `2.2.0` as Patternfly release note is only up to `2022.13`.

I guess we have to be on blocking for now.

tthvo:
Also on this note: 

> I refactored the table to use the new TableComposable

I think we should convert all legacy tables to use this new component so that we can use new features later.

andrewazores:
Nice work with the refactoring. Could you open a separate PR to just add the bottom margin then so we can get that merged in the short term, and hopefully for the next release we replace it with this PR and an updated Patternfly version?

tthvo:
> Nice work with the refactoring. Could you open a separate PR to just add the bottom margin then so we can get that merged in the short term, and hopefully for the next release we replace it with this PR and an updated Patternfly version?

Oh right! I opened #590 now :D

tthvo:
Seems like the fix in this Patternfly core is already applied. But the menu is not flipping correctly, for some reason, because the Dropdown was placed within a `Td`. I will take a closer look and might have to go with another way.

tthvo:
The menu is now correctly flipped. Should be ready for review again :D

andrewazores:
> Edit: Actually, we just need to append the menu to document.body instead of parent. Though, this is discouraged for accessibility support.

Are the other changes still related/required then?

tthvo:
> > Edit: Actually, we just need to append the menu to document.body instead of parent. Though, this is discouraged for accessibility support.
> 
> Are the other changes still related/required then?

Oh yes they are required. Sorry, I mean we need to set `document.body` for the new `AgentTemplateAction` defined here instead of `parent` previously before upgrading patternfly/core.

andrewazores (approved):

PR #588, opened 2022-10-28
build(webpack): code-split bundles and add content-hash to output files

### LINKED ISSUE #563: [Task] Web assets should not be cached across version changes
Seems like the way web assets is currently built causes the browser to cache assets if downgrading to older version of Cryostat. From previous discussions:

The issue: 

> I've just been trying to run the cryostat 2.1.0 image directly off quay with CRYOSTAT_IMAGE="theimage", but im not sure where the web-assets are being used because i still see stuff from 2.2.0 when going to localhost:8181

Explanations:

> the way that the web-client build works, your browser may think that it already has up-to-date assets for the cryostat console and use cached versions rather than actually downloading the web-client from the cryostat instance you're connecting to if that's the case then a browser hard refresh (ctrl-shift-r or something) should fix it

> I guess the HTTP headers in use would be ones like If-Modified-Since, and if you're downgrading from a recent build to an older build, then the browser asks the cryostat vertx webserver how old the app.bundle.js is and decides not to download it because the browser has a newer file with the same name from the same origin despite the fact that the file has actually changed and probably has different other attributes like filesize, those attributes wouldn't be checked for

### LINKED ISSUE #587: [Task] Bundle analyzer cannot parse bundles
It seems the analyzer cannot parse the generated bundle go report `parsed` size of files: https://github.com/webpack-contrib/webpack-bundle-analyzer#size-definitions

Following the doc above, analyzer cli usage:

```bash
$ webpack-bundle-analyzer <bundleStatsFile> [bundleDir] [options]
```

So, I think a quick fix would be `webpack-bundle-analyzer ./stats.json dist` in `package.json`.

tthvo:
Fixes #563 
Fixes #587 

- Added configurations to add content-hash to output files and passed the bundle directory to analyzer.

- Applied route-based code splitting: https://reactjs.org/docs/code-splitting.html#route-based-code-splitting. Below is the graph for output bundles sizes:

Screenshot from 2022-10-28 06-07-43

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "prettier:check": "prettier --check './src/**/*.{tsx,ts}'",
     "prettier:apply": "prettier --write './src/**/*.{tsx,ts}'",
     "build:bundle-profile": "webpack --config webpack.prod.js --profile --json > stats.json",
-    "build:bundle-analyze": "webpack-bundle-analyzer ./stats.json",
+    "build:bundle-analyze": "webpack-bundle-analyzer ./stats.json dist",
     "build:bundle-profile-analyze": "npm-run-all -l build:bundle-profile build:bundle-analyze",
     "yarn:install": "yarn install",
     "yarn:frzinstall": "yarn install --frozen-lockfile"
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "@types/victory": "^33.1.5",
     "@typescript-eslint/eslint-plugin": "^4.32.0",
     "@typescript-eslint/parser": "^4.32.0",
+    "@vue/preload-webpack-plugin": "^2.0.0",
     "css-loader": "^5.2.6",
     "css-minimizer-webpack-plugin": "^4.0.0",
     "dotenv-webpack": "^7.1.0",
--- a/src/app/About/About.tsx
+++ b/src/app/About/About.tsx
@@ -43,7 +43,9 @@ import { BreadcrumbPage } from '@app/BreadcrumbPage/BreadcrumbPage';
 import { AboutDescription, CRYOSTAT_TRADEMARK } from './AboutDescription';
 import { Brand, Card, CardBody, CardFooter, CardHeader } from '@patternfly/react-core';
 
-export const About = () => {
+export interface AboutProps {}
+
+export const About: React.FunctionComponent<AboutProps> = (props) => {
   return (
     <BreadcrumbPage pageTitle="About">
       <Card>
@@ -58,3 +60,5 @@ export const About = () => {
     </BreadcrumbPage>
   );
 };
+
+export default About;
--- a/src/app/Archives/Archives.tsx
+++ b/src/app/Archives/Archives.tsx
@@ -62,7 +62,7 @@ export const uploadAsTarget: Target = {
 
 export interface ArchivesProps {}
 
-export const Archives: React.FunctionComponent<ArchivesProps> = () => {
+export const Archives: React.FunctionComponent<ArchivesProps> = (props) => {
   const context = React.useContext(ServiceContext);
   const addSubscription = useSubscriptions();
   const [activeTab, setActiveTab] = React.useState(0);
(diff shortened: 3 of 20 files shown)


tthvo:
Seemed to split the app.bundle.js into chunks but some chunks are split pretty oddly (i.e. `RecordingFilters`). Any guide?

andrewazores:
I wouldn't worry too much about chunks seemingly split in weird ways. It should be something that Webpack is determining makes sense based on our components' dependency graph.

I do notice that the `favicon.ico` fails to load with a `404` now, though. The screen "flashing" between changing views also seems like an unfortunate side-effect of needing to try to download the relevant assets, so the browser displays nothing at all until it either receives the file or loads it from browser cache. I'm not sure if this is caused more by the HTML/JS or by the CSS being loaded late.

Maybe using React's Suspense component and a transition loader helps: https://mono.software/2019/12/30/code-splitting-and-preloading-react-routes-with-webpack/

That post also talks about preload/prefetch which I think would solve a lot of problems. It allows the application to be split into smaller bundles so the first bundles required to display a minimum working web-client can load quickly, and the others load asynchronously after that but not necessarily waiting for the user to click and load a route. I think these are better resources to read about for that:
https://github.com/vuejs/preload-webpack-plugin
https://www.npmjs.com/package/html-prefetch-css-webpack-plugin

Or, worst case scenario, maybe we just put an empty loading state or a loading spinner in the root DOM component that becomes visible between loading transitions: https://stackoverflow.com/questions/40987309/react-display-loading-screen-while-dom-is-rendering

tthvo:
>  The screen "flashing" between changing views also seems like an unfortunate side-effect of needing to try to download the relevant assets, so the browser displays nothing at all until it either receives the file or loads it from browser cache. I'm not sure if this is caused more by the HTML/JS or by the CSS being loaded late.

Ahh I just checked that because I placed the `Suspense` component too high up the tree, the whole page is re-rendered so it takes some time in between. So, now I move it down just right above the `Switch`. This allows the Sidecar to stay intact. I think this somewhat solved the flashing issue.

tthvo:
> I do notice that the favicon.ico fails to load with a 404 now, though. 

hmm this still loads for me...Something odd?

Screenshot from 2022-10-28 15-48-24

Screenshot from 2022-10-28 15-48-42

andrewazores:
This seems to have helped a fair bit at a quick glance - now only the content view has to load, but the main "application frame" (top bar, side nav bar, and grey background) remain the same so the loading flash is much less jarring feeling.

I think adding the preload/prefetch would still be great to cap this off with, if we can.

tthvo:
Right I added `prefetch` configurations since it fits our goal more than `preload`. Looks like https://github.com/vuejs/preload-webpack-plugin also helps to prefetch css files too.

Since our `.css` are loaded in initial chunk, not sure if we need to prefetch them. Right now, `index.html` will contain 2 `<link>` for css: sync load and prefetch. I think sync load will take place?

tthvo:
> Since our .css are loaded in initial chunk, not sure if we need to prefetch them. Right now, index.html will contain 2 <link> for css: sync load and prefetch. I think sync load will take place?

Actually, I will exclude css from prefetching as they are critical and loaded in initial chunk then.

tthvo:
> There are also a duplicate <script> tag for each script with defer attribute. Perhaps, webpack internal work due to code-splitting?

Hmm i think npm bundles and app.bundle are needed asap so I will exclude them again in `prefetch` group.

tthvo:
Generated `index.html` from this settings: https://gist.github.com/tthvo/2cd26ab084d8cd260324bcb1d7f404c9

andrewazores (commented):
LGTM, were there any further changes you had planned?

maxcao13 (approved):
I like the changes with the spinner as well, this is really helpful and the web feels nice to navigate!

tthvo:
> LGTM, were there any further changes you had planned?
> I like the changes with the spinner as well, this is really helpful and the web feels nice to navigate!

Great! I think thats all the changes I planned :D

andrewazores (approved):

PR #590, opened 2022-10-28
fix(agent): add margin bottom for probe template table

tthvo:
Related to #583
Related to #584 

As a temporary solution, the Probe Template table now has a margin bottom to allow the action menu of the last row to be in viewport. 

I used `1.5em` for bottom, just enough to get the menu in viewport but still looks symmetric with top padding/margin.

Screenshot from 2022-10-28 13-19-45

Looks like if the xml spans more than 3 rows, the out-of-view-port is not an issue anymore.

Screenshot from 2022-10-28 13-21-38


maxcao13 (approved):
LGTM!

andrewazores (approved):

PR #604, opened 2022-11-01
fix(recordings): correctly update recording tables on errors

### LINKED ISSUE #600: [Task] Recording filters are not correctly updated if error while connecting to target
To reproduce:

1. Select a target (i.e. authenticated or no-auth) with some recordings.
2. Switch to another target that causes SSL error or Authentication Failure.
3. Select Filters category of any kind (e.g. Label, Name).
4. See the possible filter values of the previous target are still there.

SSL Error:

Screenshot from 2022-11-01 15-05-52

Authentication Failure:

Screenshot from 2022-11-01 15-07-31

Not severely affects the output of the recording tables as error message is shown anyway. But we should correct this and disable the filters somehow if error.

### LINKED ISSUE #603: [Task] Archived recording table needs an error handler
Currently, this table has no error handler. If there is an error, the table is stuck in loading state.

Screenshot from 2022-11-01 17-18-59

tthvo:
Fixes #600 
Fixes #603
Related to #605

**Fixes**
- Clean up recording states if there is an error.
- Added an error handler for ArchivedRecordingTable to avoid getting stuck in loading state.

**Chore**
- Remove call to `refreshRecordings` in ArchivedRecording upload table as notifications will handle update.

### DIFF
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -160,8 +160,9 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
     (error) => {
       setIsLoading(false);
       setErrorMessage(error.message);
+      setRecordings([]);
     },
-    [setIsLoading, setErrorMessage]
+    [setIsLoading, setErrorMessage, setRecordings]
   );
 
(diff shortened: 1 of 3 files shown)


maxcao13 (approved):
Manual testing and code looks good!

PR #611, opened 2022-11-07
chore(graphql): add operation names and extract variables

andrewazores:
Related to cryostatio/cryostat#1201

See also:
- https://graphql.org/learn/queries/
- https://graphql.org/graphql-js/passing-arguments/

Adding operation names makes it easier to search logs server-side for specific queries for debugging. Extracting variables
allows queries to be reused on the server as explained in links above, which allows query caching for better API performance.


andrewazores:
This works independently of cryostatio/cryostat#1202 , but that backend PR includes a parsed query caching enhancement that makes the separated query/variables better performing. Without that backend change this doesn't really do much in practical terms.

tthvo (commented):
Looks good and worked as expected! Just a question and logs seem to show any logging for requests. Oh nvm, just saw https://github.com/cryostatio/cryostat/pull/1214

src/app/Shared/Services/Api.service.tsx
```diff
@@ -792,13 +803,14 @@ export class ApiService {
       filter((target) => target !== NO_TARGET),
       first(),
       concatMap((target) =>
-        this.graphql<any>(`
-        query {
-          targetNodes(filter: { name: "${target.connectUrl}" }) {
+        this.graphql<any>(
+          `
+        query PostActiveRecordingMetadataA($connectUrl: String, $recordingName: String, $labels: String) {
```
tthvo, line 808: -> `PostActiveRecordingMetadata`? There is an extra `A`?

src/app/RecordingMetadata/BulkEditLabels.tsx
```diff
@@ -190,9 +190,9 @@ export const BulkEditLabels: React.FunctionComponent<BulkEditLabelsProps> = (pro
         : context.target.target().pipe(
             filter((target) => target !== NO_TARGET),
             concatMap((target) =>
-              context.api.graphql<any>(`
-              query {
-                targetNodes(filter: { name: "${target.connectUrl}" }) {
+              context.api.graphql<any>(
+                `query GetActiveRecordingsForTarget($connectUrl: String) {
```
tthvo, line 194: I think some operation names do not have `Get` prefix. Maybe we should either remove it or add for others?

tthvo: Like `query AllTargetsArchives` does not have `Get`.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -807,7 +819,9 @@ export class ApiService {
               }
             }
           }
-        }`)
+        }`,
+          { connectUrl: target.connectUrl, recordingName, labels: this.stringifyRecordingLabels(labels) }
```
maxcao13, line 823: To keep consistent, would it be `recordingName: recordingName` ?

andrewazores: It could be, this is just a JS/TS shorthand. It's probably safer to explicitly set the key in the object so that the parameter name can be refactored without changing the structure of the object, but we already do this in many other places too...

maxcao13 (approved):
Looks good to me then!

tthvo (approved):
Looks good to me too :D

PR #624, opened 2022-11-09
fix(label-upload): add error handler for metadata uploads

### LINKED ISSUE #623: [Task] Metadata file upload should handle parsing error
If the metadata file has invalid contents, the UI should handle this: either by showing an error text or just return empty. Right now, the UI will be stuck in loading state.

Screenshot from 2022-11-09 16-46-44

tthvo:
Fixes #623 

Handled parsing error if user uploads invalid metadata label files. The invalid files are shown in a popover on top of the button since inline texts are taking quite some spaces and misaligned. In the background, invalid file contents are ignored (i.e. return empty []).

Screenshot from 2022-11-09 18-11-06

### DIFF
--- a/src/app/RecordingMetadata/RecordingLabelFields.tsx
+++ b/src/app/RecordingMetadata/RecordingLabelFields.tsx
@@ -36,12 +36,15 @@
  * SOFTWARE.
  */
 import * as React from 'react';
-import { CloseIcon, PlusCircleIcon, UploadIcon } from '@patternfly/react-icons';
+import { CloseIcon, ExclamationCircleIcon, FileIcon, PlusCircleIcon, UploadIcon } from '@patternfly/react-icons';
 import {
   Button,
   FormHelperText,
   HelperText,
   HelperTextItem,
+  List,
+  ListItem,
+  Popover,
   Split,
   SplitItem,
   Text,
@@ -51,7 +54,7 @@ import {
 import { parseLabelsFromFile, RecordingLabel } from '@app/RecordingMetadata/RecordingLabel';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
 import { LoadingView } from '@app/LoadingView/LoadingView';
-import { Observable, zip } from 'rxjs';
+import { catchError, Observable, of, zip } from 'rxjs';
 
 export interface RecordingLabelFieldsProps {
   labels: RecordingLabel[];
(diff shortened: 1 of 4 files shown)


tthvo:
Any thoughts on this design?

andrewazores:
Design looks good. I think the two texts are mostly saying the same thing though, so either one should be rephrased or the one inside the component body should be removed.

Maybe,
title: "Invalid Selection"
message: "The following files did not contain valid recording metadata:"

andrewazores (commented):
Code looks good

tthvo:
> I think the two texts are mostly saying the same thing though, so either one should be rephrased or the one inside the component body should be removed.

Nice! Thanks I updated the modal now.

Screenshot from 2022-11-09 18-41-23

andrewazores:
LGTM, I'll give @maxcao13 a chance to look at this too

src/app/RecordingMetadata/RecordingLabelFields.tsx
```diff
@@ -187,9 +194,31 @@ export const RecordingLabelFields: React.FunctionComponent<RecordingLabelFieldsP
...
+            isVisible={!!invalidUploads.length}
+            aria-label="uploading warning"
+            alertSeverityVariant="danger"
+            headerContent="Invalid Selection"
+            headerComponent="h1"
+            shouldClose={closeWarningPopover}
+            headerIcon={<ExclamationCircleIcon />}
+            bodyContent={
+              <>
+                <Text component="h4">The following files did not contain valid recording metadata:</Text>
```
maxcao13, line 207: The only thing I would consider is that making files -> `file(s)` because there might be only 1 file. But otherwise looks good!

tthvo: Got it! Updated now :D

andrewazores: May as well dynamically set the text to singular or plural depending on list size then ;-)

andrewazores: Nice, I see Thuan just did that haha.

Gonna be fun when we get around to localizing these strings!

tthvo: Hoping its not gonna be a nightmare XD

andrewazores: #625

PR #633, opened 2022-11-10
build(deps): remove unused express dep

andrewazores:
Related to #629

Remove unused dependency. Express is a web framework for server-side (Node) applications. This project is only built to static assets to be served by the Cryostat backend and we never use Express for anything.


tthvo (approved):
Looks good to me! I guess the express that is still listed in `yarn.lock` is from `webpack` for hot reload?

andrewazores:
Yea exactly, it looks like the webpack-dev-server needs it so it is a transitive dep still, but we don't directly depend on it.

PR #634, opened 2022-11-10
build(deps): remove unused type definitions for react-test-renderer

andrewazores:
Related to #631

This dependency adds TypeScript type definitions for the `react-test-renderer` testing library, which we don't actually use or even have a direct dependency on itself - so there is definitely no need to add TypeScript type definitions.


tthvo (approved):
Looks good!

PR #640, opened 2022-11-10
feat(spinner): add spinner to action buttons

### LINKED ISSUE #597: [Task] Add progress spinner to action buttons
Related to https://github.com/cryostatio/cryostat/pull/1089
Related to #592

Creating a custom target definition results in Cryostat trying to connect to that target to determine its unique JVM ID. This takes some time and may time out after a few seconds if the definition used a JMX URL that points to nothing, for example. The custom target creation UI should be updated to handle this more nicely - the text boxes should be disabled and a spinner displayed overtop when a request is in progress. On success the modal should close, on failure the spinner should disappear and the text boxes re-enable but the modal remain.

tthvo:
Fixes #597 

The following actions are configured to show progress indicator:

- [x] Target
  - [ ] Create
  - [ ] Delete
- [ ] Automatic Rule
  - [ ] Create
- [ ] Recording
  - [ ] Create
  - [ ] Delete
  - [ ] Stop
  - [ ] Archive 
  - [ ] Save metadata labels
- [ ] Security
   - [ ] Store credentials


andrewazores:
Anything else left todo @tthvo ?

tthvo:
Oh just going through each one again to check if error handler is added or else it will be stuck in loading state :D

tthvo:
Ready for review now. Please let me know if any other places need adding.

andrewazores:
Haven't reviewed the code yet but manual test looks great.

src/app/Shared/ProgressIndicator.tsx
```diff
@@ -35,33 +35,10 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-import * as React from 'react';
-import { EventTemplate } from '../CreateRecording/CreateRecording';
```
maxcao13, line 38: Will there be more in this file? I'm just wondering cause there is just a props type definition called LoadingPropsType inside a tsx file called ProgressIndicator and I'm not sure if those are obviously correlated.

tthvo:
Also just a note on this: buttons now are a little wider. I think its the effect of adding loading props.

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -313,18 +319,44 @@ export const ActiveRecordingsTable: React.FunctionComponent<ActiveRecordingsTabl
...
+    checkedIndices,
+    handleRowCheck,
+    context.api,
+    addSubscription,
+    setActionLoadings,
+    handlePostActions,
+  ]);
 
   const handleStopRecordings = React.useCallback(() => {
+    setActionLoadings((old) => ({ ...old, STOP: true }));
```
maxcao13, line 359: is this equivalent to handlePostActions('STOP') ?

tthvo: hmm actually, the `handlePostActions` will reset this to `false` to disable loading.

src/app/Shared/ProgressIndicator.tsx
andrewazores: I'm not sure if we would do linear progress inside the button. It seems like that would either be a very small and hard to see progress bar, or a very large/wide button.

https://www.patternfly.org/v4/components/button/design-guidelines#progress-button

I'm not sure what to do about the checkmark icon design guideline here, it doesn't seem to fit in well with our specific use cases so maybe we just ignore that for now.

https://www.patternfly.org/v4/components/progress

I think we would use this component, perhaps during a file upload for example, but it would appear within the upload modal above the buttons and below the content perhaps rather than being part of the button.

image

src/app/Recordings/ActiveRecordingsTable.tsx
maxcao13: Ah okay.

src/app/Shared/ProgressIndicator.tsx
tthvo: Niceee! This is exactly what I have in mind for multi-upload issue. Thanks for pointing out this. It would help a lot!

andrewazores: Some more reading:
https://ilikekillnerds.com/2020/09/file-upload-progress-with-the-fetch-api-is-coming/
https://www.npmjs.com/package/web-streams-polyfill

maxcao13 (approved):
LGTM!

andrewazores (approved):

Issue #656, opened 2022-11-15
[Task] "Session credentials" mode seems unreliable for JMX connections


andrewazores:
Using "Session credentials" seems unreliable or intermittently fails when connecting to different JMX target applications or Cryostat itself. "Backend credentials" stored in the JMX keyring seem reliable, and the backend itests exercising the `X-JMX-Authorization` header (the mechanism behind Session credentials) seem solid, so this might just be a frontend error handling bug.

maxcao13:
> This might be something we wait to fix until Cryostat 3.0, honestly. It seems to me the best way to handle this is to rearchitect a lot of Cryostat's internal infastructure to be scope to the lifetime of the request rather than having a static lifetime for the whole application. ie, there would be a new TargetConnectionManager instance created for each request that needs one, with direct access to the request context for retrieving X-JMX-Authorization header credentials, and a new ActiveRecordingHelper also scoped to the request lifecycle for the same reason, etc.

Reposting for visibility from https://github.com/cryostatio/cryostat-web/pull/672#issuecomment-1320438850

What should the plan be then? It seems like the whole session storage might have to be scrapped until then...

andrewazores:
Well, there is one other hack we can apply in the meantime. I don't think I like it and don't take this as me committing to it.

In ex. `AbstractAuthenticatedRequestHandler`, or other implementations that handle the `X-JMX-Authorization` header, we can check for the header and if it's present, define new credentials into the keyring using that and the `targetId`/`sourceTarget` from the request URL path. If we do this then we can also add an `endHandler` onto the `RoutingContext` that will remove these credentials from the keyring when the request is completed.

This should make the `X-JMX-Authorization` header mechanism work again, but it is no longer a session-only, no-storage passthrough mechanism - the credentials would still end up briefly in the database before being deleted again. This also means that if clients are making different requests to the same target JVM with different JMX credentials for different JMX users that may exist then things can get messed up and interleaved unless we define all the endpoints as ordered so that Vert.x only processes them serially, but that kills throughput.

maxcao13:
Wondering, when is 2.2.1 slated to be released? I believe this feature should definitely be fixed in the bug fix release.

I definitely think there are some hacks that can be put in place until the big refactor like the one you mention, and I dont *think* it should be too difficult to put in.

andrewazores:
It's too late for 2.2.1, the merge window for that would've been a couple weeks back. The downstream release is in progress and should be delivered any day now.

[...]

PR #672, opened 2022-11-18
fix(credentials): set backend storage as default

tthvo:
Related to #656 

Use Backend storage as default. This does not fix the issue but just switch it default Backend.

### DIFF
--- a/src/app/Settings/CredentialsStorage.tsx
+++ b/src/app/Settings/CredentialsStorage.tsx
@@ -39,7 +39,6 @@
 import * as React from 'react';
 import { Link } from 'react-router-dom';
 import { Select, SelectOption, SelectVariant, Text } from '@patternfly/react-core';
-
 import { UserSetting } from './Settings';
 import { getFromLocalStorage, saveToLocalStorage } from '@app/utils/LocalStorage';
 
@@ -69,12 +68,12 @@ const getLocation = (key: string): Location => {
       return l;
     }
   }
-  return Locations.BROWSER_SESSION;
+  return Locations.BACKEND;
 };
 
 const Component = () => {
   const [isExpanded, setExpanded] = React.useState(false);
-  const [selection, setSelection] = React.useState(Locations.BROWSER_SESSION.key);
+  const [selection, setSelection] = React.useState(Locations.BACKEND.key);
 
   const handleSelect = React.useCallback(
     (_, selection) => {
@@ -87,7 +86,7 @@ const Component = () => {
   );
 
   React.useEffect(() => {
-    handleSelect(undefined, getFromLocalStorage('JMX_CREDENTIAL_LOCATION', Locations.BROWSER_SESSION.key));
+    handleSelect(undefined, getFromLocalStorage('JMX_CREDENTIAL_LOCATION', Locations.BACKEND.key));
   }, [handleSelect, getFromLocalStorage]);
 
   return (
--- a/src/app/Shared/Services/JmxCredentials.service.tsx
+++ b/src/app/Shared/Services/JmxCredentials.service.tsx
@@ -52,7 +52,7 @@ export class JmxCredentials {
   constructor(private readonly api: () => ApiService) {}
 
   setCredential(targetId: string, username: string, password: string): Observable<boolean> {
-    let location = getFromLocalStorage('JMX_CREDENTIAL_LOCATION', Locations.BROWSER_SESSION.key);
+    let location = getFromLocalStorage('JMX_CREDENTIAL_LOCATION', Locations.BACKEND.key);
     switch (location) {
       case Locations.BACKEND.key:
         return this.api().postCredentials(`target.connectUrl == "${targetId}"`, username, password);
@@ -66,7 +66,7 @@ export class JmxCredentials {
   }
 
   getCredential(targetId: string): Observable<Credential | undefined> {
-    let location = getFromLocalStorage('JMX_CREDENTIAL_LOCATION', Locations.BROWSER_SESSION.key);
+    let location = getFromLocalStorage('JMX_CREDENTIAL_LOCATION', Locations.BACKEND.key);
     switch (location) {
       case Locations.BACKEND.key:
         // if this is stored on the backend then Cryostat should be using those and not prompting us to request from the user
--- a/src/test/Settings/CredentialsStorage.test.tsx
+++ b/src/test/Settings/CredentialsStorage.test.tsx
@@ -37,7 +37,7 @@
  */
 import * as React from 'react';
 import renderer, { act } from 'react-test-renderer';
-import { render, cleanup, screen, waitFor, within } from '@testing-library/react';
+import { cleanup, screen, waitFor, within } from '@testing-library/react';
 import { renderDefault } from '../Common';
 import { CredentialsStorage } from '@app/Settings/CredentialsStorage';
 import { getFromLocalStorage, saveToLocalStorage } from '@app/utils/LocalStorage';
@@ -73,15 +73,15 @@ describe('<CredentialsStorage/>', () => {
     expect(tree.toJSON()).toMatchSnapshot();
   });
 
-  it('defaults to Session storage', async () => {
+  it('defaults to Backend storage', async () => {
     renderDefault(React.createElement(CredentialsStorage.content, null));
 
     expect(getFromLocalStorage).toHaveBeenCalledTimes(1);
     expect(saveToLocalStorage).toHaveBeenCalledTimes(1);
-    expect(saveToLocalStorage).lastCalledWith(storageKey, sessionStorageValue);
+    expect(saveToLocalStorage).lastCalledWith(storageKey, backendStorageValue);
 
-    expect(screen.getByText(sessionStorageValue)).toBeVisible();
-    expect(screen.queryByText(backendStorageValue)).toBeFalsy();
+    expect(screen.getByText(backendStorageValue)).toBeVisible();
+    expect(screen.queryByText(sessionStorageValue)).toBeFalsy();
   });
 
   it('sets value to local storage when dropdown is clicked', async () => {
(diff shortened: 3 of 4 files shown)


src/app/AppLayout/AuthModal.tsx
```diff
@@ -52,39 +52,37 @@ export interface AuthModalProps {
...
-        );
-      });
+    (username: string, password: string) => {
+      setLoading(true);
+      addSubscription(
+        context.target
+          .target()
+          .pipe(
+            filter((target) => target !== NO_TARGET),
+            first(),
```
tthvo, line 66: I changed the order of `first()` after `filter()` to be wait for any valid target (consistent with other views). However, i don't think it really matters since the modal disables any interaction below it.

Just wanna give a head ups :D

tthvo:
Actually, I messed around with UI and somehow found this way to consistently reproduce it. The steps are similar to the description of #673

## To reproduce
1. Run backend with:
    ```bash
    $ CRYOSTAT_DISABLE_SSL=true CRYOSTAT_CORS_ORIGIN=http://localhost:9000 sh smoketest.sh
    ```
3. Run front-end with:
    ```bash
    $ yarn start:dev
    ```
4. Login and select target `io.cryostat.Cryostat`.
5. Enter `http://localhost:9000/recordings/create` on browser.
6. Dismiss auth form modal.
7. Choose snapshot recording tab.
8. Click create.
9. Enter credentials (`smoketest`).
10. See error occurs.

Could you try if this happens on your end too?


## Observations

Requests are sent with correct username/password (i.e. credential map holds correct record and console.log shows correct field being added).

andrewazores:
Yes, these reproduction steps work for me.

Interestingly, the `Authorization` header *and* the `X-JMX-Authorization` header are actually being sent with the first request according to the browser devtools network tab, and yet the response is a 427 failure. But if I navigate over to the Events view, for example, and enter the JMX credentials on the modal, then I can see requests going out with the same headers and yet receiving 200 responses.

andrewazores:
So it looks like this is probably actually a backend bug after all. Setting the credentials storage to Backend is still an effective workaround.

andrewazores:
Also very interesting. After reproducing the bug with Thuan's steps above, I tried setting the storage to Backend, then closing and reopening the web-client browser tab. Then I went back to the Recordings view. There were multiple `snapshot-n` recordings on the target that were just created in the minutes prior. So somehow the frontend was actually sending requests to create snapshots even though I had only navigate to the snapshot creation view but not clicked the creation button, and yet the responses to those requests were 427 statuses even though it appears the request succeeded.

tthvo:
> Also very interesting. After reproducing the bug with Thuan's steps above, I tried setting the storage to Backend, then closing and reopening the web-client browser tab. Then I went back to the Recordings view. There were multiple snapshot-n recordings on the target that were just created in the minutes prior. So somehow the frontend was actually sending requests to create snapshots even though I had only navigate to the snapshot creation view but not clicked the creation button, and yet the responses to those requests were 427 statuses even though it appears the request succeeded.

Interesting, i saw the same issue. It seems like its only broken in Recording view. Should we transfer this issue to the backend repo? Also are we good with this workaround?

andrewazores:
Let's confirm the exact root cause of the problem before transferring the issue, just so we don't end up transferring it back and forth. Worth considering still that there are backend itests that exercise what should be the same code path and those pass, so there's something odd going on at some level or another and I don't think we've fully nailed down what or where it is yet.

Anyway I'm good with the workaround of setting the default to backend. That's actually closer to what 2.1 was doing anyway.

Could you split out the change that changes the default into a separate PR, which can be merged and backported separately without these other refactoring fixes? Only the change of default should go into a backport for 2.2.1.

tthvo:
> Could you split out the change that changes the default into a separate PR, which can be merged and backported separately without these other refactoring fixes? Only the change of default should go into a backport for 2.2.1.

Okayy updated :D I will open another PR for the clean up.

andrewazores (approved):

maxcao13:
Just confirming that I see the same issue.

My first instinct is that it might be because computing a jvmId uses back-end credentials instead of using session memory? I'm not sure if it does, but I see
```
org.openjdk.jmc.rjmx.ConnectionException caused by java.lang.SecurityException: Authentication failed! Invalid username or password
	at org.openjdk.jmc.rjmx.internal.RJMXConnection.connect(RJMXConnection.java:441)
	at io.cryostat.core.net.JFRConnection.attemptConnect(JFRConnection.java:310)
	at io.cryostat.core.net.JFRConnection.connect(JFRConnection.java:267)
	at io.cryostat.core.net.JFRConnection.getJvmId(JFRConnection.java:178)
	at io.cryostat.recordings.JvmIdHelper.lambda$computeJvmId$4(JvmIdHelper.java:161)
	at io.cryostat.net.TargetConnectionManager.lambda$executeConnectedTaskAsync$2(TargetConnectionManager.java:148)
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:646)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:483)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
Caused by: java.lang.SecurityException: Authentication failed! Invalid username or password
```
then maybe it tries to use credentials it can't find...

andrewazores:
Aha, that's a good insight. I didn't check the backend logs during the test run but that makes a lot of sense. I think that's the next most obvious candidate bug to work through. If that is the cause then in order for the Session storage to work properly again we'll need to introduce some larger notion of a request state/session on the backend so that the JMX credentials can follow through the execution path. That might be tough.

maxcao13:
> Aha, that's a good insight. I didn't check the backend logs during the test run but that makes a lot of sense. I think that's the next most obvious candidate bug to work through. If that is the cause then in order for the Session storage to work properly again we'll need to introduce some larger notion of a request state/session on the backend so that the JMX credentials can follow through the execution path. That might be tough. Though, I'm not sure then why this bug manifests on the Recordings view and not Events.

I think it's just cause eventGet endpoint doesn't generate or get jvmIds but things like RecordingPost, SnapshotPost, MetadataPost all do... I think everything glaringly points to jvmIds at this point, and I agree there the two methods should somehow follow the same execution path to compute the ids for the session state to work.

andrewazores:
Ahh, those things need the JVM ID present because they do things like storing metadata or archives, and nothing in the Events view does that so the JVM ID code path never gets hit. I see.

Yea... that's a little rough. So for session authentication to continue working we do need some way to allow those credentials to pass from the API handlers all the way through the call stack. Not conceptually hard but that's going to be a large refactoring surface.

maxcao13:
Yes, stuff like 
```
this.targetConnectionManager.executeConnectedTaskAsync(
                            new ConnectionDescriptor(
                                    sr.getServiceUri().toString(),
                                    credentialsManager.getCredentials(sr)),
                            connection -> {
                                return connection.getJvmId();
```
wouldn't work because the `credentialsManager` is only searching the backend for credentials and is unaware of any session state.

andrewazores:
So then perhaps we don't actually have any integration tests that exercise this path then, right? I know we have tests that exercise starting recordings and taking snapshots, test that exercise retrieving event types and templates, and tests that use JMX Auth, but maybe we don't have tests that try to start recordings on targets that need auth.

```
$ ag X-JMX-Authorization src/test/java/itest/
src/test/java/itest/JmxAuthIT.java
60:    private static final String X_JMX_AUTHORIZATION = "X-JMX-Authorization";

src/test/java/itest/util/http/JvmIdWebRequest.java
86:                    "X-JMX-Authorization",

src/test/java/itest/AutoRulesIT.java
291:                        "X-JMX-Authorization",
376:                            "X-JMX-Authorization",
475:                            "X-JMX-Authorization",
599:                            "X-JMX-Authorization",
638:                            "X-JMX-Authorization",

src/test/java/itest/InterleavedExternalTargetRequestsIT.java
211:                                        "X-JMX-Authorization",
241:                            "X-JMX-Authorization",
285:                                        "X-JMX-Authorization",
```

andrewazores:
`RecordingWorkflowIT` would be the obvious candidate for an integration test to update and add `X-JMX-Authorization` credentials to for simulating the frontend session setting, assuming none of those tests above are already doing something similar.

maxcao13:
> `RecordingWorkflowIT` would be the obvious candidate for an integration test to update and add `X-JMX-Authorization` credentials to for simulating the frontend session setting, assuming none of those tests above are already doing something similar.
> 
> Though, it looks like `InterleavedExternalTargetRequestsIT` is already doing that:
> 
> ```java
>     private void createInMemoryRecordings() throws Exception {
>         List<CompletableFuture<Void>> cfs = new ArrayList<>();
>         for (int i = 0; i < CONTAINERS.size(); i++) {
>             final int fi = i;
>             CompletableFuture<Void> cf = new CompletableFuture<>();
>             cfs.add(cf);
>             Podman.POOL.submit(
>                     () -> {
>                         MultiMap form = MultiMap.caseInsensitiveMultiMap();
>                         form.add("recordingName", "interleaved-" + fi);
>                         form.add("events", "template=Continuous");
>                         webClient
>                                 .post(
>                                         String.format(
>                                                 "/api/v1/targets/%s/recordings",
>                                                 Podman.POD_NAME + ":" + (9093 + fi)))
>                                 .putHeader(
>                                         "X-JMX-Authorization",
>                                         "Basic "
>                                                 + Base64.getUrlEncoder()
>                                                         .encodeToString(
>                                                                 "admin:adminpass123".getBytes()))
>                                 .sendForm(
>                                         form,
>                                         ar -> {
>                                             if (assertRequestStatus(ar, cf)) {
>                                                 cf.complete(null);
>                                             }
>                                         });
>                     });
>         }
>         CompletableFuture.allOf(cfs.toArray(new CompletableFuture[0]))
>                 .get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
>     }
> ```

Is JMX auth enabled for those targets? Don't they need a key map of "JMX_AUTH, true" or something like that?

andrewazores:
This might be something we wait to fix until Cryostat 3.0, honestly. It seems to me the best way to handle this is to rearchitect a lot of Cryostat's internal infastructure to be scope to the lifetime of the request rather than having a static lifetime for the whole application. ie, there would be a new `TargetConnectionManager` instance created for each request that needs one, with direct access to the request context for retrieving `X-JMX-Authorization` header credentials, and a new `ActiveRecordingHelper` also scoped to the request lifecycle for the same reason, etc.

andrewazores:
> > Though, it looks like `InterleavedExternalTargetRequestsIT` is already doing that:
> > ...code
> 
> Is JMX auth enabled for those targets? Don't they need a key value pair of "JMX_AUTH, true" in the `new Podman.ImageSpec` envs or something like that?

Ha. You are correct, the test is broken and the external app containers don't actually require JMX auth. Wow...

andrewazores:
```java
    @BeforeAll
    static void setup() throws Exception {
        Set<Podman.ImageSpec> specs = new HashSet<>();
        for (int i = 0; i < NUM_EXT_CONTAINERS; i++) {
            specs.add(
                    new Podman.ImageSpec(
                            FIB_DEMO_IMAGESPEC, Map.of("JMX_PORT", String.valueOf(9093 + i))));
                            // This Map also needs "USE_AUTH", "true"
        }
        for (Podman.ImageSpec spec : specs) {
            CONTAINERS.add(Podman.run(spec));
        }
        CompletableFuture.allOf(
                        CONTAINERS.stream()
                                .map(id -> Podman.waitForContainerState(id, "running"))
                                .collect(Collectors.toList())
                                .toArray(new CompletableFuture[0]))
                .join();
        waitForDiscovery(NUM_EXT_CONTAINERS);
    }

```

andrewazores:
@maxcao13 @tthvo would one of you like to take on fixing that itest above so it actually sets up the containers to require credentials? It might end up failing right now, so I think we can mark it as a known failure like this: https://stackoverflow.com/questions/4055022/mark-unit-test-as-an-expected-failure-in-junit , at least until we come up with some fix for the implementation.

maxcao13:
> would one of you like to take on fixing that itest above so it actually sets up the containers to require credentials? It might end up failing right now, so I think we can mark it as a known failure like this: https://stackoverflow.com/questions/4055022/mark-unit-test-as-an-expected-failure-in-junit , at least until we come up with some fix for the implementation.

Sure I can do that.

PR #687, opened 2022-11-22
fix(create-recordings): add error-views and refactor target listener

### LINKED ISSUE #673: [Task] View sub-route should check for target authentication and ssl
Right now, sub-route does not show error view if auth fails or missing ssl certificates.

## How to reproduce

1. Run backend with:
    ```bash
    $ CRYOSTAT_DISABLE_SSL=true CRYOSTAT_CORS_ORIGIN=http://localhost:9000 sh smoketest.sh
    ```
3. Run front-end with:
    ```bash
    $ yarn start:dev
    ```
4. Login and select target `io.cryostat.Cryostat`.
5. Enter `http://localhost:9000/recordings/create` on browser.
6. Dismiss auth form modal.
7. See recording form is still shown instead of error view.

### LINKED ISSUE #686: [Task] Switching target does not update event template list in recording creation form
## How To Reproduce

Step 1: Go to Recordings and click Create (might need to enter credentials).

Step 2: View the event templates dropdown.

Step 3: Choose another target that has a different list of event templates (i.e extra custom templates or requires authentications, if needing auth, dismiss).

Step 4: Check the event template dropdown again. The content does not update.

## What Is Expected

The event template dropdown updates when a new target is selected.

## What Actually Happened

The event template dropdown does not update when a new target is selected.

broken-target-listener

tthvo:
Fixed #673 
Fixed #686 

The following fixes were applied:

- [x] Add a missing error view to `CustomRecordingForm`.
- [x] Add a missing error view to `SnapShotRecordingForm` (only auth failure and ssl error will show).
- [x] Refactor target selection listener of `CustomRecordingForm` to update UI when the target selection is changed.

Others:
- [x] Add unit tests for `SnapShotRecordingForm`.

### DIFF
--- a/src/app/CreateRecording/CustomRecordingForm.tsx
+++ b/src/app/CreateRecording/CustomRecordingForm.tsx
@@ -72,6 +72,7 @@ import { LoadingPropsType } from '@app/Shared/ProgressIndicator';
 import { TemplateType } from '@app/Shared/Services/Api.service';
 import { authFailMessage, ErrorView, isAuthFail } from '@app/ErrorView/ErrorView';
 import { NO_TARGET } from '@app/Shared/Services/Target.service';
+import { forkJoin } from 'rxjs';
 
 export interface CustomRecordingFormProps {
   prefilled?: {
@@ -280,7 +281,7 @@ export const CustomRecordingForm: React.FunctionComponent<CustomRecordingFormPro
     handleCreateRecording,
   ]);
 
-  const refreshTemplates = React.useCallback(() => {
+  const refeshFormOptions = React.useCallback(() => {
     addSubscription(
       context.target
         .target()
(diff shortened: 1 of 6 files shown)


tthvo:
Error handler for `SnapShotRecordingForm` is a little complex: 
- Must add listener for `authRetry` to update form because failed api requests from custom recording tab also trigger any listener for `authFailure` and valid credentials do not update snapshot tab because it originates from custom recording tab.
- Must add listener for `sslFailure` because `context.api.createRecording` only returns true/false. Other errors are shown through notification messages.
- Must add listener for `context.target.target` to reset form if the target selection is changed.

maxcao13 (approved):
Nice, I like the changes, great job!

tthvo:
Ohh thanks! Looking back again at `CustomRecordingForm`, I think having 2 independent api calls for recording form data (i.e. template dropdown and recordingOptions) might cause an issue if only 1 call fails (i.e the result of the api request that comes later will decide if the error view should be shown).

To tackle this, I added a small fix with `forkJoin` that avoid above issues :D You can review again^^

andrewazores:
Looks good, just rebase to resolve the conflict please.

andrewazores (approved):

Issue #651, opened 2022-11-14
[Task] Support for aborting api requests


tthvo:
Api requests in modals or forms should be able to be aborted when clicking on `Cancel` or `Close` button. Though, I think there should be some server-side clean-up?

PR #607, opened 2022-11-02
feat(labels): support multiple metadata label uploads

tthvo:
Fixes #606 

Users can now choose multiple label files to upload into the label editor. This helps avoiding uploading files one by one.

PR #691, opened 2022-11-25
feat(uploads): allow users to upload multiple files

### LINKED ISSUE #613: [Task] Support for multiple-file uploading of archived recordings
Related to https://github.com/cryostatio/cryostat-web/issues/606

We can follow this example: https://www.patternfly.org/v4/components/file-upload---multiple

tthvo:
Fixes #613 
Partially solved #651 
Related to #607 
Depends on #684 (need that refactoring of upload modals)
Depends on #708 (refresh causes modal to close unexpectedly)
Depends on https://github.com/cryostatio/cryostat/pull/1273 (backend for #708)

The following upload modals now allow multiple uploads. Each file initiates a separate api request.

- [x] Automatic Rules
- [x] Event Templates
- [x] Agent Templates
- [x] Archives
- [x] SSL Certificates
- [x] Metadata Files (fixed in #607)

One issue is that fetch API does not have support for upload progress yet (or only certain browser versions have support). To work around this, I "mirrored" the current `sendRequest` to one that uses `XMLHttpRequest`. This way we can take advantage of upload progress and guarantee support from most browsers. 

Also, by using XMLHttpRequest, we can no longer use `AbortController` for aborting. Aborting mechanisms need investigating and adding back in #651 

**Other fixes and chores**
- [x] Use forkJoin to join multiple api calls instead of zip (will be deprecated later).
- [x] Clean up api call handlers (i.e. internally calls first(), remove unneeded notification, add missing error catch).
- [x] Add missing tests for uploading custom event template.

**How it looks**

multi-file-upload-demo

### DIFF
--- a/src/app/Agent/AgentProbeTemplates.tsx
+++ b/src/app/Agent/AgentProbeTemplates.tsx
@@ -35,55 +35,55 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-import * as React from 'react';
-import { ServiceContext } from '@app/Shared/Services/Services';
+import { ErrorView } from '@app/ErrorView/ErrorView';
+import { LoadingView } from '@app/LoadingView/LoadingView';
+import { DeleteWarningModal } from '@app/Modal/DeleteWarningModal';
+import { DeleteWarningType } from '@app/Modal/DeleteWarningUtils';
+import { FUpload, MultiFileUpload, UploadCallbacks } from '@app/Shared/FileUploads';
+import { LoadingPropsType } from '@app/Shared/ProgressIndicator';
+import { ProbeTemplate } from '@app/Shared/Services/Api.service';
 import { NotificationCategory } from '@app/Shared/Services/NotificationChannel.service';
+import { ServiceContext } from '@app/Shared/Services/Services';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
 import {
   ActionGroup,
   Button,
-  FileUpload,
+  Dropdown,
+  DropdownItem,
+  DropdownPosition,
+  EmptyState,
+  EmptyStateIcon,
   Form,
   FormGroup,
+  KebabToggle,
   Modal,
   ModalVariant,
+  Stack,
+  StackItem,
+  TextInput,
+  Title,
   Toolbar,
   ToolbarContent,
   ToolbarGroup,
   ToolbarItem,
-  TextInput,
-  StackItem,
-  Stack,
-  Dropdown,
-  DropdownItem,
-  KebabToggle,
-  DropdownPosition,
-  EmptyState,
-  EmptyStateIcon,
-  Title,
 } from '@patternfly/react-core';
 import { SearchIcon, UploadIcon } from '@patternfly/react-icons';
 import {
-  TableVariant,
   ISortBy,
   SortByDirection,
-  Tr,
-  Td,
-  ThProps,
   TableComposable,
+  TableVariant,
   Tbody,
+  Td,
   Th,
   Thead,
+  ThProps,
+  Tr,
 } from '@patternfly/react-table';
-import { first } from 'rxjs/operators';
-import { LoadingView } from '@app/LoadingView/LoadingView';
-import { ErrorView } from '@app/ErrorView/ErrorView';
-import { ProbeTemplate } from '@app/Shared/Services/Api.service';
-import { DeleteWarningType } from '@app/Modal/DeleteWarningUtils';
-import { DeleteWarningModal } from '@app/Modal/DeleteWarningModal';
+import * as React from 'react';
+import { forkJoin, Observable, of } from 'rxjs';
+import { catchError, defaultIfEmpty, first, tap } from 'rxjs/operators';
 import { AboutAgentCard } from './AboutAgentCard';
-import { NotificationsContext } from '@app/Notifications/Notifications';
-import { LoadingPropsType } from '@app/Shared/ProgressIndicator';
 
 export interface AgentProbeTemplatesProps {
   agentDetected: boolean;
@@ -148,12 +148,9 @@ export const AgentProbeTemplates: React.FunctionComponent<AgentProbeTemplatesPro
   const handleDelete = React.useCallback(
     (template: ProbeTemplate) => {
       addSubscription(
-        context.api
-          .deleteCustomProbeTemplate(template.name)
-          .pipe(first())
-          .subscribe(() => {
-            /** Do nothing. Notifications hook will handle */
-          })
+        context.api.deleteCustomProbeTemplate(template.name).subscribe(() => {
+          /** Do nothing. Notifications hook will handle */
+        })
       );
     },
     [addSubscription, context.api]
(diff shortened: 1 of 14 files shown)


src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -356,61 +354,152 @@ export interface AgentProbeTemplateUploadModalProps {
...
+  const handleFileRemove = React.useCallback(
+    (removedFileUpload: FUpload) => {
+      setFileUploads((old) => old.filter((fileUpload) => fileUpload.file.name !== removedFileUpload.file.name));
+    },
+    [setFileUploads]
+  );
+
+  const getProgressUpdateCallback = React.useCallback(
+    (filename: string) => {
+      return (process: string | number) => {
```
andrewazores, line 403: process -> progress?

src/app/Archives/ArchiveUploadModal.tsx
```diff
@@ -122,40 +136,126 @@ export const ArchiveUploadModal: React.FunctionComponent<ArchiveUploadModalProps
     }
   }, [uploading, setShowCancelPrompt, reset, props.onClose]);
 
+  const getProgressUpdateCallback = React.useCallback(
+    (filename: string) => {
+      return (process: string | number) => {
```
andrewazores, line 141: same here

src/app/Archives/ArchiveUploadModal.tsx
```diff
@@ -122,40 +136,126 @@ export const ArchiveUploadModal: React.FunctionComponent<ArchiveUploadModalProps
...
+                },
+              } as FUpload,
+            ];
+          }
+          return old;
+        });
+      };
+    },
+    [setFileUploads]
+  );
```
andrewazores, line 161: Is there any refactoring that can be done to extract these callbacks so they're reusable across components? It seems like they're basically just copied and pasted between here and the `AgentProbleTemplates` for example. It'd be nice to have a single reused implementation.

andrewazores:
Code looks good other than minor comments above.

The upload cancellation for these "legacy" requests doesn't seem like it should be too hard to do with [`xhr.abort()`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort), but the question is more around how do we pass that through from the component to the `xhr` without directly exposing the `xhr` in the function return value from `Api.service`. The `AbortController`/`AbortSignal` from the `fetch` API seems like a reasonable enough design to reuse or at least take inspiration from. There could be an optional argument for an abort signaller that components pass to the `Api.service` function they're calling, and if they invoke a function on the signaller then it simply follows through to `xhr.abort()`.

For example, the signaller could be an `Observable` that the component passes into the `Api.service` function call. Inside `Api.service`, if the signaller is defined, then `sendLegacyRequest` just adds a `signaller.subscribe(() => xhr.abort())` or so. The component could have a `const cancelUpload = new Subject<void>()` that it passes as `cancelUpload.asObservable()` to the `Api.service` function, and in order to cancel the upload the component just does `cancelUpload.next()` (in response to a cancellation button click for example).

andrewazores:
Looks really good.

I ran into a minor bug:

1. Select 5 valid files to upload
2. Set browser devtools network tab to throttle connection to something rather slow
3. Wait for some files to successfully upload, but not all
4. Press Cancel button and confirm cancellation
5. Upload modal shows ex. 3/5 complete and 2/5 cancelled
6. Press Submit button again
7. Allow upload to finish
8. Uploaded files in archive table shows all 5/5 files as expected, but also the original completed 3/5 have repeated copies

image

tthvo:
> Looks really good.
> 
> I ran into a minor bug:
> 
>     1. Select 5 valid files to upload
> 
>     2. Set browser devtools network tab to throttle connection to something rather slow
> 
>     3. Wait for some files to successfully upload, but not all
> 
>     4. Press Cancel button and confirm cancellation
> 
>     5. Upload modal shows ex. 3/5 complete and 2/5 cancelled
> 
>     6. Press Submit button again
> 
>     7. Allow upload to finish
> 
>     8. Uploaded files in archive table shows all 5/5 files as expected, but also the original completed 3/5 have repeated copies
> 
> 
> image

Ahh I think right now re-submit will upload all defined files regardless of previous states. Would you rather have the aborted ones re-submitted and ignore the previously successful ones?

andrewazores:
Yea exactly, I think that would be best.

src/app/Agent/AgentProbeTemplates.tsx
```diff
@@ -148,12 +148,9 @@ export const AgentProbeTemplates: React.FunctionComponent<AgentProbeTemplatesPro
   const handleDelete = React.useCallback(
     (template: ProbeTemplate) => {
       addSubscription(
-        context.api
-          .deleteCustomProbeTemplate(template.name)
-          .pipe(first())
-          .subscribe(() => {
-            /** Do nothing. Notifications hook will handle */
-          })
+        context.api.deleteCustomProbeTemplate(template.name).subscribe(() => {
```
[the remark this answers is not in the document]
tthvo: Ahh `deleteCustomProbeTemplate` calls `first` internally already so we don't have to call it here again.

tthvo:
> Yea exactly, I think that would be best.

New changes include these behaviours for re-submitting:

1. Only resubmit files that have errors (i.e. httpError, abortError, any others). If additional files are selected, new files + errored files will be submitted.
2. If the user removes errored files, any previously successful + additionally selected files will be submitted.

Case 2 is a little tricky. I am thinking if they remove all errored files, leave previously successful files and click submit again, we can assume they want to reupload those successfully uploaded files?

tthvo:
> I think I would only assume the user wants to re-upload files if they close the modal, re-open it, and re-select the same files again. To me the modal is a context of its own and the submit button should just try to get each file from "selected for upload" to "finished upload". Re-submitting is just a retry for failures. If the user removes all failures and leaves only successes listed then I think the submit button should behave as if there are no files selected at all, since in either case there are no files that need to progress toward the "finished upload" state.

Make sense to me! I updated so that all already uploaded files will show 'Already uploaded' state now instead of re-uploading.

andrewazores:
> Re-submitting is just a retry for failures.

On that note... maybe individual retry buttons for each failed file would make sense, and Submit would retry all failures.

That's an enhancement to think about later though. The user can already simulate that in a sense by closing and re-opening the modal and trying each failed file separately, if they ever run into a situation where this might help somehow.

tthvo:
> On that note... maybe individual retry buttons for each failed file would make sense, and Submit would retry all failures.

> That's an enhancement to think about later though. The user can already simulate that in a sense by closing and re-opening the modal and trying each failed file separately, if they ever run into a situation where this might help somehow.

Ahh this would be very nicee! We might need to see what Patternfly design guideline suggests for this. I open a new issue to track this at #720.

andrewazores:
If I get myself into this state:

image

Should the `Submit` button still be active? Clicking it once again leads to this:

image

And repeated clicks from here do nothing. I think the button should either disable or maybe the modal should just close.

maxcao13:
One thing that I'm just wondering for UX, just a thought, does it make sense for the modal to close if the requests are successful, but modal doesn't close if they aren't? My thoughts as user is that it may be jarring for the upload to be so quick and close, that I'm not sure if the requests were successful or not unless I try to find the recordings in the table or look at all the notifications. Thoughts?

andrewazores:
Yea, I was thinking about that too. If the files to be uploaded are large and/or the connection to the Cryostat instance is slow, then the user might want to just get up from the computer or switch to another tab and come back later. If the dialog remains open then they can definitely see that everything succeeded, but if it disappears then they may wonder if it just silently failed.

tthvo:
Hmm how about if all recordings are successful, we can replace the `Submit` and `Cancel` button with a button called 'Close'?

src/app/Rules/RulesUploadModal.tsx
```diff
@@ -67,87 +67,89 @@ export const parseRule = (file: File): Observable<Rule> => {
...
-    );
-  }, [context.api, notifications, setUploading, uploadFile, rejected, handleClose]);
+      );
+    },
+    [setUploading, context.api, handleClose, addSubscription, setAllOks]
+  );
 
   const handleAbort = React.useCallback(() => {
-    abort.abort();
+    // FIXME: Properly abort requests
```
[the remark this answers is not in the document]
tthvo: Opps totally forgot to remove these.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -690,51 +748,64 @@ export class ApiService {
...
-        const body = new window.FormData();
-        body.append('recording', file);
-        body.append('labels', JSON.stringify(labels));
+  uploadRecording(
+    file: File,
+    labels: Object,
+    onUploadProgress?: (progress: number) => void,
+    abortSignal?: Observable<void>
+  ): Observable<string> {
+    window.onbeforeunload = (event: BeforeUnloadEvent) => event.preventDefault();
```
andrewazores, line 757: Should the other upload functions have this too?

src/app/Rules/RulesUploadModal.tsx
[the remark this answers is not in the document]
tthvo: Removed now.

src/app/Shared/Services/Api.service.tsx
tthvo: Missed that too. Added back :D

PR #699, opened 2022-11-30
fix(labeler): fix labeler config file name

maxcao13:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] Signed the last commit: `git commit --amend --signoff`
_______________________________________________

Fixes: #696 

## Description of the change:
Fix labeler config file to be named correctly.

## Motivation for the change:
This is because it was named like `.labeler.yml` when it should be `labeler.yml`

## How to manually test:
Look at https://github.com/maxcao13/cryostat-web/pull/4 and notice that the action succeeds.
HOWEVER, I'm not sure if it actually works because it doesn't not apply any labels but I'm not sure if that's because it is on a fork. There are some issues within the action labeler repo that talk about how this doesn't work on forks.

Issue #592, opened 2022-10-28
[Story] "Test connection" feature


andrewazores:
As a user, I do not want to have to update a connection configuration, attempt to load a content view, and then delete and recreate my connection configuration if I entered the wrong values by mistake.

- [ ] #593
- [x] #594

tthvo:
Looks like we need to define an API endpoint to open a probe connection to the target. Just wondering if we can safely use any existing endpoint for this purpose?

andrewazores:
I think reusing the existing custom targets definition endpoint makes sense. I would probably call the parameter `dryrun` or something, but `probe` could work too - though we already use that term in relation to the JMC Agent and inserting bytecode.

That endpoint handler is already basically set up to do what we need it to, since it will attempt to open a connection to the target and retrieve its hash ID before adding it to the discovery database. So the `probe/dryrun` parameter would just skip that last step. It looks like the handler also needs to notify the JVM ID cache to invalidate the ID entry if the definition is not actually added to the database, or maybe instead a way to retrieve the JVM ID without caching the result to begin with.

PR #700, opened 2022-12-01
chore(requests): allow skipping status code check

tthvo:
# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/cryostat/pull/1257
Depends on #699 (updating ci)

## Description of the change:

This added a parameter to `sendRequest` function in Api.services to allow skipping status code check (i.e. internal error handler).

## Motivation for the change:

Since https://github.com/cryostatio/cryostat/pull/1257, target creation now allows `dryrun`. Auth failure (i.e. status 427) during `dryrun` should not trigger Auth modal, which is the default behaviour. This flag will allow disable this behaviour if needed.

## How to manually test:

Not yet used anywhere until #592.

### DIFF
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -866,7 +866,8 @@ export class ApiService {
     apiVersion: ApiVersion,
     path: string,
     config?: RequestInit,
-    suppressNotifications = false
+    suppressNotifications = false,
+    skipStatusCheck = false
   ): Observable<Response> {
     const req = () =>
       this.login.getHeaders().pipe(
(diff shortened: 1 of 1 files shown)


src/app/Shared/Services/Api.service.tsx
```diff
@@ -866,7 +866,8 @@ export class ApiService {
     apiVersion: ApiVersion,
     path: string,
     config?: RequestInit,
-    suppressNotifications = false
+    suppressNotifications = false,
+    skipStatusCheck = false
```
andrewazores, line 870: What if instead of `skipStatusCheck = false` this was `errorHandler?: Function`?

Further down,

```typescript
if (!errorHandler) {
    return this.handleError<Response>(err, req, suppressNotifications);
} else {
    errorHandler.apply(err, req);
}
```

tthvo: Yeh I was thinking about this but I decided to just throw the error and handle it outside sendRequest like below.

```typescript
testTarget(connectUrl: string, username?: string, password?: string) {
  return this.sendRequest(...).pipe(catchError(error) => { /* do something */ });
}
```

Just a thought to keep it clean and as close to "common practice". What do you think?

andrewazores: Sure, that makes sense too.

PR #726, opened 2022-12-08
build(deps): fix peer deps, upgrade react router

andrewazores:
https://github.com/remix-run/react-router/discussions/8753

current state: deeps properly upgraded, application builds and works. tests very broken


tthvo:
Hey @andrewazores, I am working on upgrading to React 18 for #1110. But run into lots of type issues:

```
Route' cannot be used as a JSX component.
  Its instance type 'Route<{}, string>' is not a valid JSX element.
    The types returned by 'render()' are incompatible between these types.
      Type 'import("/home/thvo/workspace/cryostat-web/node_modules/@types/react-router/node_modules/@types/react/index").ReactNode' is not assignable to type 'React.ReactNode'.ts(2786)
```

Might need to upgrade to Router v6 before that. Would u like me to continue this? Not sure if I can push to ur fork but I will open a new one with these commits.

andrewazores:
Ah, that looks annoying. I haven't had time to look any deeper into this and see what the upgrade path looks like or what else may be involved, but if you're interested to continue working on this then I would really appreciate it. We're starting to fall fairly far behind on upgrading this fairly major dependencies.

tthvo:
Sounds good! Thanks! I will continuue on this.

tthvo:
Work continue on #1143.

Issue #728, opened 2022-12-08
[Task] Make Dashboard cards dynamic stack layout


andrewazores:
![image](https://user-images.githubusercontent.com/3787464/206549097-e3f557be-8627-4298-8ee6-9cc86e92af43.png)

Currently, the Dashboard contains one hard-coded card. The Dashboard should behave similar to the following:

https://www.patternfly.org/v4/demos/card-view

Since the cards we display are likely to be larger and contain meaningful content views (analysis reports, charts) they will be larger than the mini cards in the demo. For a first pass, let's implement a stack layout:

https://www.patternfly.org/v4/layouts/stack

with the "add a new card" element at the bottom of the stack, and new cards added to the top of the stack. Each card should have an action menu in the top right corner with a Remove action that removes that card from the stack. For simplicity, the "add a new card" should allow the user to add or re-add any possible card to the stack, even if it would be a duplicate. The contents of the stack, ie the types of cards added and their order, should be persisted so that the same Dashboard layout is restored each time the user opens the application.

PR #731, opened 2022-12-08
feat(dashboard): dynamic stack layout

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #728

## Description of the change:
Implements a dynamic stack layout of cards on the Dashboard.

## Motivation for the change:
Allows users to choose their own layout according to their workflow and preferences.

* [x] Remove hardcoded analysis report card from dashboard
* [x] Add dynamic stack layout to dashboard
* [x] Add "add a new card" card
* [x] User can re-add analysis report card to dashboard
* [x] User can remove analysis report card from dashboard
* [x] Layout is persisted and restored


src/app/Dashboard/DashboardCardActionMenu.tsx
```diff
@@ -0,0 +1,64 @@
...
+    (_) => {
+      setOpen(false);
+    },
+    [setOpen]
+  );
+
+  return (
+    <>
+      <Dropdown isPlain isOpen={isOpen} toggle={<KebabToggle onToggle={setOpen} />} onSelect={onSelect}>
+        <DropdownItem key="Remove" onClick={props.onRemove}>
```
tthvo, line 58: I noticed this pretty odd dot right next to the action menu:

dropdown-dot

I think Dropdown is not composable and requires a dropDownItems prop. Also, we might want to use flip feature and open the menu from the right since it is "right-sided" action menu.


```diff
+  const dropdowmItems = React.useMemo(
+    () => [
+      <DropdownItem key="Remove" onClick={props.onRemove}>
+        Remove
+      </DropdownItem>,
+    ],
+    [props.onRemove]
+  );
+
   return (
     <>
-      <Dropdown isPlain isOpen={isOpen} toggle={<KebabToggle onToggle={setOpen} />} onSelect={onSelect}>
-        <DropdownItem key="Remove" onClick={props.onRemove}>
-          Remove
-        </DropdownItem>
-      </Dropdown>
+      <Dropdown
+        isPlain
+        isFlipEnabled
+        menuAppendTo={'parent'}
+        position={'right'}
+        isOpen={isOpen}
+        toggle={<KebabToggle onToggle={setOpen} />}
+        onSelect={onSelect}
+        dropdownItems={dropdowmItems}
+      />
     </>
```

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -755,7 +760,7 @@ export const AutomatedAnalysisCard: React.FunctionComponent<AutomatedAnalysisCar
   }, [usingArchivedReport, usingCachedReport, report, isLoading, errorMessage]);
 
   return (
-    <Card id="automated-analysis-card" isRounded isCompact isExpanded={isCardExpanded}>
+    <Card id="automated-analysis-card" isRounded {...props} isExpanded={isCardExpanded}>
       <CardHeader
         isToggleRightAligned
```
tthvo, line 765: Since now we have action menu on the right, do you think we should move the expanded arrow to the left?

Something like this: https://www.patternfly.org/v4/components/card#expandable-cards

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -766,6 +771,7 @@ export const AutomatedAnalysisCard: React.FunctionComponent<AutomatedAnalysisCar
           'aria-expanded': isCardExpanded,
         }}
       >
+        <CardActions>{...props.actions || []}</CardActions>
```
tthvo, line 774: I guess we should do `...(props.actions || [])` since props.action can be `undefined`?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
tthvo: We might need to adjust the card title tho since it shifts everything to left now.

How about this?

Screenshot from 2022-12-08 17-52-11

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
andrewazores: I think that's what I wrote originally actually, but `yarn format:apply` seems to set it back to how it is now.

https://www.typescriptlang.org/play?ssl=1&ssc=1&pln=2&pc=1#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBBCcQQ0MwAUVWDEQC0gADVHBQGNJ3KAALygABEAAkYNAMOB4GRonzFBTBPB3AERcwABS0+mM9ysygc9wASnJjVl8sVSoCpoA2gBdHUAbmlBrlkAV8GVpoCXG4kGSGUgrgAPr7QJabcEVDhmh4yF5EOLnMD2DQvm4zMx2AxmJB+LBoOpEhz4HoRF4+XSGUzBHykti+eAHKgAOTCABykAkfKCKgA8l8yO9BBmtJAjqhHApXNI8KAnNGdMMsPoUAhENDhJdExJxdERAYONiFkxWBweEESkA

It seems that the TS compiler handles this in a typesafe way anyway, though it definitely doesn't look right reading it.

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
tthvo:
```diff
diff --git a/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx b/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
index 7591686..14aa2fb 100644
--- a/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
+++ b/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
@@ -76,6 +76,7 @@ import {
   Checkbox,
   Grid,
   GridItem,
+  Label,
   LabelGroup,
   Split,
   SplitItem,
@@ -90,7 +91,7 @@ import {
   ToolbarItem,
   Tooltip,
 } from '@patternfly/react-core';
-import { OutlinedQuestionCircleIcon, Spinner2Icon, TrashIcon } from '@patternfly/react-icons';
+import { InfoCircleIcon, OutlinedQuestionCircleIcon, Spinner2Icon, TrashIcon } from '@patternfly/react-icons';
 import * as React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { filter, first, map, tap } from 'rxjs';
@@ -745,24 +746,17 @@ export const AutomatedAnalysisCard: React.FunctionComponent<AutomatedAnalysisCar
       return undefined;
     }
     return (
-      <div
-        style={{
-          color: 'var(--pf-global--Color--200)',
-          fontWeight: 'lighter',
-          fontSize: '0.90em',
-          fontFamily: 'inherit',
-        }}
-      >
-        {`${usingArchivedReport ? 'Archived' : usingCachedReport ? 'Cached' : 'Active'} report name:`}{' '}
-        <span style={{ color: 'var(--pf-global--Color--100)', fontWeight: 'bold', fontStyle: 'italic' }}>{report}</span>
-      </div>
+      <>
+        <Label icon={<InfoCircleIcon />} color={'cyan'}>
+          {`${usingArchivedReport ? 'Archived' : usingCachedReport ? 'Cached' : 'Active'} report name=${report}`}
+        </Label>
+      </>
     );
   }, [usingArchivedReport, usingCachedReport, report, isLoading, errorMessage]);
 
   return (
     <Card id="automated-analysis-card" isRounded {...props} isExpanded={isCardExpanded}>
       <CardHeader
-        isToggleRightAligned
         onExpand={onCardExpand}
         toggleButtonProps={{
           id: 'automated-analysis-toggle-details',
@@ -771,13 +765,13 @@ export const AutomatedAnalysisCard: React.FunctionComponent<AutomatedAnalysisCar
           'aria-expanded': isCardExpanded,
         }}
       >
-        <CardActions>{...props.actions || []}</CardActions>
-        <CardTitle component="h4">
-          <Split>
-            <SplitItem isFilled>Automated Analysis</SplitItem>
+        <CardTitle>
+          <Split hasGutter>
+            <SplitItem>Automated Analysis</SplitItem>
             <SplitItem>{reportSource}</SplitItem>
           </Split>
         </CardTitle>
+        <CardActions>{...props.actions || []}</CardActions>
       </CardHeader>
       <CardExpandableContent>
         <Stack hasGutter>
```

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
tthvo: Ahh interesting. Thanks for explaning. Agreed it was super odd XD

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
tthvo: On this note, do you think we should move the caret of `TargetSelect` to the left too?

tthvo:
I am thinking if we can allow users to see dashboard cards individually in 'full-screen' mode similar to Grafana. Then, the user can hit `Go back` (i.e. breadcrumbs) to go back main dashboard.

Screenshot from 2022-12-08 18-13-35

 What do you think?

tthvo:
We might also need to support `back-to-top` since the dashboard can grow pretty large.

https://www.patternfly.org/v4/components/back-to-top/react-demos

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -766,9 +763,10 @@ export const AutomatedAnalysisCard: React.FunctionComponent<AutomatedAnalysisCar
...
         <CardTitle component="h4">
-          <Split>
-            <SplitItem isFilled>Automated Analysis</SplitItem>
+          <Split hasGutter>
```
tthvo, line 768: Maybe we should move the label out of the card title to avoid title-styles applied to the label.

It would look like this. Seems to align better and follow Patternfly guide.

Screenshot from 2022-12-08 18-35-49



```diff
diff --git a/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx b/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
index 98e69b4..275b308 100644
--- a/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
+++ b/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
@@ -78,6 +78,7 @@ import {
   GridItem,
   Label,
   LabelGroup,
+  Level,
   Split,
   SplitItem,
   Stack,
@@ -764,12 +765,10 @@ export const AutomatedAnalysisCard: React.FunctionComponent<AutomatedAnalysisCar
         }}
       >
         <CardActions>{...props.actions || []}</CardActions>
-        <CardTitle component="h4">
-          <Split hasGutter>
-            <SplitItem>Automated Analysis</SplitItem>
-            <SplitItem>{reportSource}</SplitItem>
-          </Split>
-        </CardTitle>
+        <Level hasGutter>
+          <CardTitle component="h4">Automated Analysis</CardTitle>
+          {reportSource}
+        </Level>
       </CardHeader>
       <CardExpandableContent>
         <Stack hasGutter>
```

src/app/Dashboard/Dashboard.tsx
```diff
@@ -36,20 +36,82 @@
...
 
+export interface DashboardCardProps {
+  actions?: JSX.Element[];
+}
+
+export const DashboardCards: CardConfig[] = [
+  {
+    component: AutomatedAnalysisCard,
+    props: {
+      isLarge: true,
```
tthvo, line 61: Look like if we set this to true. There are some misalignments between toolbar and content in term of padding in the automatic analysis card.

Screenshot from 2022-12-08 18-54-34


We guess we can keep `isCompact` or add more padding to the toolbar (i.e. maybe around 1rem)?

maxcao13 (commented):
I think it would be nice if: 

after selecting a card type from the dropdown, the + icon becomes blue and you could press the button to add the card as well. As a user, that's my instinct on what the design would be able to do anyways.

src/app/Dashboard/Dashboard.tsx
```diff
@@ -36,20 +36,82 @@
...
   return (
     <TargetView pageTitle="Dashboard" compactSelect={false} hideEmptyState>
-      <Grid hasGutter>
-        <GridItem xl2={8} xl={10} lg={10} md={12} sm={12}>
-          <AutomatedAnalysisCard />
-        </GridItem>
-      </Grid>
+      <Stack hasGutter>
+        {cardNames.map(getConfigByName).map((cfg, idx) => (
+          <StackItem key={idx}>
```
tthvo, line 104: Since we are aiming to support re-ordering, I think we should not use the index as key, but instead something that is unique the card. Something like `${cardName}-${duplicate-version}` or its hash code?

maxcao13:
Also, as of now having multiple AA cards has a global filter applied to both at the same time. I'm not sure what the best thing to do is here, but I would sort of expect that filters would be per card, which of course I did implement so if you think it's a good idea, I can refactor so that the filters would be applied per card instead of globally.

src/app/Dashboard/Dashboard.tsx
```diff
@@ -36,20 +36,82 @@
...
+  for (const choice of DashboardCards) {
+    if (choice.component.name === name) {
+      return choice;
+    }
+  }
+  throw new Error(`Unknown card type selection: ${name}`);
+}
+
 export const Dashboard: React.FunctionComponent<DashboardProps> = (props) => {
+  const [cardNames, setCardNames] = React.useState(getFromLocalStorage('DASHBOARD_CFG', []));
```
tthvo, line 76: Maybe save this in Redux store?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -740,24 +746,15 @@ export const AutomatedAnalysisCard: React.FunctionComponent<AutomatedAnalysisCar
...
-      </div>
+      <Label icon={<InfoCircleIcon />} color={'cyan'}>
+        {`${usingArchivedReport ? 'Archived' : usingCachedReport ? 'Cached' : 'Active'} report name=${report}`}
+      </Label>
     );
   }, [usingArchivedReport, usingCachedReport, report, isLoading, errorMessage]);
 
   return (
-    <Card id="automated-analysis-card" isRounded isCompact isExpanded={isCardExpanded}>
+    <Card id="automated-analysis-card" isRounded {...props} isExpanded={isCardExpanded}>
```
maxcao13, line 756: On that note, these ids would have to be generated to be unique somehow to both adhere to HTML and to help with having per-card filters.

tthvo (commented):
> I think it would be nice if:
> after selecting a card type from the dropdown, the + icon becomes blue and you could press the button to add the card as well. As a user, that's my instinct on what the design would be able to do anyways.

On this one, I look around and see this pretty nice design.

https://support.nutshell.com/hc/article_attachments/4403174325911/Jun-24-2021_13-06-26.gif (please ignore everything after selecting the card type)

We basically has a dotted border card as `Add` which opens a modal to choose which type of card we want. This way, we can have some description, icons, and search filters for the users to choose from.

Could be an improvement over the flat list of card types. What do you think?

src/app/Dashboard/AddCard.tsx
```diff
@@ -0,0 +1,117 @@
...
+}
+
+export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardProps) => {
+  const [selection, setSelection] = React.useState('None');
+  const [selectOpen, setSelectOpen] = React.useState(false);
+
+  const options = React.useMemo(() => {
+    return [
+      <SelectOption key={0} isPlaceholder value="None" />,
+      ...DashboardCards.map((choice, idx) => <SelectOption key={idx + 1} value={choice.component.name} />),
```
maxcao13, line 65: I think I would prefer having the add card title menu options to have spaces, so just adding a field like "cardTitle" to `CardConfig` and mapping the `value` to `{choice.cardTitle}` I think would look nicer.

maxcao13:
> > I think it would be nice if:
> > after selecting a card type from the dropdown, the + icon becomes blue and you could press the button to add the card as well. As a user, that's my instinct on what the design would be able to do anyways.
> 
> On this one, I look around and see this pretty nice design.
> 
> https://support.nutshell.com/hc/article_attachments/4403174325911/Jun-24-2021_13-06-26.gif (please ignore everything after selecting the card type)
> 
> We basically has a dotted border card as `Add` which opens a modal to choose which type of card we want. This way, we can have some description, icons, and search filters for the users to choose from.
> 
> Could be an improvement over the flat list of card types. What do you think?

That does look very nice as an improvement.

andrewazores:
> I am thinking if we can allow users to see dashboard cards individually in 'full-screen' mode similar to Grafana. Then, the user can hit `Go back` (i.e. breadcrumbs) to go back main dashboard. I think we can implement this in the card action menu.
> 
> Screenshot from 2022-12-08 18-13-35
> 
> What do you think?

Sure, that's an interesting idea. Worth filing a stretch goal issue for.

src/app/Dashboard/Dashboard.tsx
andrewazores: I'll leave that for later since my current definition of the "grid" is really just a reflowing of a stack, but it could become a real grid with x/y coordinates too.

andrewazores:
> Also, as of now having multiple AA cards has a global filter applied to both at the same time. I'm not sure what the best thing to do is here, but I would sort of expect that filters would be per card, which of course I did implement so if you think it's a good idea, I can refactor so that the filters would be applied per card instead of globally, after this pr is merged.

I'll give it some thought. I don't know if it makes sense to have multiple of this kind of card on the dashboard, at least not in a long term configuration. Having a second instance in a transient way because the user is changing the configuration (using a different event template maybe) so they add a new card with the new config before deleting the old one should be supported I think, so I don't want to enforce a cap on the number of instances, but adding extra logic to enable different instances of this card to store different internal configs might just be more work and more maintenance burden for a usecase that doesn't actually matter to users in the end.

src/app/Dashboard/AddCard.tsx
andrewazores: Ah yea, I was thinking of that at some point but forgot to actually do it :-)

andrewazores:
> > I think it would be nice if:
> > after selecting a card type from the dropdown, the + icon becomes blue and you could press the button to add the card as well. As a user, that's my instinct on what the design would be able to do anyways.
> 
> On this one, I look around and see this pretty nice design.
> 
> https://support.nutshell.com/hc/article_attachments/4403174325911/Jun-24-2021_13-06-26.gif (please ignore everything after selecting the card type)
> 
> We basically has a dotted border card as `Add` which opens a modal to choose which type of card we want. This way, we can have some description, icons, and search filters for the users to choose from.
> 
> Could be an improvement over the flat list of card types. What do you think?

Something like that is my eventual goal. I was picturing it as the AddCard's content would change to reveal a list of card types, then change again to allow any configurations specific to that selected card type. Along the lines of

https://www.patternfly.org/v4/components/wizard

But, since we only have the one kind of card so far, I didn't want to get too far ahead on this design since I'm not quite certain yet what kind of configuration options we might need to make available.

src/app/Dashboard/AddCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+            </EmptyState>
+          </Bullseye>
+        </Card>
+      </>
+    );
+  } else {
+    return (
+      <>
+        <Wizard onClose={handleStop} onSave={handleAdd}>
+          <WizardStep id="card-type-select" name="Card Type">
```
tthvo, line 130: Looks very good with the wizard :D 

Just noticed that our current patternfly core does not fully support composition pattern yet

> If you seek a wizard solution that allows for more composition, see the [React next](https://www.patternfly.org/v4/components/wizard/react-next) tab.

Seems like the last step should have `Finish` as the accept but now its showing Next. Do you think we should still keep this or refactor to use `steps` props on `Wizard`?

tthvo:
> I'll give it some thought. I don't know if it makes sense to have multiple of this kind of card on the dashboard, at least not in a long term configuration. Having a second instance in a transient way because the user is changing the configuration (using a different event template maybe) so they add a new card with the new config before deleting the old one should be supported I think, so I don't want to enforce a cap on the number of instances, but adding extra logic to enable different instances of this card to store different internal configs might just be more work and more maintenance burden for a usecase that doesn't actually matter to users in the end.

On this note, I noticed if you have 2 AA cards, these scenarios happen:

1. If you have 2 cards using the same configuration (i.e. using default), delete one does not update the others.

Screenshot from 2022-12-09 14-05-26


3. Start fresh with no recordings. Choose default for one card, create report. Then, try to create another report using tempalte=Continuous on another. It fails with error:

Screenshot from 2022-12-09 14-08-01

maxcao13:
> > I'll give it some thought. I don't know if it makes sense to have multiple of this kind of card on the dashboard, at least not in a long term configuration. Having a second instance in a transient way because the user is changing the configuration (using a different event template maybe) so they add a new card with the new config before deleting the old one should be supported I think, so I don't want to enforce a cap on the number of instances, but adding extra logic to enable different instances of this card to store different internal configs might just be more work and more maintenance burden for a usecase that doesn't actually matter to users in the end.
> 
> On this note, I noticed if you have 2 AA cards, these scenarios happen:

If we don't have multiple of this card, then I guess disabling the ability to add it twice would be nice, and also maybe some indicator on the wizard that it is unique.

andrewazores:
I added some description text so the user knows what they're selecting without having to just add it and see what it does. In that text I suggest that the card should be unique. I still haven't enforced anything to prevent the user from creating duplicates - maybe I'm wrong and it should be possible to add more of them. I think if the uniqueness enforcement is desired then it should at least wait until after the cards can be reordered.

src/app/Dashboard/AddCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const [showWizard, setShowWizard] = React.useState(false);
+  const [selection, setSelection] = React.useState(DashboardCards[0].title);
+  const [selectOpen, setSelectOpen] = React.useState(false);
+  const dispatch = useDispatch<StateDispatch>();
+
+  const options = React.useMemo(() => {
+    return [
+      ...DashboardCards.map((choice, idx) => (
+        <SelectOption key={idx} value={choice.component.name}>
+          {choice.title}
```
tthvo, line 73: Not sure what happened but seems to run into error if I select an option for card type.

tthvo: I think the value should be `choice.title`?

andrewazores: Hmm, can't seem to reproduce this. Maybe try clearing site data for `localhost` so the localstorage is reset?

src/app/Dashboard/AddCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+    setShowWizard(false);
+  }, [setShowWizard]);
+
+  return (
+    <>
+      <Card isRounded isLarge>
+        {showWizard ? (
+          <Wizard onClose={handleStop} onSave={handleAdd}>
+            <WizardStep id="card-type-select" name="Card Type" footer={{ nextButtonText: 'Finish' }}>
+              <Stack hasGutter>
```
tthvo, line 113: Maybe we can use the footer prop on `SelectOption` to achieve this. Just looks a little nicer and more compact. Also, the user can see the description without selecting one.

Screenshot from 2022-12-09 15-04-39

src/app/Dashboard/AddCard.tsx
tthvo: Yehh still happens if I clear localStorage

```
Uncaught Error: Unknown card type selection: AutomatedAnalysisCard
getConfigByTitle Dashboard.tsx:91
AddCard AddCard.tsx:120
React 19
```

tthvo: I am on the last commit on upstream now at [b0ea246](https://github.com/cryostatio/cryostat-web/pull/731/commits/b0ea2466c4b85bff3f564784f2f8fe4147548dfe)

andrewazores: Oh wait, I triggered it now. I just didn't quite understand what you meant.

src/app/Dashboard/AddCard.tsx
```diff
@@ -0,0 +1,145 @@
...
```
tthvo, line 113: For better visual and following guideline, do you think we should make this content into a form? Then, we can have some sort of form title?

andrewazores: I could do it just so that the selection box and description are styled like they would be within a form context but this isn't really a form, so I don't know if that's necessary. I can try it and see what it looks like I guess.

andrewazores: The `FormSelectOption` component doesn't seem to have a nice way to include the short description within the `Select`, and putting a `Form` within the `WizardStep` seems to make the container scrollable... I'm not sure this is worth the effort.

tthvo: Ahh I was thinking of something like this:

Screenshot from 2022-12-09 15-52-54

Hmm now that you mention, its a little too extra?

andrewazores: https://www.patternfly.org/v4/components/wizard/design-guidelines

I guess the Wizard example does state that there should be an additional title, which may just repeat what the step title is. I think that's maybe enough.

Right now the wizard could easily just be a form instead, but once there's a Step 2 of configuring the card then the configuration options shown will be specific to each card type and so I think a wizard is the right design pattern.

tthvo: I am trying these changes and looks pretty much achieving the result:

```diff
diff --git a/src/app/Dashboard/AddCard.tsx b/src/app/Dashboard/AddCard.tsx
index c0556e9..28a1c05 100644
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -44,6 +44,8 @@ import {
   EmptyStateBody,
   EmptyStateIcon,
   EmptyStateVariant,
+  Form,
+  FormGroup,
   Select,
   SelectOption,
   Stack,
@@ -110,16 +112,25 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
         {showWizard ? (
           <Wizard onClose={handleStop} onSave={handleAdd}>
             <WizardStep id="card-type-select" name="Card Type" footer={{ nextButtonText: 'Finish' }}>
-              <Stack hasGutter>
-                <StackItem>
-                  <Select onToggle={handleToggle} isOpen={selectOpen} onSelect={handleSelect} selections={selection}>
-                    {options}
-                  </Select>
-                </StackItem>
-                <StackItem>
-                  <Text>{getConfigByTitle(selection).descriptionFull}</Text>
-                </StackItem>
-              </Stack>
+              <Form>
+                <FormGroup label="Select a card type" isRequired>
+                  <Stack hasGutter>
+                    <StackItem>
+                      <Select
+                        onToggle={handleToggle}
+                        isOpen={selectOpen}
+                        onSelect={handleSelect}
+                        selections={selection}
+                      >
+                        {options}
+                      </Select>
+                    </StackItem>
+                    <StackItem>
+                      <Text>{getConfigByTitle(selection).descriptionFull}</Text>
+                    </StackItem>
+                  </Stack>
+                </FormGroup>
+              </Form>
             </WizardStep>
           </Wizard>
         ) : (
```

andrewazores: That looks a lot like I tried and got the weird scrolling behaviour. I'll try that again, maybe I did something a little wrong/different.

andrewazores: Oh - I removed the stack, you have it within the form. Maybe that's what the difference was.

maxcao13:
Should the wizard have a dropdown state for None selected? So that the user can explicitly set the Card they want before adding it?

tthvo (commented):
Looking great to me!!

Might need some tests for Dashboard but it will likely change in next PRs. So, maybe after closing parent issue then.

Will approve after resolving https://github.com/cryostatio/cryostat-web/pull/731#issuecomment-1344782901.

maxcao13 (approved):
Looks great to me!

Issue #727, opened 2022-12-08
[Story] Dashboard contents/layout should be configurable


andrewazores:
As of #618 the Dashboard view contains one hardcoded card with an automated analysis report feature. Now that there is possible card-based content to be displayed here, this content should be made configurable so that other card features described in #468 can be added and configured by the user.

- [x] #728 
- [x] #729
- [x] #730
- [x] #866 
- [x] #764
- [x] #765
- [x] #739

andrewazores:
Someone had the idea that the Add card wizard should display a little preview of the card types to the user so that they can have a better sense of what each card type is/does/looks like without having to step all the way through creating each one.

https://www.patternfly.org/v4/components/tile

or perhaps

https://www.patternfly.org/v4/components/card#selectable-cards

could be good components to use to replace the current `Select` on the wizard's card type selection step. This way the card types can be displayed as a gallery of miniature preview cards so the user can scan them all at a glance, rather than clicking through each Select value to see the previews.

Issue #730, opened 2022-12-08
[Task] Allow drag-and-drop reordering of cards in Dashboard grid


andrewazores:
After #729, the grid should become dynamic and the order of the cards should be mutable. By grabbing the card header the user should be able to drag a card in between or overtop of another card and drop it there, which will move the dragged card from its original position into the new position where it was dropped, and shift all cards to close the gap if needed. These layout mutations should also be persisted.

Issue #729, opened 2022-12-08
[Task] Extend Dashboard layout to grid with user-defined card sizes


andrewazores:
After #728, the Dashboard layout system should be further extended to a Grid:

https://www.patternfly.org/v4/layouts/grid

When a user adds a card to the grid it should have a default size, specified by the type of card being added. This default should be the minimum size that the card can reasonably render and display its content. The top right action menu should have a Resize action that allows the user to change the size of the card and expand it, or bring it back down to the default/minimum. The default value should be indicated in the resize dialog content. This size should also be persisted in the dashboard layout. The card layout will still be modelled as a stack or list, but the rendering on screen will allow smaller cards to be rendered side-by-side, forming rows, instead of forcing a vertical stack orientation.

PR #747, opened 2022-12-12
feat(dashboard): card configuration

andrewazores:
# Welcome to Cryostat! 👋

Related to #727

## Description of the change:
This change adds a step to the Add card on the Dashboard. The new step builds a dynamic form for configuring
the selected card type. This form becomes `props` passed to the card(s) when created. This configuration is
stored in the Redux store and persisted to localstorage along with the previous implementation's simple linear
layout of card names, so the configuration props are also persistent.

## Motivation for the change:
This allows more flexibility for card implementations with multiple instances since the instances can be
distinguished simply by props rather than needing to implement more complex partitioning of the Redux store.
This will also be useful for configuration of card sizes in the dynamic grid layout. For the Grafana cards
we have discussed, this system could be used for configuring which specific Grafana chart should be displayed
within the card.

image

image

### DIFF
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -115,7 +115,6 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
   const handleAdd = React.useCallback(() => {
     setShowWizard(false);
     const cardTitle = getConfigByTitle(selection).component.name;
-    console.info('creating card', { cardTitle, ...propsConfig });
     dispatch(addCardIntent(cardTitle, propsConfig));
   }, [setShowWizard, dispatch, addCardIntent, selection, propsConfig]);
 
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -153,7 +153,7 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
     <>
       <Card isRounded isLarge>
         {showWizard ? (
-          <Wizard onClose={handleStop} onSave={handleAdd} height={640} nav={customNav}>
+          <Wizard onClose={handleStop} onSave={handleAdd} height={'30rem'} nav={customNav}>
             <WizardStep
               id="card-type-select"
               name="Card Type"
@@ -305,9 +305,7 @@ const PropsConfigForm = (props: PropsConfigFormProps) => {
           );
           break;
         case 'select':
-          input = (
-            <SelectControl handleChange={handleChange(ctrl.key)} config={props.config[ctrl.key]} control={ctrl} />
-          );
+          input = <SelectControl handleChange={handleChange(ctrl.key)} config={props.config} control={ctrl} />;
           break;
         default:
           input = <Text>Bad config</Text>;
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -269,12 +269,7 @@ const PropsConfigForm = (props: PropsConfigFormProps) => {
       switch (ctrl.kind) {
         case 'boolean':
           input = (
-            <Switch
-              label="On"
-              labelOff="Off"
-              isChecked={props.config[ctrl.key]}
-              onChange={handleChange(ctrl.key)}
-            />
+            <Switch label="On" labelOff="Off" isChecked={props.config[ctrl.key]} onChange={handleChange(ctrl.key)} />
           );
           break;
         case 'number':
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -138,6 +138,7 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
             <WizardStep id="card-props-config" name="Configuration" footer={{ nextButtonText: 'Finish' }}>
               {selection ? (
                 <PropsConfigForm
+                  cardTitle={selection}
                   initialState={propsConfig}
                   controls={getConfigByTitle(selection).propControls}
                   onChange={setPropsConfig}
@@ -170,6 +171,7 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
 };
 
 interface PropsConfigFormProps {
+  cardTitle: string;
   controls: PropControl[];
   initialState: any;
   onChange: ({}) => void;
@@ -268,7 +270,9 @@ const PropsConfigForm = (props: PropsConfigFormProps) => {
     <>
       {props.controls.length > 0 ? (
         <Form>
-          <FormGroup label="Configure the card">{props.controls.map((ctrl) => createControl(ctrl))}</FormGroup>
+          <FormGroup label={`Configure the ${props.cardTitle} card`}>
+            {props.controls.map((ctrl) => createControl(ctrl))}
+          </FormGroup>
         </Form>
       ) : (
         <Text>No configuration required.</Text>
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -124,7 +124,7 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
     <>
       <Card isRounded isLarge>
         {showWizard ? (
-          <Wizard onClose={handleStop} onSave={handleAdd} height={400}>
+          <Wizard onClose={handleStop} onSave={handleAdd} height={500}>
             <WizardStep id="card-type-select" name="Card Type" footer={{ isNextDisabled: !selection }}>
               <Form>
                 <FormGroup label="Select a card type" isRequired isStack>
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -251,60 +251,63 @@ const PropsConfigForm = (props: PropsConfigFormProps) => {
     [props, props.onChange, setPropsConfig]
   );
 
-  const createControl = React.useCallback((ctrl: PropControl): JSX.Element => {
-    let input: JSX.Element;
-    switch (ctrl.kind) {
-      case 'boolean':
-        input = <Switch label={ctrl.name} isChecked={propsConfig[ctrl.key]} onChange={handleChange(ctrl.key)} />;
-        break;
-      case 'number':
-        input = (
-          <NumberInput
-            inputName={ctrl.name}
-            inputAriaLabel={`${ctrl.name} input`}
-            value={propsConfig[ctrl.key]}
-            onChange={handleNumeric(ctrl.key)}
-            onPlus={handleNumericStep(ctrl.key, 1)}
-            onMinus={handleNumericStep(ctrl.key, -1)}
-          />
-        );
-        break;
-      case 'string':
-        input = (
-          <TextInput
-            type="text"
-            aria-label={`${ctrl.key} input`}
-            value={propsConfig[ctrl.key]}
-            onChange={handleChange(ctrl.key)}
-          />
-        );
-        break;
-      case 'text':
-        input = (
-          <TextArea
-            type="text"
-            aria-label={`${ctrl.key} input`}
-            value={propsConfig[ctrl.key]}
-            onChange={handleChange(ctrl.key)}
-          />
-        );
-        break;
-      default:
-        input = <Text>Bad config</Text>;
-        break;
-    }
-    return (
-      <FormGroup
-        key={`${ctrl.key}}`}
-        label={ctrl.kind == 'number' ? ctrl.name : undefined}
-        helperText={ctrl.description}
-        isInline
-        isStack
-      >
-        {input}
-      </FormGroup>
-    );
-  }, [propsConfig, handleChange, handleNumeric, handleNumericStep]);
+  const createControl = React.useCallback(
+    (ctrl: PropControl): JSX.Element => {
+      let input: JSX.Element;
+      switch (ctrl.kind) {
+        case 'boolean':
+          input = <Switch label={ctrl.name} isChecked={propsConfig[ctrl.key]} onChange={handleChange(ctrl.key)} />;
+          break;
+        case 'number':
+          input = (
+            <NumberInput
+              inputName={ctrl.name}
+              inputAriaLabel={`${ctrl.name} input`}
+              value={propsConfig[ctrl.key]}
+              onChange={handleNumeric(ctrl.key)}
+              onPlus={handleNumericStep(ctrl.key, 1)}
+              onMinus={handleNumericStep(ctrl.key, -1)}
+            />
+          );
+          break;
+        case 'string':
+          input = (
+            <TextInput
+              type="text"
+              aria-label={`${ctrl.key} input`}
+              value={propsConfig[ctrl.key]}
+              onChange={handleChange(ctrl.key)}
+            />
+          );
+          break;
+        case 'text':
+          input = (
+            <TextArea
+              type="text"
+              aria-label={`${ctrl.key} input`}
+              value={propsConfig[ctrl.key]}
+              onChange={handleChange(ctrl.key)}
+            />
+          );
+          break;
+        default:
+          input = <Text>Bad config</Text>;
+          break;
+      }
+      return (
+        <FormGroup
+          key={`${ctrl.key}}`}
+          label={ctrl.kind == 'number' ? ctrl.name : undefined}
+          helperText={ctrl.description}
+          isInline
+          isStack
+        >
+          {input}
+        </FormGroup>
+      );
+    },
+    [propsConfig, handleChange, handleNumeric, handleNumericStep]
+  );
 
   return (
     <>
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -55,7 +55,14 @@ import {
   TextInput,
   Title,
 } from '@patternfly/react-core';
-import { Wizard, WizardStep } from '@patternfly/react-core/dist/js/next';
+import {
+  CustomWizardNavFunction,
+  Wizard,
+  WizardControlStep,
+  WizardNav,
+  WizardNavItem,
+  WizardStep,
+} from '@patternfly/react-core/dist/js/next';
 import { PlusCircleIcon } from '@patternfly/react-icons';
 import { useDispatch } from 'react-redux';
 import { StateDispatch } from '@app/Shared/Redux/ReduxStore';
@@ -121,12 +128,38 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
     setShowWizard(false);
   }, [setShowWizard]);
 
+  // custom nav for disabling subsequent steps (ex. configuration) if a card type hasn't been selected first
+  const customNav: CustomWizardNavFunction = React.useCallback(
+    (
+      isExpanded: boolean,
+      steps: WizardControlStep[],
+      activeStep: WizardControlStep,
+      goToStepByIndex: (index: number) => void
+    ) => {
+      return (
+        <WizardNav isExpanded={isExpanded}>
+          {steps.map((step, idx) => (
+            <WizardNavItem
+              key={step.id}
+              id={step.id}
+              content={step.name}
+              isCurrent={activeStep.id === step.id}
+              isDisabled={step.isDisabled || (idx > 0 && !selection)}
+              stepIndex={step.index}
+              onNavItemClick={goToStepByIndex}
+            />
+          ))}
+        </WizardNav>
+      );
+    },
+    [selection]
+  );
+
   return (
-    // FIXME wizard navigation by clicking left-side step names should be disabled if the first step selection is empty
     <>
       <Card isRounded isLarge>
         {showWizard ? (
-          <Wizard isStepVisitRequired onClose={handleStop} onSave={handleAdd} height={500}>
+          <Wizard isStepVisitRequired onClose={handleStop} onSave={handleAdd} height={500} nav={customNav}>
             <WizardStep id="card-type-select" name="Card Type" footer={{ isNextDisabled: !selection }}>
               <Form>
                 <FormGroup label="Select a card type" isRequired isStack>
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -78,16 +78,6 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
   const [selectOpen, setSelectOpen] = React.useState(false);
   const dispatch = useDispatch<StateDispatch>();
 
-  React.useEffect(() => {
-    const c = {};
-    if (selection) {
-      for (const ctrl of getConfigByTitle(selection).propControls) {
-        c[ctrl.key] = ctrl.defaultValue;
-      }
-    }
-    setPropsConfig(c);
-  }, [getConfigByTitle, selection, setPropsConfig]);
-
   const options = React.useMemo(() => {
     return [
       <SelectOption key={0} value={'None'} isPlaceholder />,
@@ -108,8 +98,16 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
     (_, selection, isPlaceholder) => {
       setSelection(isPlaceholder ? '' : selection);
       setSelectOpen(false);
+
+      const c = {};
+      if (selection) {
+        for (const ctrl of getConfigByTitle(selection).propControls) {
+          c[ctrl.key] = ctrl.defaultValue;
+        }
+      }
+      setPropsConfig(c);
     },
-    [setSelection, setSelectOpen]
+    [setSelection, setSelectOpen, getConfigByTitle, setPropsConfig]
   );
 
   const handleAdd = React.useCallback(() => {
@@ -196,7 +194,7 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
               {selection && (
                 <PropsConfigForm
                   cardTitle={selection}
-                  selectedPropConfig={propsConfig}
+                  config={propsConfig}
                   controls={getConfigByTitle(selection).propControls}
                   onChange={setPropsConfig}
                 />
@@ -237,37 +235,37 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
 interface PropsConfigFormProps {
   cardTitle: string;
   controls: PropControl[];
-  selectedPropConfig: any;
+  config: any;
   onChange: ({}) => void;
 }
 
 const PropsConfigForm = (props: PropsConfigFormProps) => {
   const handleChange = React.useCallback(
     (k) => (e) => {
-      const copy = { ...props.selectedPropConfig };
+      const copy = { ...props.config };
       copy[k] = e;
       props.onChange(copy);
     },
-    [props.selectedPropConfig, props.onChange]
+    [props.config, props.onChange]
   );
 
   const handleNumeric = React.useCallback(
     (k) => (e) => {
       const value = (e.target as HTMLInputElement).value;
-      const copy = { ...props.selectedPropConfig };
+      const copy = { ...props.config };
       copy[k] = value;
       props.onChange(copy);
     },
-    [props.selectedPropConfig, props.onChange]
+    [props.config, props.onChange]
   );
 
   const handleNumericStep = React.useCallback(
     (k, v) => (e) => {
-      const copy = { ...props.selectedPropConfig };
-      copy[k] = props.selectedPropConfig[k] + v;
+      const copy = { ...props.config };
+      copy[k] = props.config[k] + v;
       props.onChange(copy);
     },
-    [props.selectedPropConfig, props.onChange]
+    [props.config, props.onChange]
   );
 
   const createControl = React.useCallback(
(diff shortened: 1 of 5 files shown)


src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -791,3 +791,33 @@ export const AutomatedAnalysisCard: React.FunctionComponent<AutomatedAnalysisCar
...
+  component: AutomatedAnalysisCard,
+  propControls: [
+    {
+      name: 'Compact Style',
+      key: 'isCompact',
+      description: 'Apply PatternFly compact Card styling to reduce padding',
+      kind: 'boolean',
+      defaultValue: true,
+    },
+    // TODO remove one or both of these, they are useful for development testing but not for a user
```
maxcao13, line 814: I assume later, the actual AutomatedAnalysis config would be displayed here some sort of prop controls. Maybe having a boolean switch button that says "Use default config" and then when it gets switched to "off", then the AutomatedAnalysisConfig drawer may show up and somehow give config props back to the `AddCard` parent component for the Card.

maxcao13:
I'm not sure how relevant this is, but I was thinking during the CardType config step where you select a Card type, it may be useful if users can see a preview image of the card type they are selecting? Just some generic default image, not a full pre-rendered preview. There is a lot of white space there right now so that might be good as a way for users to just have a nicer experience as well as use up some of the space.

src/app/Dashboard/AddCard.tsx
```diff
@@ -46,28 +46,41 @@ import {
...
   const [selection, setSelection] = React.useState('');
+  const [propsConfig, setPropsConfig] = React.useState({});
   const [selectOpen, setSelectOpen] = React.useState(false);
   const dispatch = useDispatch<StateDispatch>();
 
+  React.useEffect(() => {
+    const c = {};
+    if (selection) {
+      for (const ctrl of getConfigByTitle(selection).propControls) {
+        c[ctrl.key] = ctrl.defaultValue;
```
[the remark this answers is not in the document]
andrewazores: I'm not sure what you mean...

src/app/Dashboard/AddCard.tsx
```diff
@@ -94,47 +107,48 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
     setShowWizard(true);
   }, [setShowWizard]);
 
   const handleStop = React.useCallback(() => {
+    setSelection('');
     setShowWizard(false);
   }, [setShowWizard]);
 
   return (
+    // FIXME wizard navigation by clicking left-side step names should be disabled if the first step selection is empty
```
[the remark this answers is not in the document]
andrewazores: Done now.

src/app/Dashboard/AddCard.tsx
```diff
@@ -94,47 +107,48 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+    console.info('creating card', { cardTitle, ...propsConfig });
+    dispatch(addCardIntent(cardTitle, propsConfig));
+  }, [setShowWizard, dispatch, addCardIntent, selection, propsConfig]);
 
   const handleStart = React.useCallback(() => {
     setShowWizard(true);
   }, [setShowWizard]);
 
   const handleStop = React.useCallback(() => {
+    setSelection('');
```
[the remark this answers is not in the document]
andrewazores: Good catch

src/app/Dashboard/AddCard.tsx
```diff
@@ -94,47 +107,48 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
 
   const handleAdd = React.useCallback(() => {
     setShowWizard(false);
-    dispatch(addCardIntent(getConfigByTitle(selection).component.name));
-  }, [dispatch, selection]);
+    const cardTitle = getConfigByTitle(selection).component.name;
+    console.info('creating card', { cardTitle, ...propsConfig });
```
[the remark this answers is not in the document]
andrewazores: Nope, removed

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
andrewazores: Makes sense. The configuration setup right now is much more basic than that and only allows configurations to be added that generate a simple form with simple controls, so the "default config" switch is already doable but embedding the config drawer is not yet possible. It could be done though, and in that particular instance I think it shouldn't be too hard since that component already uses Redux rather than needing to pass state back to the form or to the card component props.

andrewazores:
> I'm not sure how relevant this is, but I was thinking during the CardType config step where you select a Card type, it may be useful if users can see a preview image of the card type they are selecting? Just some generic default image, not a full pre-rendered preview. There is a lot of white space there right now so that might be good as a way for users to just have a nicer experience as well as use up some of the space.
> 
> I understand this is still sort of a beta view so, this doesn't have to be a thing, especially when this is likely to change.

Interesting idea. I imagine that with #729 / #730 that the Add card in particular would probably default to a somewhat smaller size with less whitespace, but maybe not. It's worth exploring this concept.

src/app/Dashboard/Dashboard.tsx
```diff
@@ -97,27 +153,27 @@ export function getConfigByTitle(title: String): CardConfig {
...
   const dispatch = useDispatch<StateDispatch>();
-  const cardNames = useSelector((state: RootState) => state.dashboardConfigs.list);
+  const cardConfigs = useSelector((state: RootState) => state.dashboardConfigs.list);
 
   const handleRemove = React.useCallback(
     (idx: number) => {
       dispatch(deleteCardIntent(idx));
     },
-    [dispatch, cardNames]
+    [dispatch, cardConfigs]
```
maxcao13, line 162: I don't think the `cardConfigs` is needed here.

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +203,120 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+        const copy = { ...old };
+        copy[k] = old[k] + v;
+        props.onChange(copy);
+        return copy;
+      });
+    },
+    [props, props.onChange, setPropsConfig]
+  );
+
+  const createControl = (ctrl: PropControl): JSX.Element => {
```
maxcao13, line 254: React.useMemo ?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
andrewazores: On this note, I'm thinking of adding an optional parameter to the `DashboardCardDescriptor` like `advancedConfig?: JSX.Element`. If this is present then the AddCard Wizard would either add a third config step, or maybe combine the basic config and advanced config into substeps of the config step, or something along these lines. The config element would be something like the AA card's drawer or form that can just be embedded here, and like before, it would be expected to persist state into redux/localstorage rather than being used to specify props.

maxcao13: Substeps make sense. 

If we have three config steps, it would make me think that the two different config steps are for different functions. For example step 2 only configures the look of the card, and step 3 only configures the data that backs that card.

src/app/Dashboard/Dashboard.tsx
```diff
@@ -36,21 +36,29 @@
...
-  component: React.FunctionComponent;
-  props?: React.PropsWithChildren<any>;
+  component: React.FunctionComponent<any>;
+  propControls: PropControl[];
+}
+
+export interface PropControl {
+  name: string;
+  key: string;
+  description: string;
```
tthvo, line 59: Should this `description` prop be optional? Same for `defaultValue`?

src/app/Dashboard/Dashboard.tsx
```diff
@@ -36,21 +36,29 @@
...
+  kind: 'boolean' | 'number' | 'string' | 'text';
```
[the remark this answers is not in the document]
tthvo: I grab this one from the internal React module

```typscript
type ReactNode = ReactElement | string | number | ReactFragment | ReactPortal | boolean | null | undefined;
```

src/app/Dashboard/Dashboard.tsx
```diff
@@ -59,25 +67,73 @@ export interface DashboardCardProps {
   actions?: JSX.Element[];
 }
 
-export const DashboardCards: CardConfig[] = [
+// TODO remove this
```
tthvo, line 70: To remove?

src/app/Dashboard/Dashboard.tsx
andrewazores: I don't think so, it seems like these should be required properties for the generated form.

src/app/Dashboard/Dashboard.tsx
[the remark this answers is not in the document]
andrewazores: That would fit in with the `advancedConfig?: JSX.Element` idea I was talking about in the other comment thread, I think. This is meant only for encoding the basic card props form controls, so a generic React component node doesn't fit in with the others here which all map to ex. `TextInput`, `Switch`

src/app/Dashboard/AddCard.tsx
```diff
@@ -94,47 +114,74 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
     <>
       <Card isRounded isLarge>
         {showWizard ? (
-          <Wizard onClose={handleStop} onSave={handleAdd} height={300}>
-            <WizardStep
-              id="card-type-select"
-              name="Card Type"
-              footer={{ nextButtonText: 'Finish', isNextDisabled: !selection }}
-            >
+          <Wizard isStepVisitRequired onClose={handleStop} onSave={handleAdd} height={500} nav={customNav}>
```
tthvo, line 161: I think we can remove `isStepVisitRequired`. This allows the Configuration (Step 2) to be enabled if a valid card type is selected. Placeholder selection will disable this step 2.

src/app/Dashboard/AddCard.tsx
```diff
@@ -46,28 +46,48 @@ import {
...
 
+  React.useEffect(() => {
+    const c = {};
+    if (selection) {
+      for (const ctrl of getConfigByTitle(selection).propControls) {
+        c[ctrl.key] = ctrl.defaultValue;
+      }
+    }
+    setPropsConfig(c);
+  }, [props, getConfigByTitle, selection, setPropsConfig]);
```
tthvo, line 89: We can remove `props` as deps here?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -791,3 +787,19 @@ export const AutomatedAnalysisCard: React.FunctionComponent<AutomatedAnalysisCar
...
+    `,
+  descriptionFull: `
+Creates a recording and periodically evalutes various common problems in application configuration and performance.
+Results are displayed with scores from 0-100 with colour coding and in groups.
+This card should be unique on a dashboard.
+      `,
+  component: AutomatedAnalysisCard,
+  propControls: [],
+  // FIXME this form gets embedded within a form, and should not have its own independent create/save controls
+  advancedConfig: <AutomatedAnalysisConfigForm isSettingsForm={false}></AutomatedAnalysisConfigForm>,
```
tthvo, line 804: This needs a a wrapper around this form to add a switch to choose default or custom settings?

andrewazores: That's part of what I meant about this component not being properly set up right now. There should be a switch to hide or reveal these controls, but they also need to be refactored so that they can be used separately/outside of a `<Form>`. The `Create` button probably shouldn't be there and the `Save Configuration` should probably just happen every time a change occurs rather than requiring the user to click a button so that they aren't having to save within the config screen and then also separately press `Finish` on the Wizard. Or, the `Finish` button from the wizard needs to be hooked into the `Save Configuration` action as well. Seems out of scope of this particular PR.

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +232,123 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+      );
+    },
+    [propsConfig, handleChange, handleNumeric, handleNumericStep]
+  );
+
+  return (
+    <>
+      {props.controls.length > 0 ? (
+        <Form>
+          <FormGroup label={`Configure the ${props.cardTitle} card`}>
```
tthvo, line 345: `createControl` already returns the FormGroup, so do we still need another outter one here?

I guess the label should be apply to each FormGroup returned from `createControl` instead?

andrewazores: The outer formgroup provides the topmost "title" within the step, and the inner formgroups provide labels to the individual input controls. I can place the title a different way.

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +233,124 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+
+interface PropsConfigFormProps {
+  cardTitle: string;
+  controls: PropControl[];
+  initialState: any;
+  onChange: ({}) => void;
+}
+
+const PropsConfigForm = (props: PropsConfigFormProps) => {
+  const [propsConfig, setPropsConfig] = React.useState(props.initialState);
```
tthvo, line 245: Looks like we are having selected propConfig as a state here and its parent.

I think we could pass in a prop as `selectedPropConfig` in `PropsConfigFormProps` from parent and remove this state. This way, we can reduce re-rendering and limit "source of truth"?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
tthvo: Ahh right makes sense to me! Just a little odd looking at the template menu is set as required but wizard is allowed to finish. We will fix this later then.

src/app/Dashboard/Dashboard.tsx
```diff
@@ -97,27 +155,27 @@ export function getConfigByTitle(title: String): CardConfig {
 
 export const Dashboard: React.FunctionComponent<DashboardProps> = (props) => {
   const dispatch = useDispatch<StateDispatch>();
-  const cardNames = useSelector((state: RootState) => state.dashboardConfigs.list);
+  const cardConfigs = useSelector((state: RootState) => state.dashboardConfigs.list);
```
tthvo, line 158: Are we planning to have these configurations specific for each target? I think right now its pretty much applying to all targets?

tthvo (commented):
Looks great! Just some more questions above :D

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +233,124 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+    [props.onChange, setPropsConfig]
+  );
+
+  const createControl = React.useCallback(
+    (ctrl: PropControl): JSX.Element => {
+      let input: JSX.Element;
+      switch (ctrl.kind) {
+        case 'boolean':
+          input = (
+            <Switch label={ctrl.name} isReversed isChecked={propsConfig[ctrl.key]} onChange={handleChange(ctrl.key)} />
```
tthvo, line 290: I think the `label/labelOff` on `Switch` are messages for when the switch is on/off? The `ctrl.name` should be part of the FormGroup label?

https://www.patternfly.org/v4/components/switch#switch

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +233,124 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+          );
+          break;
+        default:
+          input = <Text>Bad config</Text>;
+          break;
+      }
+      return (
+        <FormGroup
+          key={`${ctrl.key}}`}
+          label={ctrl.kind !== 'boolean' ? ctrl.name : undefined}
```
tthvo, line 332: On this note https://github.com/cryostatio/cryostat-web/pull/747/files#r1046563123, do you think we might need the form label for switch?

src/app/Dashboard/AddCard.tsx
tthvo: Also, no reverse too? Just thinking we could keep it similar to Openshift console.


Screenshot from 2022-12-12 20-08-05

src/app/Dashboard/AddCard.tsx
```diff
@@ -46,28 +46,48 @@ import {
...
 interface AddCardProps {}
 
 export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardProps) => {
   const [showWizard, setShowWizard] = React.useState(false);
   const [selection, setSelection] = React.useState('');
+  const [propsConfig, setPropsConfig] = React.useState({});
   const [selectOpen, setSelectOpen] = React.useState(false);
   const dispatch = useDispatch<StateDispatch>();
 
+  React.useEffect(() => {
```
tthvo, line 81: Looks like its called for each selection to set default. Could we move this into `handleSelect` so that state changes are batched?

src/app/Dashboard/AddCard.tsx
tthvo: Screenshot from 2022-12-12 20-27-52

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
tthvo: On this note, we can have an expandable section like this?

Opened:

Screenshot from 2022-12-12 20-37-12

Closed:

Screenshot from 2022-12-12 20-36-58

src/app/Dashboard/Dashboard.tsx
andrewazores: Correct, it applies to all targets. That's by design. Perhaps this can be extended with multiple dashboard layouts, but the layouts would be identified by name rather than being associated with specific targets.

src/app/Dashboard/AddCard.tsx
andrewazores: Oh right, thanks. These things were all pretty broken before when I had the nested formgroup setup, but now it works properly.

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +233,124 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+              inputName={ctrl.name}
+              inputAriaLabel={`${ctrl.name} input`}
+              value={propsConfig[ctrl.key]}
+              onChange={handleNumeric(ctrl.key)}
+              onPlus={handleNumericStep(ctrl.key, 1)}
+              onMinus={handleNumericStep(ctrl.key, -1)}
+            />
+          );
+          break;
+        case 'string':
```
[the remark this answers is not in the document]
andrewazores: I added support in the last couple of commits. The SelectOptions can be populated with a static compile-time array, or an `Observable<T[]>` that adds all the options as a batch possibly async, or an `Observable<T>` that adds the options from an async stream. This should be useful for things like the Grafana charts cards where presumably one of the prop configs would be the ID or URL for the chart to display, which obviously isn't known until after some async API call at runtime.

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
andrewazores: I hesitate to add it as an expandable section since the substep functionality is right there already, so it seems like that would make things feel less cluttered and cramped into the wizard which is already boxed in inside a non-fullscreen card.

Right now it's implemented as a progressive wizard rather than substeps but it shouldn't be *too* hard to update that. The custom nav function that allows for the config steps to be optional just needs to do the same thing for substeps as well, I think.

FWIW I think most cards will only need to use the first "basic props" config step anyway, since the only options would be things like a chart or metric name/ID and then the sizing properties (ex. grid width). Most of the kinds of cards I envision just do and present one kind of information and just repeat that function for the current selected target, so they may not have any configuration at all. The AA card is a bit of a special case since it has to carry config for starting recordings too. Maybe we need something vaguely similar for the Grafana charts for uploading recordings to `jfr-datasource`.

maxcao13 (approved):
Looks good to me!

src/app/Dashboard/Dashboard.tsx
tthvo: Right that makes sense! Thanks for exlaining :D

src/app/Dashboard/AddCard.tsx
```diff
@@ -77,64 +89,121 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+    const cardTitle = getConfigByTitle(selection).component.name;
+    dispatch(addCardIntent(cardTitle, propsConfig));
+  }, [setShowWizard, dispatch, addCardIntent, selection, propsConfig]);
 
   const handleStart = React.useCallback(() => {
     setShowWizard(true);
   }, [setShowWizard]);
 
   const handleStop = React.useCallback(() => {
+    setSelection('');
```
tthvo, line 119: Should also clean up the `propsConfig` state?

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +226,166 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+    addSubscription(
+      obs.subscribe({
+        next: (v) =>
+          setOptions((old) => {
+            if (Array.isArray(v)) {
+              return v;
+            }
+            return [...old, v];
+          }),
+        complete: () => {},
```
tthvo, line 371: Any need from this callback? Maybe we should handle error instead?

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +226,166 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+              type="text"
+              aria-label={`${ctrl.key} input`}
+              value={props.config[ctrl.key]}
+              onChange={handleChange(ctrl.key)}
+            />
+          );
+          break;
+        case 'select':
+          input = (
+            <SelectControl handleChange={handleChange(ctrl.key)} config={props.config[ctrl.key]} control={ctrl} />
```
tthvo, line 309: I think `config` prop should just be `props.config`? Right now, selecting something from menu does not render the selected option.

src/app/Dashboard/AddCard.tsx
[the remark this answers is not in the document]
tthvo: That's actually very niceeely done! Thanks :D

src/app/Dashboard/AddCard.tsx
```diff
@@ -77,64 +89,121 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+    },
+    [selection]
+  );
 
   return (
     <>
       <Card isRounded isLarge>
         {showWizard ? (
-          <Wizard onClose={handleStop} onSave={handleAdd} height={300}>
+          <Wizard onClose={handleStop} onSave={handleAdd} height={640} nav={customNav}>
```
tthvo, line 156: It's 640px right? Should we have some more "adaptive" unit?

andrewazores: It's px yes. I'll change it to use `rem`.

src/app/Dashboard/AddCard.tsx
andrewazores: No, it's not needed - I'll remove it.

Not sure what to do for error handling. Probably nothing here. I figure this is probably going to be an Observable returned by an `Api.service` method which will already handle the error and display a notification. I guess I could wipe out the existing options and add a fake option that just says `Error` or something?

src/app/Dashboard/AddCard.tsx
```diff
@@ -77,64 +89,122 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
   const handleStart = React.useCallback(() => {
     setShowWizard(true);
   }, [setShowWizard]);
 
   const handleStop = React.useCallback(() => {
     setShowWizard(false);
-  }, [setShowWizard]);
+    setSelection('');
+    setPropsConfig({});
+  }, [setSelection, setShowWizard]);
```
tthvo, line 122: Need prop deps for `setPropsConfig`?

src/app/Dashboard/AddCard.tsx
tthvo: > I guess I could wipe out the existing options and add a fake option that just says Error or something?

Or i guess we could just wipe out and left as empty. Or maybe show some custom compact error view instead of the select component should work too.

tthvo: But probably later since lots of things could change...

src/app/Shared/Redux/DashboardConfigReducer.tsx
```diff
@@ -40,15 +40,20 @@ import { getFromLocalStorage } from '@app/utils/LocalStorage';
...
 const initialState = {
-  list: getFromLocalStorage('DASHBOARD_CFG', []) as string[],
+  list: getFromLocalStorage('DASHBOARD_CFG', []) as CardConfig[],
 };
 
 export const dashboardConfigReducer = createReducer(initialState, (builder) => {
   builder
     .addCase(addCardIntent, (state, { payload }) => {
-      state.list.splice(payload.idx || 0, 0, payload.name);
+      state.list.splice(0, 0, payload);
```
tthvo, line 56: Ahh one thing i was meant to ask: Seems like we are adding it to the front. Should we add to the end instead?

It's just closer to the `AddCard` so the user can see them being "appended". If we have a long stack of cards, the user might wonder where the added card goes since current scrolling position is kept.

andrewazores: Sure, makes sense in this stack layout and easy enough to change.

src/app/Dashboard/AddCard.tsx
tthvo: Oh yeh just saw the new commit. Looking good :D

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +227,169 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+      })
+    );
+  }, [props.control, props.control.values, of, addSubscription, setOptions, setErrored]);
+
+  return (
+    <Select
+      onToggle={setSelectOpen}
+      isOpen={selectOpen}
+      onSelect={handleSelect}
+      selections={props.config[props.control.key]}
```
tthvo, line 386: Should this take into account error? 

For example, previously users choose a valid selection (i.e. still fetching options) - async0. But later the fetch failed but the selected holder is till the previously valid one (async0) tho the list of option only contains error.

tthvo: If so, we might need to do clean up current propConfig to reset to default.

andrewazores: I think regardless of attempts to clean up, if an error occurs when populating options on this form then the result is going to be an invalid card configuration. If anything the wizard should probably lock and prevent the user from moving forward, so they have to cancel out and clear the state that way.

andrewazores: That probably also ties in to a need for this wizard to allow form configurations that have required fields and validations too, and if any fields are unfilled/invalid/erorred then the wizard should lock. I think a higher priority than adding all this nice error handling and cleanup is to focus instead on adding card implementations that can even generate these invalid/incomplete configs and errors ;-)

src/app/Dashboard/AddCard.tsx
```diff
@@ -157,3 +227,166 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
+              type="text"
+              aria-label={`${ctrl.key} input`}
+              value={props.config[ctrl.key]}
+              onChange={handleChange(ctrl.key)}
+            />
+          );
+          break;
+        case 'select':
+          input = (
+            <SelectControl handleChange={handleChange(ctrl.key)} config={props.config[ctrl.key]} control={ctrl} />
```
tthvo, line 310: Should be named `selectedConfig`?

andrewazores: Not sure which part you're referring to

src/app/Dashboard/AddCard.tsx
tthvo: Ah right makes sense :D We can refactor and add these in the future.

src/app/Dashboard/AddCard.tsx
tthvo: Ah sorry the prop `config`?

andrewazores: Err... the prop's name is `config`

https://github.com/cryostatio/cryostat-web/pull/747/files#diff-05851ec53c3e2fc6b4eaf23d7057435b91f0fde619760e9c319bd79ba0fe05f2R340

tthvo: Ahhh I mean since that prop is used in `selections` prop of the `Select` menu, maybe we should called it `selectedConfig`?

tthvo (approved):
Looks great to me :D

Issue #725, opened 2022-12-08
[Task] Improve performance of persisting redux states to local storage


tthvo:
We are using `store.subscribe` to save redux states into local storage:

https://github.com/cryostatio/cryostat-web/blob/7be743715d8b23efde44892c8f65331d432a9773/src/app/Shared/Redux/ReduxStore.tsx#L62

This works but there are some issues:

1. Entire store state is persisted regardless of which reducer is called.
2. `JSON.parse` is a pretty expensive operation and should not be called too frequent.

We can explore this package as an alternative: https://www.npmjs.com/package/redux-persist

On this note, as our store grows (i.e. more actions, reducers, configs), we might also need a better organized directory structure under `@app/Shared/Redux`. Some useful readings: https://redux.js.org/faq/code-structure

tthvo:
> We can explore this package as an alternative: https://www.npmjs.com/package/redux-persist

Development process seems to be stopped for a long time. Looking into alternatives.

PR #753, opened 2022-12-13
perf(redux): use redux-persist to handle localStorage

tthvo:
# Welcome to Cryostat! 👋

Fixes: #725 
Depends on #747 (needs updates on dashboardCondfigReducer)

## Description of the change:

This PR includes following changes:
- [ ] Organized redux reducers and actions into slices. 
  - [ ] Each slice now belongs to "kind" directory.
  - [ ] Each slice consists of related actions and reducers.
  - [ ] ReduxStore.tsx nows exports all actions so that we can just import from there. This helps if in the futures, this redux folder changes.
- [ ] Use redux-persist to persist store states instead of raw localStorage save on entire store states.

## Motivation for the change:
- As redux store grows, the number of reducers and actions increase. Therefore, we need a more organized folder structure.
- As for state persisting, redux-persist is the recommended package to handle this. A few features from redux-persist:
    - Many supported storage types.
    - Allow blacklisting/whitelisting. 
    - Granular persisting instead of persisting entire store at any times like previously.


tthvo:
There was a little hiccup from `redux-persist` for tests:
- `_persist` key is automatically added to state, but is marked as required (https://github.com/rt2zz/redux-persist/pull/1170). Tests must add this with:

  ```typescript
  {
    _persist: {
      version: 1:
      rehydrated: false
    },
  ...
  }
  ```

There is a PR already opened to fix but has not been merged for long :((

andrewazores:
> There was a little hiccup from `redux-persist` for tests:
> 
>     * `_persist` key is automatically added to state, but is marked as required ([Make '_persist' key optional in state returned by persistReducer rt2zz/redux-persist#1170](https://github.com/rt2zz/redux-persist/pull/1170)). Tests must add this with:
>       ```ts
>       {
>         _persist: {
>           version: 1:
>           rehydrated: false
>         },
>       ...
>       }
>       ```
> 
> 
> That PR has not been merged for long :((

I'm a little concerned about this package given that that PR has been open so long, despite being so small and seemingly low risk. Development activity looks pretty dead in general. What if it stays dead and falls behind the latest Redux versions?

Maybe this is an opportunity to re-investigate usage of Redux as a whole, and find something comparable that also offers localStorage persistence?

https://github.com/pmndrs/zustand
https://docs.pmnd.rs/zustand/integrations/persisting-store-data

This looks like it operates from a very similar principle to Redux but probably even simpler, and comes with persistence support. Redux probably does better when you have large and complex global application state, but our application state is mostly on the backend actually and all the client state storage is (hopefully) small and simple UI state by design.

tthvo:
> As redux store grows, the number of reducers and actions increase. Therefore, we need a more organized folder structure.

After some readings and thoughts, based on our current project structure, its hard to follow their recommended structure so I chose the following suitable one for ours:

```
/app
--/Shared
----/Redux
------/KindOfState1
--------SpecificKindOfState1Slicer.tsx
--------SpecificKindOfState2Slicer.tsx
------/KindOfState2
--------SpecificKindOfState1Slicer.tsx
--------SpecificKindOfState2Slicer.tsx
```

What do you think?

tthvo:
> I'm a little concerned about this package given that that PR has been open so long, despite being so small and seemingly low risk. Development activity looks pretty dead in general. What if it stays dead and falls behind the latest Redux versions?

> Maybe this is an opportunity to re-investigate usage of Redux as a whole, and find something comparable that also offers localStorage persistence?

> https://github.com/pmndrs/zustand
> https://docs.pmnd.rs/zustand/integrations/persisting-store-data

> This looks like it operates from a very similar principle to Redux but probably even simpler, and comes with persistence support. Redux probably does better when you have large and complex global application state, but our application state is mostly on the backend actually and all the client state storage is (hopefully) small and simple UI state by design.

Right, there has not been any major dev since Oct 2021. I will look into this new package then. I guess we will move away from Redux then.

andrewazores:
Well, not necessarily. We could stick to Redux since that's quite well established and active, just avoid redux-persist. If there is nothing else out there similar then maybe there are lessons we can take away and implement small pieces of as utilities within our own project at least to improve the local storage performance.

tthvo:
> Well, not necessarily. We could stick to Redux since that's quite well established and active, just avoid redux-persist. If there is nothing else out there similar then maybe there are lessons we can take away and implement small pieces of as utilities within our own project at least to improve the local storage performance.

I think apart from redux-persist, I was exploring writing a custom Redux middleware to persist data but stopped after seeing that package. I could try seeing if I could replicate current behaviour with middleware.

andrewazores:
Frontend notifications could probably also be handled using Redux/Zustand/whatever, now that I think of it.

andrewazores:
https://jotai.org/

This one looks good as well, from a cursory glance.

andrewazores:
https://github.com/pmndrs/jotai/issues/13

tthvo:
@andrewazores Hey, with this latest commit, I tried to write a simple middleware that persists  part of the store based on the action.

What do you think? Meanwhile I will play around with other packages in case we want to migrate our redux setup to it.

andrewazores:
I think that middleware looks reasonable for achieving the goal you set out for (don't read/write the entire store when only one piece changed).

But, this is evolving the way I anticipated it would - it's a big application-wide global storage that lives outside of the component hierarchy and lifecycle, and then to give it better encapsulation and better performance characteristics for those components it ends up getting cut down into separate pieces that map pretty closely to components and their lifecycle within the hierarchy. In that situation we should question why there's such a global store being established if in the end we divide it up again. Separate pieces of state storage attached to the components do effectively the same thing, except with less extra management pieces that try to glob everything up together into one big chunk. I think that model probably makes more sense for larger applications using server-side rendering where there is just a lot more state, but here it seems like Redux is a lot of overkill. On the other hand, the previous system I set up using just `React.useState()`, Contexts, and direct `localStorage` isn't great either because there's a lot of hand-rolled code that means each component using persistence is doing it in its own way with its own implementation quirks.

I think that comment I linked earlier from the Jotai issues board sums up the situation very well. Zustand looks to me a lot like Redux in concept, but with a nicer and smoother API. Jotai looks a lot like `React.useState()`, and I really like that. I can imagine that our problems with duplicate AA cards on the dashboard would be solved quite easily by asking the user to give a unique title to the AA card when they're configuring it, and then doing `const cardConfigAtom = atomWithStorage(props.cardTitle, defaultConfigOptions); const [cardConfig, setCardConfig] = useAtom(cardConfigAtom)` . There needs to be a cleanup step to delete the config from storage when the card is removed from the dashboard, but that should be quite straightforward to do, too.

Not saying we should switch to Jotai necessarily. But, it looks to me like the Redux optimizations are tending in a direction like that, so it's at least a design to keep in mind either with this extra middleware for Redux, or for removing Redux and adding some "middleware" that is inspired by Jotai and goes back to simply wrapping around useState/localStorage.

andrewazores:
tl;dr:

1. I think this PR looks good, it just needs some cleaning up for review because there are a lot of apparent spurious changes (import order etc.) that make it hard to read, and `redux-persist` is still added as a build dep. In general I'm good with this approach.
2. If you're interested in continuing to explore this area then I would suggest prototyping with Jotai primarily, and do it in pieces. ie replace only the new dashboard card layout and props Redux store with Jotai, leaving Redux in place for the rest. If that proves to be a good fit for our needs then the replacement can be done piecewise like this. If it doesn't work out then we can evaluate the other options (back to plain handrolled useEffect+localStorage, or zustand, or mobx, etc)

tthvo:
Oh right, that makes sense. Originally, I use Redux to persist states cross route changes (i.e. /recordings, /dashboard) without reading from LocalStorage all the times. Seemed to overkill it. I will some migrating to Jotai and see if it works. Do you think it's something as stretch goal or TODO for 2.3.0?

tthvo:
Also, I remove the deps on redux-persist now. Some imports need fixing since the Redux directory now changes. But import orders are reverted so that eslint apply PR will do that later :D

tthvo:
Checking imports again. Seemed like there is some circular deps in imports/exports that causes filter state for `uploads` target to have

```typescript
[
  {
    target: 'uploads',
    active: { selectedCategory: 'Name', filters: undefined },
    archived: { selectedCategory: 'Name', filters: undefined }
  }
]

```

andrewazores:
> Oh right, that makes sense. Originally, I use Redux to persist states cross route changes (i.e. /recordings, /dashboard) without reading from LocalStorage all the times. Seemed to overkill it from reading your comments. I will start some slow migrating to Jotai and see if it works. Do you think it's something as stretch goal or TODO for 2.3.0?

It can be a stretch goal (doesn't need to get done) but if you want to start working on it now that sounds good.

src/app/Shared/Redux/Middlewares/PersistMiddleware.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const result = next(action);
+    // Extract new state here
+    const rootState = getState();
+    if (AutomatedAnalysisFilterActions.has(action.type)) {
+      saveToLocalStorage('AUTOMATED_ANALYSIS_FILTERS', rootState.automatedAnalysisFilters);
+    } else if (RecordingFilterActions.has(action.type)) {
+      saveToLocalStorage('TARGET_RECORDING_FILTERS', rootState.recordingFilters);
+    } else if (DashboardConfigActions.has(action.type)) {
+      saveToLocalStorage('DASHBOARD_CFG', rootState.dashboardConfigs);
+    }
```
andrewazores, line 59: Maybe put an `else` here and just `console.warn` something. Could be a pitfall for a developer trying to add something new and forgetting to update these cases.

tthvo:
Ready for review again... :D

In these latest commits:

- Resolve circular dependencies issue by moving all required "empty" state to redux slice files so that they are always initialized. Though, there actually still circular deps on interfaces but exported const are now resolved. We might need to be careful with future imports lint.
- Renamed action intent (i.e. action generator) to match specific slice kind.

src/app/Shared/Redux/Middlewares/PersistMiddleware.tsx
tthvo: Sure! Added it :D

andrewazores:
If I run a previous build of the web-client and it persists some data into `localStorage`, then run a build with this PR, I get a broken application (full white screen) and devtools console messages like:

```
Uncaught TypeError: cardConfigs is undefined
    Dashboard 71148:226
    React 21
    91832 npm.react-dom.79f07555ac3ff8c9.bundle.js:48
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    fn runtime.2846e04d228a8620.bundle.js:331
    <anonymous> React
    54142 npm.react-dom.79f07555ac3ff8c9.bundle.js:58
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    fn runtime.2846e04d228a8620.bundle.js:331
    React 2
    73116 npm.react-dom.79f07555ac3ff8c9.bundle.js:18
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    fn runtime.2846e04d228a8620.bundle.js:331
    <anonymous> React
    73935 npm.react-dom.79f07555ac3ff8c9.bundle.js:28
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    fn runtime.2846e04d228a8620.bundle.js:331
    <anonymous> 25796:3
    25796 app.55d1a74d04c08471.bundle.js:348
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    __webpack_exec__ app.55d1a74d04c08471.bundle.js:464
    <anonymous> app.55d1a74d04c08471.bundle.js:465
    O runtime.2846e04d228a8620.bundle.js:83
    <anonymous> app.55d1a74d04c08471.bundle.js:466
    webpackJsonpCallback runtime.2846e04d228a8620.bundle.js:1228
    <anonymous> app.55d1a74d04c08471.bundle.js:10
```

```
The above error occurred in the <Dashboard> component:

Dashboard@webpack-internal:///71148:219:76
Route@webpack-internal:///16550:647:29
RouteWithTitleUpdates@webpack-internal:///86887:215:21
Switch@webpack-internal:///16550:849:29
Suspense
LastLocationProvider@webpack-internal:///12556:161:47
C@webpack-internal:///16550:904:31
AppRoutes@webpack-internal:///86887:229:54
main
div
DrawerContentBody@webpack-internal:///16537:16:54
div
div
DrawerMain@webpack-internal:///97311:16:34
DrawerContent@webpack-internal:///78135:20:131
div
Drawer@webpack-internal:///5436:30:138
div
div
Page@webpack-internal:///16488:37:9
AppLayout@webpack-internal:///90428:103:20
Router@webpack-internal:///16550:266:30
BrowserRouter@webpack-internal:///73727:57:35
Provider@webpack-internal:///60682:14:18
App

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://reactjs.org/link/error-boundaries to learn more about error boundaries.
```

```
Uncaught TypeError: cardConfigs is undefined
    Dashboard 71148:226
    React 21
    91832 npm.react-dom.79f07555ac3ff8c9.bundle.js:48
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    fn runtime.2846e04d228a8620.bundle.js:331
    <anonymous> React
    54142 npm.react-dom.79f07555ac3ff8c9.bundle.js:58
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    fn runtime.2846e04d228a8620.bundle.js:331
    React 2
    73116 npm.react-dom.79f07555ac3ff8c9.bundle.js:18
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    fn runtime.2846e04d228a8620.bundle.js:331
    <anonymous> React
    73935 npm.react-dom.79f07555ac3ff8c9.bundle.js:28
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    fn runtime.2846e04d228a8620.bundle.js:331
    <anonymous> 25796:3
    25796 app.55d1a74d04c08471.bundle.js:348
    __webpack_require__ runtime.2846e04d228a8620.bundle.js:36
    __webpack_exec__ app.55d1a74d04c08471.bundle.js:464
    <anonymous> app.55d1a74d04c08471.bundle.js:465
    O runtime.2846e04d228a8620.bundle.js:83
    <anonymous> app.55d1a74d04c08471.bundle.js:466
    webpackJsonpCallback runtime.2846e04d228a8620.bundle.js:1228
    <anonymous> app.55d1a74d04c08471.bundle.js:10
```

Going into browser settings and clearing localstorage for `localhost` fixes it. It makes sense that the data fails to load in this case since the key it's stored under is different and the data may be laid out differently within the storage, but is there anything we can do to catch this and just clear or at least ignore the invalid state so the application doesn't end up broken?

andrewazores:
^ I don't think that behaviour is unique to this PR though, but it might be a good opportunity to address it as well.

tthvo:
Oh yes there is actually a feature in redux-persist that does migrating previous state. I will look into it and do a simpler version of it.

tthvo:
I am thinking of doing versioning:

- Add an optional _version field in the reducer state.
- Then, if there is a version mismatch, we can just discard the version in the localStorage.

What do you think?

andrewazores:
Sounds like a reasonable plan to me. Would the version be for the store as a whole or per slice? Losing AA card filter states on version updates isn't so bad, but losing dashboard layouts would be very unfortunate.

tthvo:
I think versioning per slice is more convenient as changes mostly just tough a single slice.

andrewazores:
Agreed, sounds good. I suppose there could actually be a top-level version too which follows the same logic. It would be unlikely to change, but if it did, we would be glad it's versioned.

tthvo:
>  I suppose there could actually be a top-level version too which follows the same logic. It would be unlikely to change, but if it did, we would be glad it's versioned.

Ahh got a little unfortunate here as store state is constructed from reducers. Adding a custom state on root state is not allowed... I guess we could just version slices then.

tthvo:
There is no clean up for localStorage if there is a version mismatch. I think we should let it be in case the user switches to an older version of web console that uses that version.

andrewazores:
It would probably only be developers who downgrade to an older version, not end users, so I wouldn't worry too much about trying to maintain storage compatibility in both directions.

andrewazores (approved):

Issue #757, opened 2022-12-13
[Bug] Notifications can overwhelm the screen and cover the notification drawer


andrewazores:
## Current Behavior:
When a lot of notifications are present, they tend to overwhelm the right side of the screen, even partially obscuring the notification icon.

## Expected Behavior:
A large stream of notifications should not block access to the notification drawer or the actions within. The "mark all read" action within the notification drawer becomes inaccessible, which would be a reasonable remedy for this large stream of notifications obscuring the screen. The user's best bet currently is probably to go to Settings and disable all notifications, or at least the categories that are frequently occurring.

Maybe the notification boxes can appear shifted slightly lower so they do not obscure the notification badge. And when the notification badge is clicked, the notification drawer slides out, which contains all the same information as the boxes - so perhaps the boxes can be hidden when the drawer is displayed.

## Screenshots
![image](https://user-images.githubusercontent.com/3787464/207450764-ea0af157-d0da-416d-9ff5-de089bc87224.png)

## Steps To Reproduce:
1. Create a rule that applies to several target applications and copies recordings to archives frequently.
2. Ensure all notifications are enabled

Issue #661, opened 2022-11-16
[Task] Revamp unit test framework to isolate snapshot testing


tthvo:
Related to #648 

It appears sometimes that including snapshot testing in the same test suite with other tests causes some irregularities in test setup (i.e. `setState on unmounted component`, or `TypeError: unable to read property from null (reading children`).

To reproduce a simple example:

1. Go to `src/test/Recordings/Recordings.test.tsx`.
2. Move the snapshot test up front as the first test.
3. Add `.mockReturnValueOnce(of(true))` as first call to `.spyOn(defaultServices.api, 'isArchiveEnabled')` for mocking this service call.
4. Run the test suite
    ```bash
    $ yarn test -t "<Recordings /> -u " // Need to update snapshot
    ```
5. See error:
```log
TypeError: Cannot read properties of null (reading 'children')

      at Tabs._this.countOverflowingElements (node_modules/@patternfly/react-core/dist/js/components/Tabs/Tabs.js:35:51)
      at Tabs.Object.<anonymous>.Tabs.componentDidUpdate (node_modules/@patternfly/react-core/dist/js/components/Tabs/Tabs.js:191:49)
      at commitLifeCycles (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:10449:22)
      at commitLayoutEffects (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:13295:7)
      at HTMLUnknownElement.callCallback (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:9972:14)
      at HTMLUnknownElement.callTheUserObjectsOperation (node_modules/jsdom/lib/jsdom/living/generated/EventListener.js:26:30)
      at innerInvokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:338:25)
      at invokeEventListeners (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:274:3)
      at HTMLUnknownElementImpl._dispatch (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:221:9)
      at HTMLUnknownElementImpl.dispatchEvent (node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:94:17)
      at HTMLUnknownElement.dispatchEvent (node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:231:34)
[...]

PR #761, opened 2022-12-14
fix(notifications): handle overflow

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #757

## Description of the change:
This change adds overflow handling to the notification alert group and notification drawer.

https://www.patternfly.org/v4/components/alert-group#toast-alert-group-with-overflow-capture

## Motivation for the change:
This reduces visual clutter on the screen and keeps the UI from feeling overfilled and chaotic when many actions are happening within a short timeframe. It is also easier for the user to dismiss all notifications when there are very many since the notification drawer can be summoned in a second way (clicking the overflow message), and summoning the drawer also hides the toast boxes.

## How to manually test:
There is a new temporary plus button on the top bar to help testing. This is to be removed before merge. Clicking it adds a new notification immediately.

Generate notifications with the temporary button or by any other means and observe the Overflow message that appears when there are too many notifications. Go to the Settings view and try changing the visible notifications count to various values and repeat.


src/app/Settings/NotificationControl.tsx
```diff
@@ -92,7 +115,26 @@ const Component = () => {
     <>
       <Stack hasGutter>
         <StackItem key="all-notifications">
-          <Switch id="all-notifications" label="All Notifications" isChecked={allChecked} onChange={handleCheckAll} />
+          <Level>
```
tthvo, line 118: Do you think we should have the number spinner right below the switch? Then, the switch and the spinner, each can have a label (similar to how we labeling formgroup).

Right, if viewing in a large screen, it might be hard recognize there is a spinner on the right side.

Screenshot from 2022-12-15 16-01-43

tthvo: Looking at this, the spinner should allow manual input?

src/app/AppLayout/AppLayout.tsx
```diff
@@ -153,10 +172,15 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({ children }) => {
...
 
   const handleMarkNotificationRead = React.useCallback(
-    (key) => notificationsContext.setRead(key, true),
+    (key) => () => notificationsContext.setRead(key, true),
     [notificationsContext.setRead]
   );
 
+  const handleTimeout = React.useCallback(
+    (key) => () => notificationsContext.setHidden(key),
+    [notificationsContext, notificationsContext.setHidden]
```
tthvo, line 181: I think we remove dep on `notificationsContext` here.

tthvo (commented):
Looks nicee! Just some comments :D

src/app/Shared/FeatureFlag/FeatureFlag.tsx
```diff
@@ -0,0 +1,60 @@
...
+  BETA = 1,
+  PRODUCTION = 2,
+}
+
+export interface FeatureFlagProps {
+  level: FeatureLevel;
+  children?: React.ReactNode | undefined;
+}
+
+export const FeatureFlag: React.FunctionComponent<FeatureFlagProps> = ({ level, children }) => {
```
tthvo, line 52: Is this setting coming into actual release/production?

This means the user can see some "hack" component for testing right? Or are we removing `DEVELOPMENT` on release?

tthvo: Also, changing the feature level does not seem to change the UI right away (i.e. prod -> dev). Reload or moving to a different route will show the changes.

andrewazores: Yea, I think this could be a useful feature to ship. We had basically this same thing but more extensive in a previous project I worked on. I'm thinking of maybe hiding the `DEVELOPMENT` option unless the `process.env.CRYOSTAT_AUTHORITY` is set, ie. it's being run with the Webpack dev server.

tthvo: Ahh okayyy its actually very nice to have this :D

src/app/Settings/FeatureLevels.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { FeatureLevel } from '@app/Shared/Services/Settings.service';
+import { Select, SelectOption } from '@patternfly/react-core';
+
+const Component = () => {
+  const context = React.useContext(ServiceContext);
+  const addSubscription = useSubscriptions();
+  const [state, setState] = React.useState(0);
+  const [open, setOpen] = React.useState(false);
+
+  React.useLayoutEffect(() => {
```
tthvo, line 52: Same here?

src/app/Settings/FeatureLevels.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { ServiceContext } from '@app/Shared/Services/Services';
+import { useSubscriptions } from '@app/utils/useSubscriptions';
+import { UserSetting } from './Settings';
+import { FeatureLevel } from '@app/Shared/Services/Settings.service';
+import { Select, SelectOption } from '@patternfly/react-core';
+
+const Component = () => {
+  const context = React.useContext(ServiceContext);
+  const addSubscription = useSubscriptions();
+  const [state, setState] = React.useState(0);
```
tthvo, line 49: Can we use the `FeatureLevel` here instead of `number`?

src/app/Settings/FeatureLevels.tsx
```diff
@@ -0,0 +1,95 @@
...
+      setState(v);
+      context.settings.setFeatureLevel(v);
+      setOpen(false);
+    },
+    [setState, setOpen, context.settings.setFeatureLevel]
+  );
+
+  return (
+    <>
+      <Select isOpen={open} onToggle={handleToggle} selections={FeatureLevel[state]} onSelect={handleSelect}>
```
tthvo, line 71: feture-level

Maybe we should use `<Select  isFlipEnabled menuAppendTo={'parent'} />` to enable flip here?

tthvo: flip-enabled

src/app/Shared/FeatureFlag/FeatureFlag.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const context = React.useContext(ServiceContext);
+  const addSubscription = useSubscriptions();
+  const [activeLevel, setActiveLevel] = React.useState(FeatureLevel.PRODUCTION);
+
+  React.useLayoutEffect(() => {
```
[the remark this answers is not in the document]
andrewazores: It's not visible right now since the only feature-flagged component is added to a spot in the app layout that doesn't cause any other components to have to re-render or re-layout when appears or disappears. If the flagged component was in between two others then since the flag wrapper defaults to `PRODUCTION` level and then loads the current level as an effect, if it's not a layout effect then all components would show and then briefly be removed again after.

I could also get around this by loading the proper level in the `useState()` hook up front rather than loading a default and then subscribing to the level observable later, but then this means the local storage key needs to be exposed and read directly in this component rather than being encapsulated by the settings service.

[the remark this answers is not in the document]
tthvo: Ahh right that makes sense! Seems like this one is a better option for our case.

tthvo:
Just a small detail:

If I have the drawer open, and notifications come in. I see them in the list and the top-right pop-ups are hidden.

Screenshot from 2022-12-16 14-20-56

If I close the notification drawer, the pop-ups appear:

Screenshot from 2022-12-16 14-21-02


Do you think if any un-read notifications that appear while the drawer is open should be hidden by timing-out early? Not sure if we can do anything about that.

andrewazores:
Hmm. Not sure it's worth bothering to handle. I guess it could be done by checking if the drawer is open when new notifications are added and setting the `hidden: true` property immediately if so, but the service that handles the notification stream doesn't currently know the state of the drawer. Maybe it should? Opening the drawer could auto-hide all of the notifications so they don't re-appear when the drawer is closed again, regardless of if the notifications were there before or came in while the drawer was open.

tthvo:
>  Maybe it should? Opening the drawer could auto-hide all of the notifications so they don't re-appear when the drawer is closed again, regardless of if the notifications were there before or came in while the drawer was open.

Yeh agreed. Do you think this PR can include these changes?

andrewazores:
Sure, I'll look at it now.

tthvo:
Looks nicee! There is some issue. To reproduce (using the dev button to add notifications):

**Issue 1**

1. With the notification drawer closed, send a new notification -> pop-up appears.
2. Open the drawer, pop-up disappears.
3. Send another notification -> Notice the original notification pop-up appears.

**Issue 2**

1. With the drawer closed, send a new notification -> pop-up appears.
2. Open the drawer, pop-up disappears.
3. Close the modal.
4. Send in another notification -> Notice the original pop-up also appears.

I think both likely originate from the same root issue....

andrewazores:
Looks like I tried to get too fancy with the observables.

src/app/Notifications/Notifications.tsx
```diff
@@ -60,17 +60,17 @@ export class Notifications {
   );
   private readonly _drawerState$: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
 
-  constructor() {
-    this._drawerState$
-      .pipe(filter((v) => v))
-      .subscribe(() => this._notifications$.next(this._notifications.map((n) => ({ ...n, hidden: true }))));
```
tthvo: > Looks like I tried to get too fancy with the observables.

XD I think it was working if we modify the notification in-place....Pretty odd.


```typescript
this._drawerState$
      .pipe(filter((v) => v))
      .subscribe(() => this._notifications$.next(this._notifications.map((n) => { 
        n.hidden = true;
        return n;
      })));
```

andrewazores: ... alright then. :-/

tthvo: Ahhh I think here we call `next` but that new array is not assigned to internal `this._notifications` array, so it was using the old one this while time.

src/app/Notifications/Notifications.tsx
```diff
@@ -57,12 +58,35 @@ export class Notifications {
...
+  setDrawerState(state: boolean): void {
+    this._drawerState$.next(state);
+  }
 
   notify(notification: Notification): void {
     if (!notification.key) {
       notification.key = nanoid();
     }
     notification.read = false;
+    if (notification.hidden === undefined) {
```
tthvo, line 87: Some reason behind this checking for `undefined`?

andrewazores: Not currently, but I thought it could be useful in the future. This would allow for a notification to be emitted with `hidden: true` explicitly set by the caller so that the notification doesn't generate a toast, even if the drawer is closed. This way the drawer state is only used as the default visibility setting if the caller hasn't specified one.

tthvo: Ahh right cool! Makes sense :D

andrewazores: I guess the websocket connection notification would be a good example of such a "quiet" notification though. May as well do it now.

src/app/Notifications/Notifications.tsx
```diff
@@ -93,8 +95,10 @@ export class Notifications {
     } else if (typeof notification.message !== 'string') {
       notification.message = JSON.stringify(notification.message);
     }
-    this._notifications.unshift(notification);
-    this._notifications$.next(this._notifications);
+    this._notifications$.pipe(first()).subscribe((prev) => {
+      prev.push(notification);
```
tthvo, line 99: We might need to push the new notifications in front?

src/app/Dashboard/Dashboard.tsx
```diff
@@ -99,89 +100,96 @@ const PlaceholderCard: React.FunctionComponent<
...
-        }),
-        defaultValue: '',
-        description: 'an async stream selection menu',
-        kind: 'select',
-      },
+export const getDashboardCards: (featureLevel?: FeatureLevel) => DashboardCardDescriptor[] = (
+  featureLevel?: FeatureLevel
+) => {
+  let cards = [AutomatedAnalysisCardDescriptor];
+  if (featureLevel === undefined || featureLevel === FeatureLevel.DEVELOPMENT) {
```
tthvo, line 107: Does `featureLevel === undefined` means defaulting to the lowest level (DEVELOPMENT)?

Then i guess we could clean up with default?

```typescript
export const getDashboardCards: (featureLevel?: FeatureLevel) => DashboardCardDescriptor[] = (
  featureLevel: FeatureLevel = FeatureLevel.DEVELOPMENT
) => {}
```

andrewazores: Ah yea, that's a good syntax trick to make it clearer to read.

It doesn't mean lowest level, it really does mean when callers don't pass an argument - like the other two `getConfigByX` functions in that file. In that case we just want the whole list of available configs since we're going to filter them out by a (hopefully) unique property.

tthvo: Ahh okayyy makes sense! I think we are good with either way since its just a syntax trick but I guess `featureLevel === undefined` will be clearer in this sense.

tthvo (approved):
Looks great to me :D

andrewazores:
Sorry, one last feature-flag related change.

image

image

image

src/app/AppLayout/AppLayout.tsx
```diff
@@ -271,7 +326,25 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({ children }) => {
   const Header = (
     <>
       <PageHeader
-        logo={<Brand alt="Cryostat" src={cryostatLogo} className="cryostat-logo" />}
+        logo={
+          <>
```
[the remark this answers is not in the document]
tthvo: Also, are there any color guide on the labels? Like Dev should be red and Beta should be green?

src/app/AppLayout/AppLayout.tsx
```diff
@@ -237,10 +280,20 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({ children }) => {
     </DropdownToggle>
   );
 
+  // TODO refactor to use Masthead, Toolbar components: https://www.patternfly.org/v4/components/page
   const HeaderTools = (
     <>
       <PageHeaderTools>
         <PageHeaderToolsGroup>
+          <FeatureFlag level={FeatureLevel.DEVELOPMENT}>
```
tthvo, line 288: Should this be using strict comparison too?

src/app/AppLayout/AppLayout.tsx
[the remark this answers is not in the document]
andrewazores: I tried the same `Level hasGutter` trick but the label no longer being aligned with the logo made it look worse, IMO.

I don't know that there is any colour guidance for this kind of thing, since feature flags aren't a PF component or concept.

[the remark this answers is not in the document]
tthvo: Ahh thats unfortunate. How about adding `&nbsp;&nbsp;`?

```diff
diff --git a/src/app/AppLayout/AppLayout.tsx b/src/app/AppLayout/AppLayout.tsx
index 8d6a635..800e5a9 100644
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -329,6 +329,7 @@ const AppLayout: React.FunctionComponent<IAppLayout> = ({ children }) => {
         logo={
           <>
             <Brand alt="Cryostat" src={cryostatLogo} className="cryostat-logo" />
+            &nbsp;&nbsp;
             <FeatureFlag strict level={FeatureLevel.DEVELOPMENT}>
               <PageHeaderToolsItem>
                 <Label isCompact color="red">
```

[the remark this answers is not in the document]
tthvo: Screenshot from 2022-12-19 13-45-01

src/app/AppLayout/AppLayout.tsx
andrewazores: It could, though it doesn't make a difference in this case.

src/app/AppLayout/AppLayout.tsx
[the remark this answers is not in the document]
andrewazores: I guess... or maybe I'll just add an inline style to the label to give it some extra margin or something.

src/app/AppLayout/AppLayout.tsx
tthvo: Ahh yeh I was thinking adding strict for clearer presentation :D

src/app/AppLayout/AppLayout.tsx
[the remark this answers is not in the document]
tthvo: > I don't know that there is any colour guidance for this kind of thing, since feature flags aren't a PF component or concept.

Oh okayy then I think the current colors are good already.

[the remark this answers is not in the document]
tthvo: > I guess... or maybe I'll just add an inline style to the label to give it some extra margin or something.

I think thats even better :D

tthvo:
Seems like ci failed the first time with the exact test failure: https://github.com/cryostatio/cryostat-web/actions/runs/3734294205/jobs/6336169378 as described in https://github.com/cryostatio/cryostat-web/issues/661#issuecomment-1349011976. Cant reproduce it very consistently tho, but seems to occur more frequently lately.

tthvo (approved):
Looks great to me!

Issue #766, opened 2022-12-15
[Bug] Automated Rule form name input does not allow pasting


tthvo:
## Current Behavior:

Pasting a text from clipboard causes the name input to invalidate.

## Expected Behavior:

Pasting from clipboard a valid name should be allowed.

## Screenshots

![broken-name-input](https://user-images.githubusercontent.com/68053619/207752463-9de49e69-0fef-4fd6-8751-ac87a34e9e67.png)

## Steps To Reproduce:

1. Run smoketest on latest main.
2. Go to Automated Rule. Click Create.
3. Copy a text somewhere (i.e `a_rule`).
4. Paste in the name input of the form.
5. Notice the form shows red as error.

tthvo:
Another odd thing when inputting the name:

1. Single trailing space is allowed.
2. Multiple trailing space is not allowed.
3. If there are multiple trailing space in name input, same as 2, it is not allowed. However, if you delete (e.g. Backspace) those spaces until there is only a single trailing space, it is now not allowed. This contradicts 1.

PR #767, opened 2022-12-15
fix(rules): fix rule form validation

tthvo:
# Welcome to Cryostat! 👋

Fixes: #766 

## Description of the change:

- [x] Fixed name input validation (i.e should check new input value instead of current one). This should allow pasting and correct validation.
- [x] Added an error helper text for matchExpression input.
- [x] Added a unit test to check if invalid helper texts are shown

## Motivation for the change:

Previously, name input validation was not checking new input name due to using incorrect variable name. This causes unusual behaviour as described in the issue.

## How to manually test:

1. Run smoketest
2. Go to `Automated Rule` tab. Click Create.
3. Enter invalid name => show error helper text.
4. Past a valid name => show successful indicator.
5. Same for matchExpression input.

### DIFF
--- a/src/app/Rules/CreateRule.tsx
+++ b/src/app/Rules/CreateRule.tsx
@@ -102,11 +102,11 @@ const Comp: React.FunctionComponent<{}> = () => {
   const [loading, setLoading] = React.useState(false);
 
   const handleNameChange = React.useCallback(
-    (evt) => {
+    (name) => {
       setNameValid(RuleNamePattern.test(name) ? ValidatedOptions.success : ValidatedOptions.error);
-      setName(evt);
+      setName(name);
     },
-    [setNameValid, name, setName]
+    [setNameValid, setName]
   );
 
   const eventSpecifierString = React.useMemo(() => {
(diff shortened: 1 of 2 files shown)


andrewazores:
While you're here, this is also a bug worth fixing:

image

Open web-client, go to Dashboard, select no target `Select target...`. Then go to Automated Rules > Create. This should just display a nice empty state instead of an error view, and the HTTP error toast notification shouldn't appear.

tthvo:
Hmm i cant reproduce it somehow. This is what I see. Though, it should not show the form anyways and I am unsure where those templates coming from. I will have a look.

Screenshot from 2022-12-15 14-57-52

andrewazores:
Oh, weird. I can't reproduce it now either. Maybe something that was cached in my browser and that bug is lurking in another PR, then.

andrewazores:
IIRC the templates being displayed there when you have no target selection are from Cryostat itself, with the `Cryostat` template filtered out.

andrewazores:
Actually now that I'm saying that, the correct fix for the bug (wherever it stemmed from) is to bring it back to this behaviour that we are seeing. The user should be able to define rules even if they have no current targets to select from.

andrewazores (approved):

Issue #615, opened 2022-11-08
[Task] Merge date and time inputs to reduce toolbar used space


tthvo:
An idea would be this example from Patternfly.

![Screenshot from 2022-11-08 15-07-31](https://user-images.githubusercontent.com/68053619/200664618-b2c129bd-d300-43c4-9549-c126971c6743.png)

Source: https://www.patternfly.org/v4/demos/date-and-time-picker/#demos

This way we can have more space for implementing time range selection or timezone options.

PR #773, opened 2022-12-17
feat(toolbar): revamp datetime picker and date display

### LINKED ISSUE #571: [Task] Datetime picker should allow local datetime search

tthvo:
# Welcome to Cryostat! 👋

Fixes: #615 
Fixes #571 

## Description of the change:

- [x] Reduce toolbar size but merge date input and time input
- [x] Add support for selecting timezone for date filter.
- [x] Format `startTime` display column to be clearer, readable and open for localization support.
- [x] Use dayJs (only 2KB) for date utilities.
- [x] Rename category displays instead of using Object keys as category names.
- [x] Add support for choosing date format (locale) and timezone in Settings.
- [x] Fixed unit tests for new changes.

## Motivation for the change:

Reduce toolbar space and add support  for date search with selected timezone.

## How to manually test:
1. Go to Recordings tab.
2. Select any of date-related filters.
3. See new changes.


tthvo:
**Normal state:**

datetime-revamped

**Time dropdown (default 24h but 12h is also supported):**

time-dropdown

**Calendar dropdown:**

calendar-dropdown

**Timezone dropdown (currently supported local timezone + UTC):**

timezone-dropdowm


**Category dropdown (with formatted names):**

category-dropdown


**Start time display in selected timezone (with tooltip to show UTC time):**

timestamp-tooltip

tthvo:
Ready for review...Ignoring tests.

**For datetime picker:**

After playing around with different designs, I decided to go with this tab version. 
- The design is closer to those of Patternfly component then modern scrolling (hard with mouse), split-panel (width control is pain) or a clock (stretch-goal?). 
- It also limits the horizontal space usage. Though, vertical spacing might be problem if browsers shrinks vertically.

Screenshot from 2023-01-19 04-37-50

Screenshot from 2023-01-19 04-37-57

<hr/>

**For datetime display**

I took this chance to implemental localization for datetime display. Fortunately, there is `day.js` that is a pretty useful and lightweight to get the job done. 
- For now, timezone and locale are configurable via setting page.
- Supported timezones are taken directly from native Intl API (day.js uses this for timezone functionality too).
- Supported locales are imported from day.js package.
- I manage to localize all places that we display a datetime (might be missing some others?): Active recording start time, datetime filter chip, and notification cards.

Screenshot from 2023-01-19 04-39-09

Screenshot from 2023-01-19 04-37-40

tthvo:
Though, there is one difficulty is that localized format might also includes words (like above with Korean). So, allowing users to input localized datetime might cause trouble validating or maybe even the users do not know they have to localize input. 

We could standardize that text-input with ISOString but we also want to display localized datetime in that text-input when selecting from the calendar modal (quite conflicting). So, for now, I just make it `readOnly`.

Any thoughts so far? I will start writing tests after reviews :D

andrewazores:
That's a tricky one re: the localized text input. I think leaving the field readonly and requiring graphical selection is okay, for now at least. Does `day.js` or some other library not provide a way to parse those localized forms into a standard format (ex. epoch seconds) which we could use for querying?

tthvo:
> Does day.js or some other library not provide a way to parse those localized forms into a standard format (ex. epoch seconds) which we could use for querying?


Hmmm I was actually trying to set locale and use localized keys as `L LTS`. But seems like day.js cannot parse, even with `CustomParseFormat` and `AdvancedFormat`. It seems not to recognize those letters: https://day.js.org/docs/en/display/format#localized-formats

I just came across this issue https://github.com/iamkun/dayjs/issues/694. The last comment seems to pull the converted format out. I could give it a try again...

tthvo:
Actually, that seems to solve the issue now. I think we can enable them again :D

Also in that note, I think we should add an example date input as placeholder instead of format keys (i.e. `YYYY/MM/DD hh:mm:ss`) because key value might be localized (for example, `A` in Korean can be 오전/오후 instead of AM/PM).

tthvo:
Okay...so with latest changes, the users can input localized datetime string. Placeholder is changed to show a localized example.

However, it comes with a specified requirement that timezone comes last, separated by a space. That timezone is in abbreviation form (i.e. EST, PST) and constrained by the browser support. Since not sure it does not look like we have a localized way for timezone, I follow Patternfly guidline here: https://www.patternfly.org/v4/components/timestamp/design-guidelines/#elements to put the timezone last.

src/app/Settings/DatetimeControl.tsx
```diff
@@ -0,0 +1,166 @@
...
+            onSelect={handleDateLocaleSelect}
+          >
+            {dateLocaleOptions}
+          </Select>
+        </FormGroup>
+      </StackItem>
+      <StackItem key={'timezone-select'}>
+        <FormGroup>
+          <HelperText>
+            <HelperTextItem>{'Select current timezone.'}</HelperTextItem>
```
andrewazores, line 147: Should these messages be localized?

src/app/Shared/Services/Settings.service.tsx
```diff
@@ -47,16 +52,36 @@ import {
 } from './Api.service';
 import { NotificationCategory } from './NotificationChannel.service';
 
+dayjs.extend(utc);
+dayjs.extend(timezone);
+dayjs.extend(advanced);
```
andrewazores, line 57: Do these `extend()` calls need to be sprinkled around so many different files? It would be nice for cleanliness/maintenance reasons if this could just be done somewhere once globally.

tthvo: Ahh dayjs is said to be immutable. I think I misunderstood what they meant. Probably, I will write a custom hook for it.

src/app/Settings/DatetimeControl.tsx
tthvo: ah yes I will start localization texts within the datetimepicker. But I will leave out any related to setting page tho (preparing another PR for that).

src/app/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,238 @@
...
+  React.useEffect(() => {
+    if (prefilledDate) {
+      setDatetime(prefilledDate);
+    }
+  }, [setDatetime, prefilledDate]);
+
+  return (
+    <Form>
+      <Tabs
+        aria-label="Select a date or time tab"
```
andrewazores, line 172: Should these also be localized?

tthvo:
Ready again :D Changes from last commit include:
- [x] Added unit tests for datetime-related components.
- [x] Localize strings in datetime-related components. 
- [x] Clean up custom hooks (i.e. fix updates-on-unmounted errors, remove unused contexts).  
- [x] Remove some unused css rules
- [x] Organize and apply eslint fixes

src/app/DateTimePicker/DateTimePicker.tsx
```diff
@@ -0,0 +1,237 @@
...
+    <Form>
+      <Tabs
+        aria-label={t('DateTimePicker.ARIA_LABELS.TABS') || ''}
+        onSelect={handleTabSelect}
+        activeKey={activeTab}
+        isFilled
+        role="region"
+      >
+        <Tab key={'date'} eventKey={'date'} title={<TabTitleText>{t('DATE', { ns: 'common' })}</TabTitleText>}>
+          <FormGroup key={'calendar'}>
```
andrewazores, line 178: Should the `'calendar'` here and the `'time'` below be extracted to constants, or a typedef for them declared?

andrewazores (commented):
LGTM, any more feedback @maxcao13 ?

maxcao13:
Is it possible to make this year selector a bit bigger? Years like 2003 are cut off while 2002 and others are fine. 
image

tthvo:
> Is it possible to make this year selector a bit bigger? Years like 2003 are cut off while 2002 and others are fine.

Yup! Just extend them a bit out by overriding pf css :D Also, localize the calendar display content too :D

maxcao13 (approved):
Looks great to me!

PR #778, opened 2022-12-19
feat(rules): allow cleanup resources when disabling automated rules

tthvo:
# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/cryostat/pull/1289

## Description of the change:

- [x] Added warning modal for users to decide whether to clean up resources created by the automated rule.
- [x] Extend delete warning modals to supporting disabling.
- [x] Fix side-effects on mockRule in tests.

## Motivation for the change:

See https://github.com/cryostatio/cryostat/issues/1283

## How to manually test:
1. Go to Automated Rule tab
2. Create or enable a existing rule.
3. Now toggle the rule to disable it. The modal should show up.

### DIFF
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -257,10 +257,10 @@ export class ApiService {
     );
   }
 
-  updateRule(rule: Rule): Observable<boolean> {
+  updateRule(rule: Rule, clean = true): Observable<boolean> {
     const headers = new Headers();
     headers.set('Content-Type', 'application/json');
-    return this.sendRequest('v2', `rules/${rule.name}`, {
+    return this.sendRequest('v2', `rules/${rule.name}?clean=${clean}`, {
       method: 'PATCH',
       body: JSON.stringify(rule),
       headers,
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -260,20 +260,30 @@ export class ApiService {
   updateRule(rule: Rule, clean = true): Observable<boolean> {
     const headers = new Headers();
     headers.set('Content-Type', 'application/json');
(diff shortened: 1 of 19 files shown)


src/app/Shared/Services/Api.service.tsx
```diff
@@ -257,10 +257,10 @@ export class ApiService {
     );
   }
 
-  updateRule(rule: Rule): Observable<boolean> {
+  updateRule(rule: Rule, clean = true): Observable<boolean> {
     const headers = new Headers();
     headers.set('Content-Type', 'application/json');
-    return this.sendRequest('v2', `rules/${rule.name}`, {
+    return this.sendRequest('v2', `rules/${rule.name}?clean=${clean}`, {
```
andrewazores, line 263: I was just looking at this actually. I think we'll be making much more use of query params in the future for properly doing searching/filtering, sorting, pagination, etc. on the backend, so may as well do a tiny bit of refactoring here to get ahead of it.

This could instead look like:

```typescript
return this.sendRequest('v2', `rules/${rule.name}?${new URLSearchParams({ clean: String(clean) /* needed because this requires values of string type */ })`, { ... });`

This way the query string isn't being manually created by hand and it's a little less error-prone. Maybe this should also be passed in to `sendRequest` as an optional argument instead of being part of the interpolated path string.
```

andrewazores:
Looks like that weird flakey unit test failed again.

src/app/Shared/Services/Api.service.tsx
tthvo: Right that would be much cleaner! Added that!

tthvo:
> Looks like that weird flakey unit test failed again.

Yeh....I am checking why but I will wait after Max finishes eslint fixes to see if that catches anything :D

andrewazores (approved):

Issue #740, opened 2022-12-09
[Task] Apply yarn format:apply to ./src/ 


maxcao13:
Dependent on https://github.com/cryostatio/cryostat-web/pull/753
Dependent on https://github.com/cryostatio/cryostat-web/pull/747

There are around 500 fixes that can be applied automatically with the new eslint plugin `import/order`. 
Unfortunately it seems `import/first` has to be disabled because of problems with mocking the Date system time, which needs to be mocked before imports in the `DateTimePicker.test.tsx`.

Then `yarn eslint:apply` can be safely applied to every PR before pushing. 

There is additional task of fixing ALL eslint warnings manually so that the CI can start to check for it. That might be another chore if it sounds like a good idea.

I will prefer to open a PR when there are not as many open right now so that rebases aren't painful.

- [x] https://github.com/cryostatio/cryostat-web/pull/753
- [x]  https://github.com/cryostatio/cryostat-web/pull/747

## Tasks

- [ ] Apply yarn format:apply to automatically fix fixable eslint issues
- [ ] Manually go through all eslint warnings to fix them
- [ ] Re-enable `yarn eslint` within `ci.yaml` as part of the CI automation process

andrewazores:
> Then yarn eslint:apply can be safely applied to every PR before pushing. There is additional task of fixing ALL eslint warnings manually so that the CI can start to check for it. That might be another chore if it sounds like a good idea.

https://github.com/cryostatio/cryostat-web/pull/208

I had the intent to do that quite a while ago, and obviously at that time I also just gave up on it. But, I think the codebase now is actually in overall better shape, and we have a lot more automated tooling that should help. So perhaps it's an easier task despite the fact that there is a lot more code now.

maxcao13:
Oh, I see. That's why the `run: yarn lint` in the ci.yaml has been disabled. 

I do think it is not too hard for our codebase to be cleaned up for the closed task. Okay, I will add an additional subtask to this task to manually fix the warnings. Shouldn't be too painful I think. I will just start after those PRs are merged to save extra work.

PR #780, opened 2022-12-20
chore(eslint): eslint apply and add eslint to the CI

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #740

## Description of the change:
* [x] Apply yarn format:apply to automatically fix fixable eslint issues
* [x] Manually go through all eslint warnings to fix them
* [x] Re-enable yarn eslint within ci.yaml as part of the CI automation process

Things to note:
1. Changed instances of React.FunctionComponent to use React.FC instead. Same code, just shorter for readability.
2. Eslint has been enabled in the CI and hence must pass in order for the build to pass CI. Rules include:
  -  import order enforced: imports should be in alphabetical order 
```
import { something... } from '@app/Events/...'
import { something } from '@patternfly/...'
import * from 'react'
import * from 'react-router-dom'
import {...} from './Dashboard/something...'
```
  - no explicit any: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-explicit-any.md
  - objects themselves shouldn't call object property methods (https://eslint.org/docs/latest/rules/no-prototype-builtins)
  - The empty interface rule has been disabled since some props interface types are empty. Empty props should be named with `_`. See this comment: https://github.com/cryostatio/cryostat-web/pull/780#discussion_r1055192891

>   I just notice that it seems when callback and memo needs function props as dependencies, eslint needs props itself to also be in the dependency list, and then having props in the dep list doesn't need the props.[functionProp] to be in the list either. However, non-function properties don't need this, and we directly list props.whateverFieldProp as a hook dependency. This can be done this way so that the props don't have to be destructured if the props itself isn't large (e.g. 10 props).

  - unused vars should be either `_` or prepended with the underscore e.g. `_myvariable` (this includes props -> empty prop objects should be named `_` until the FC prop interface contains any.
  - Others... see .eslintrc for list of enabled ad disabled rules

## Motivation for the change:
This change is helpful because the front-end should adhere to some useful eslint rules.

## How to manually test:
1. None

### DIFF
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -79,7 +79,16 @@ import {
   ToolbarGroup,
   ToolbarItem,
 } from '@patternfly/react-core';
-import { BarsIcon, BellIcon, CaretDownIcon, CogIcon, ExternalLinkAltIcon, PlusCircleIcon, QuestionCircleIcon, UserIcon } from '@patternfly/react-icons';
+import {
+  BarsIcon,
+  BellIcon,
+  CaretDownIcon,
+  CogIcon,
+  ExternalLinkAltIcon,
+  PlusCircleIcon,
+  QuestionCircleIcon,
+  UserIcon,
+} from '@patternfly/react-icons';
 import * as _ from 'lodash';
 import * as React from 'react';
 import { Link, matchPath, NavLink, useHistory, useLocation } from 'react-router-dom';
@@ -192,10 +201,7 @@ const AppLayout: React.FunctionComponent<AppLayoutProps> = ({ children }) => {
     [notificationsContext]
   );
 
-  const handleTimeout = React.useCallback(
-    (key) => () => notificationsContext.setHidden(key),
-    [notificationsContext]
-  );
+  const handleTimeout = React.useCallback((key) => () => notificationsContext.setHidden(key), [notificationsContext]);
 
   React.useEffect(() => {
     addSubscription(
@@ -401,7 +407,6 @@ const AppLayout: React.FunctionComponent<AppLayoutProps> = ({ children }) => {
     ),
     [
       notificationsContext,
-      isNotificationDrawerExpanded,
       unreadNotificationsCount,
       errorNotificationsCount,
       handleNotificationCenterToggle,
@@ -414,7 +419,7 @@ const AppLayout: React.FunctionComponent<AppLayoutProps> = ({ children }) => {
       UserInfoToggle,
       userInfoItems,
       HelpToggle,
-      helpItems
+      helpItems,
     ]
   );
 
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -79,19 +79,10 @@ import {
   ToolbarGroup,
   ToolbarItem,
 } from '@patternfly/react-core';
-import {
-  BarsIcon,
-  BellIcon,
-  CaretDownIcon,
-  CogIcon,
-  ExternalLinkAltIcon,
-  PlusCircleIcon,
-  QuestionCircleIcon,
-  UserIcon,
-} from '@patternfly/react-icons';
-import _ from 'lodash';
+import { BarsIcon, BellIcon, CaretDownIcon, CogIcon, ExternalLinkAltIcon, PlusCircleIcon, QuestionCircleIcon, UserIcon } from '@patternfly/react-icons';
+import * as _ from 'lodash';
 import * as React from 'react';
-import { matchPath, Link, NavLink, useHistory, useLocation } from 'react-router-dom';
+import { Link, matchPath, NavLink, useHistory, useLocation } from 'react-router-dom';
 import { map } from 'rxjs/operators';
 import { AuthModal } from './AuthModal';
 import { SslErrorModal } from './SslErrorModal';
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -100,7 +100,7 @@ interface AppLayoutProps {
   children: React.ReactNode;
 }
 
-const AppLayout: React.FunctionComponent<AppLayoutProps> = ({ children }) => {
+const AppLayout: React.FC<AppLayoutProps> = ({ children }) => {
   const serviceContext = React.useContext(ServiceContext);
   const notificationsContext = React.useContext(NotificationsContext);
   const addSubscription = useSubscriptions();
@@ -229,7 +229,7 @@ const AppLayout: React.FunctionComponent<AppLayoutProps> = ({ children }) => {
   );
 
   const mobileOnSelect = React.useCallback(
-    (_selected) => {
+    (_) => {
       if (isMobileView) {
         setIsNavOpenMobile(false);
       }
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -35,11 +35,19 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-import * as React from 'react';
-import * as _ from 'lodash';
-import { ServiceContext } from '@app/Shared/Services/Services';
+import { AboutCryostatModal } from '@app/About/AboutCryostatModal';
+import cryostatLogo from '@app/assets/cryostat_logo_hori_rgb_reverse.svg';
+import build from '@app/build.json';
 import { NotificationCenter } from '@app/Notifications/NotificationCenter';
+import { Notification, NotificationsContext } from '@app/Notifications/Notifications';
 import { IAppRoute, navGroups, routes } from '@app/routes';
+import { DynamicFeatureFlag, FeatureFlag } from '@app/Shared/FeatureFlag/FeatureFlag';
+import { SessionState } from '@app/Shared/Services/Login.service';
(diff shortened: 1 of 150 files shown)


tthvo:
One question: is there some option for `eslint` to be verbose and outputting which files it is checking simiar to prettier?

src/app/AppLayout/AppLayout.tsx
```diff
@@ -242,7 +229,7 @@ const AppLayout: React.FunctionComponent<AppLayoutProps> = ({ children }) => {
   );
 
   const mobileOnSelect = React.useCallback(
-    (selected) => {
+    (_selected) => {
```
tthvo, line 232: We can use `_` here?

src/app/Dashboard/AddCard.tsx
```diff
@@ -240,37 +242,37 @@ export const AddCard: React.FunctionComponent<AddCardProps> = (props: AddCardPro
...
   cardTitle: string;
   controls: PropControl[];
-  config: any;
-  onChange: ({}) => void;
+  config: object;
+  onChange: (config: object) => void;
 }
 
-const PropsConfigForm = (props: PropsConfigFormProps) => {
+const PropsConfigForm = ({ onChange, ...props }: PropsConfigFormProps) => {
```
tthvo, line 249: Can we destructure `config` from props here too?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisConfigForm.tsx
```diff
@@ -212,12 +216,12 @@ export const AutomatedAnalysisConfigForm: React.FunctionComponent<AutomatedAnaly
     addSubscription(
       context.api.createRecording(recordingAttributes).subscribe({
         next: (resp) => {
-          if (resp && resp.ok && props.onCreate) {
-            props.onCreate();
+          if (resp && resp.ok && onCreate) {
+            onCreate();
```
tthvo, line 220: Just for consistency:

```typescript
if (resp && resp.ok) {
  onCreate && onCreate();
}
```

src/app/Dashboard/AutomatedAnalysis/Filters/AutomatedAnalysisNameFilter.tsx
```diff
@@ -36,33 +36,36 @@
...
 export interface AutomatedAnalysisNameFilterProps {
   evaluations: CategorizedRuleEvaluations[];
   filteredNames: string[];
   onSubmit: (inputName: string) => void;
 }
 
-export const AutomatedAnalysisNameFilter: React.FunctionComponent<AutomatedAnalysisNameFilterProps> = (props) => {
+export const AutomatedAnalysisNameFilter: React.FunctionComponent<AutomatedAnalysisNameFilterProps> = ({
+  onSubmit,
+  ...props
```
tthvo, line 51: Actually, on this note, I notice we are destructing callbacks but leave others in props. Is this true and the recommended way?

src/app/Dashboard/AddCard.tsx
tthvo: See discussion: https://github.com/cryostatio/cryostat-web/pull/780#discussion_r1054640619

src/app/AppLayout/AppLayout.tsx
tthvo: Actually, any guides between using `_` and `_some_var`?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisFilters.tsx
```diff
@@ -192,7 +199,7 @@ export const filterAutomatedAnalysis = (
 
   let filtered = topicEvalTuple;
 
-  if (!!filters.Name.length) {
+  if (filters.Name.length) {
```
tthvo, line 202: Somewhere in this function `filterAutomatedAnalysis` is still using `!!` and `!= null` (should be checked against `undefined` or using short cut as here).

src/app/Events/EventTypes.tsx
```diff
@@ -203,7 +203,7 @@ export const EventTypes: React.FunctionComponent<EventTypesProps> = (props) => {
   );
 
   const onToggle = React.useCallback(
-    (t: EventType, index: number) => {
+    (t: EventType, _index: number) => {
```
tthvo, line 206: If `_index` is not use, we can just remove it i think. Might need to update where it is being called.

src/app/Login/BasicAuthForm.tsx
```diff
@@ -35,15 +35,15 @@
...
 import { ActionGroup, Button, Checkbox, Form, FormGroup, Text, TextInput, TextVariants } from '@patternfly/react-core';
+import { Base64 } from 'js-base64';
+import * as React from 'react';
 import { map } from 'rxjs/operators';
 import { FormProps } from './FormProps';
-import { Base64 } from 'js-base64';
-import { AuthMethod } from '@app/Shared/Services/Login.service';
 
-export const BasicAuthForm: React.FunctionComponent<FormProps> = (props) => {
+export const BasicAuthForm: React.FunctionComponent<FormProps> = ({ onSubmit }) => {
```
tthvo, line 46: Do we need `{, ...props}`? I notice we are using this notions in many places. Is this the recommended way?

src/app/Login/NoopAuthForm.tsx
```diff
@@ -39,14 +39,14 @@ import { AuthMethod } from '@app/Shared/Services/Login.service';
 import * as React from 'react';
 import { FormProps } from './FormProps';
 
-export const NoopAuthForm: React.FunctionComponent<FormProps> = (props) => {
+export const NoopAuthForm: React.FunctionComponent<FormProps> = ({ onSubmit }) => {
```
tthvo, line 42: Same as above comment? Need `{..., props}`?

src/app/Login/OpenShiftPlaceholderAuthForm.tsx
```diff
@@ -35,40 +35,39 @@
...
 
   React.useEffect(() => {
-    const sub = combineLatest([
-      serviceContext.login.getSessionState(),
-      notifications.problemsNotifications(),
-    ]).subscribe((parts) => {
-      const sessionState = parts[0];
-      const errors = parts[1];
-      const missingCryostatPermissions = errors.find((error) => error.title.includes('401')) !== undefined;
+    const sub = combineLatest([context.login.getSessionState(), notifications.problemsNotifications()]).subscribe(
```
tthvo, line 53: Should use `addSubscription` here?

src/app/Login/BasicAuthForm.tsx
maxcao13: I just notice that it seems when callback and memo needs function props as dependencies, eslint needs `props` itself to also be in the dependency list, and then having `props` in the dep list doesn't need the `props.[functionProp]` to be in the list either. However, non-function properties don't need this, and we directly list `props.whateverFieldProp` as a hook dependency. This can be done this way so that the props don't have to be destructured if the props itself isn't large (e.g. 10 props).  

For the future though, yes I will add `...props` in case for future addional `FormProps`, but I did it this way in this file only cause I knew the only prop present was the `onSubmit` function prop.

src/app/AppLayout/AppLayout.tsx
maxcao13: Well, just there is a eslint rule unused-imports/no-unused-vars which warns if there are any unused vars, but these can be bypassed if the variable matches the regex in the .eslintrc file. Here is the snippet.

```
		"unused-imports/no-unused-vars": [
			"warn",
			{ "vars": "all", "varsIgnorePattern": "^_", "args": "after-used", "argsIgnorePattern": "^_" }
    ]
```

maxcao13: I don't think I correctly chose the way I wanted to use `_selected` here instead of `_`. I think in the future `_somevar` should be used if the variable is a an argument of a function that comes from a function that we created in the cryostat-web, but if it comes as a prop (e.g. here from `mobileOnSelect` which is a function handler used as a callback for Patternfly's `Nav` component (**which we didn't create**)), then we should use `_`. I will change that here.

src/app/Dashboard/AutomatedAnalysis/Filters/AutomatedAnalysisNameFilter.tsx
maxcao13: Oh, yes. There is no mention of a recommended way other than to destructure ALL the props, but eslint doesn't complain this way. See the other comment for more information.

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisFilters.tsx
maxcao13: Whoops, I don't remember why I changed this...

maxcao13: Oh, it was automatically "fixed" :
```
error  Redundant double negation  no-extra-boolean-cast
```

src/app/Login/BasicAuthForm.tsx
maxcao13: Hmm... actually eslint will complain since `...props` won't be used. I will keep it as it is.

andrewazores:
I'll approve and merge this now. If @maxcao13 or @tthvo have any further fixes or requests then they can go in a separate, smaller and easier to review, PR.

andrewazores (approved):

Issue #625, opened 2022-11-09
[Task] Add support for UI localization


andrewazores:
https://www.npmjs.com/package/@shopify/react-i18n

This or another similar library should help. UI strings should be provided by loadable documents rather than hardcoded into components.

Error messages from the backend will still end up in English, but those should either be dealt with separately, or localized strings used instead by the frontend by looking up ex. specific HTTP status codes in particular contexts.

maxcao13:
Doing a bit of research, it and looking at different i18n frameworks for React from this list https://github.com/jpomykala/awesome-i18n, I think narrowing it down from [react-i18next](https://react.i18next.com/) and formatjs's [react-intl](https://formatjs.io/docs/react-intl/), I think I will try react-i18next because it has a smaller bundle size, but both packages seem to do the same thing, react-intl just seems to be a bit more popular on GitHub with more stars.

PR #789, opened 2023-01-04
feat(i18n): add text localization scaffolding with i18n-react

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #625 

## Description of the change:
This change adds structuring for the language localization in the future. 

There is a helper bash script called `i18n-build.bash` which, when you run it, updates the `locales` folder with a localization json file which are under each locale. The only i18n namespace for this project which is currently used is `public`, but adding new json files count as new namespaces. Since we are only productizing Cryostat, and not other plugins like OpenShift console does (which I took inspiration from when writing this PR), we probably only need 1 namespace. 

OpenShift console also uses some other helpful scripts and localization frameworks seen here https://github.com/openshift/console/tree/master/frontend/i18n-scripts that we may want to also use in the future.

Change the language by pressing the toolbar dropdown at the top of the app (only "supports" English and 中文).

This PR only adds translation localization support for now, and only has added it for the `AboutDescription` component. Adding it to the entire project may take some time...

To test, I add a i18nProvider as a wrapper and pass in the i18n instance which has been initialized in `i18n/config.ts`.

I have not added any date time localization or helper for pluralization yet, which can be done as well and should probably be done in a separate PR.

## Motivation for the change:
In the future, if we want Cryostat to be supported in other languages that OpenShift also supports, (chinese, japansese, etc.) then we should also have some sort of support too.

## How to manually test:
1. Run normally 
2. Go to the About page in Cryostat.
3. Change the language in the toolbar dropdown.


i18n-build.bash
```diff
@@ -0,0 +1,44 @@
...
+substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
```
andrewazores, line 38: This license header should probably use `# leading hash` to demarcate comments. Maybe this file needs to be added to the license check script's ignore list.

locales/en/public.json
```diff
@@ -0,0 +1,12 @@
...
+  "About": "About",
+  "Bugs": "Bugs",
+  "CRYOSTAT_TRADEMARK": "Copyright The Cryostat Authors, The Universal Permissive License (UPL), Version 1.0",
+  "File a Report": "File a Report",
+  "Homepage": "Homepage",
+  "Known Issues": "Known Issues",
+  "License": "License",
+  "Mailing List": "Mailing List",
+  "Open Source License": "Open Source License",
+  "version": "version"
```
andrewazores, line 11: I don't suppose this `i18n` library has support for splitting these keys up by component in some way?

```json
{
  "About": {
    "version": "version"
  },
  "Recordings": {
    "age": "age"
  }
}
```

or something like that would be nice. If not, even this could work:

```json
{
  "About.ABOUT": "About",
  "About.BUGS": "Bugs",
  "About.FILE_A_REPORT": "File a Report",
  ...
  "Recordings.AGE": "age"
}
```

I think keying the messages `LIKE_THIS` just makes it more clear to the reader (developer, reviewer) that what they're looking at is a localization key and not the actual message that will be rendered.

src/app/About/AboutCryostatModal.tsx
```diff
@@ -51,7 +53,7 @@ export const AboutCryostatModal = ({ isOpen, onClose }) => {
         brandImageAlt="Cryostat Logo"
         isOpen={isOpen}
         onClose={onClose}
-        trademark={CRYOSTAT_TRADEMARK}
+        trademark={t('CRYOSTAT_TRADEMARK') as string}
```
andrewazores, line 56: Is the `as string` really needed here? I'd hope that the type signature is `t(msg: string): string` :-/

src/app/AppLayout/AppLayout.tsx
```diff
@@ -351,6 +378,17 @@ const AppLayout: React.FC<AppLayoutProps> = ({ children }) => {
         <Toolbar isFullHeight isStatic>
           <ToolbarContent>
             <ToolbarGroup variant="icon-button-group" alignment={{ default: 'alignRight' }}>
+              <FeatureFlag strict level={FeatureLevel.DEVELOPMENT}>
```
andrewazores, line 381: I think this can be flagged to `BETA` (non-strict).

src/app/About/AboutCryostatModal.tsx
maxcao13: Seems like the problem is that the function potentially returns `null` which is not assignable to `string | undefined`.
Doing digging, I can solve it through this tutorial. https://www.i18next.com/overview/typescript#argument-of-type-defaulttfuncreturn-is-not-assignable-to-parameter-of-type-xyz

src/i18n/i18next.d.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import '18next';
```
tthvo, line 38: Should be `import 'i18next';`?

src/app/About/AboutCryostatModal.tsx
andrewazores: Ahh, that makes more sense.

andrewazores:
If you modify only one of the locale files and not the others by adding/removing keys or otherwise changing the structure, is there a fallback method to load the keys from a different locale? ie. if I add a new key to the `en/public.json` only and then load the application with a different language selected, what happens?

build-i18n.bash
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -exuo pipefail
+
+FILE_PATTERN="src/**/*.{js,jsx,ts,tsx,json}"
+
+i18next "${FILE_PATTERN}" [-oc] -c "./i18next-parser.config.js" -o "locales/\$LOCALE/\$NAMESPACE.json"
```
tthvo, line 7: Maybe we can move to use `npx`, which should come with `npm` to run this command without directly install `i18next`?

Something like: 

```bash
npx i18next "${FILE_PATTERN}" -c "./i18next-parser.config.js" -o "locales/\$LOCALE/\$NAMESPACE.json"
```

andrewazores: Or add it as a script within `package.json` perhaps

tthvo: yehh thats even better :D

tthvo (commented):
Looks good overall :D Just eyeing other comments for now.

maxcao13:
> If you modify only one of the locale files and not the others by adding/removing keys or otherwise changing the structure, is there a fallback method to load the keys from a different locale? ie. if I add a new key to the `en/public.json` only and then load the application with a different language selected, what happens?

If the key doesn't exist in the specified/default namespace (specified in `src/i18n/config.ts`), then i18next will fallback to look at other fallback namespaces (`fallbackNS`), then it will fallback to look at other fallback languages. https://www.i18next.com/principles/fallback . So in the example, if there is an English entry in the key if we view the page in Chinese, it will fall back to English. If there is not any entry at all in the json files, it will fallback to the key name (e.g. About.ABOUT). This functionality can probably be overriden if needed https://www.i18next.com/overview/configuration-options#missing-keys

andrewazores:
Code looks good.

Maybe for now remove the Chinese localization files if they're only going to contain the `temp` strings?

maxcao13:
> Code looks good.
> 
> Maybe for now remove the Chinese localization files if they're only going to contain the `temp` strings?

Sure, running `yarn localize` does regenerate them by the way. For now since the translations aren't a big part of the project, I think it doesn't need to be added to the CI yet.

i18next-parser.config.js
```diff
@@ -0,0 +1,76 @@
...
+      jsx: ['JsxLexer'],
+      tsx: ['JsxLexer'],
+  
+      default: ['JavascriptLexer'],
+    },
+  
+    lineEnding: 'auto',
+    // Control the line ending. See options at https://github.com/ryanve/eol
+  
+    locales: ['en', 'zh'],
```
tthvo, line 47: > Sure, running yarn localize does regenerate them by the way. For now since the translations aren't a big part of the project, I think it doesn't need to be added to the CI yet.

Maybe we could comment out `zh` here to that `yarn localize` won't add them in?

tthvo (approved):
Looks good to me :D

PR #796, opened 2023-01-07
feat(dashboard): allow user-defined dashboard card sizes

maxcao13:
Signed-off-by: Max Cao <macao@redhat.com>

# Welcome to Cryostat! 👋

Fixes: #729 

## Description of the change:
Allows cards to be resizable and for these configs to be persisted.

## Motivation for the change:
See https://github.com/cryostatio/cryostat-web/issues/729

## How to manually test:
1. Run normally and add various card configs.
2. Resize some cards by dragging the right side of them.
3. Notice the sizes are persisted when reloading or navigating elsewhere.


maxcao13:
Ah, I assume the placeholders won't be up to test with since the test image won't have `DEVELOPMENT` features enabled. I will set the Automated Analysis Card default span to 6 to show the functionality then.

andrewazores:
> Ah, I assume the placeholders won't be up to test with since the test image won't have DEVELOPMENT features enabled. I will set the Automated Analysis Card default span to 6 to show the functionality then.

You can use the PR CI test images and have development feature flags enabled:

```bash
$ CRYOSTAT_IMAGE=ghcr.io/cryostatio/cryostat-web:pr-796-f65ecf43dedc0c8c08844b8f358bb36e04832408  CRYOSTAT_CORS_ORIGIN=http://localhost:9000 CRYOSTAT_DISABLE_SSL=true sh smoketest.sh
$ #in new terminal:
$ cd cryostat-web
$ yarn start:dev
```

andrewazores:
> Wondering if this issue wanted a resizable card like a draggable resize bar (e.g. the DrawerContentBody ?). If so, it may be possible to implement this but will take some time as I would have to take a look a look at Patternfly's isResizable implementation.

Something like a draggable resize bar would be nice, or any other method that allows the user some fine-grained control. It could simply be a slider component that they can pull up somehow which allows value selection from 1-12 for the card span, for example. Though, I'm not sure if all of those sizes really make sense - is there a reason for the user to choose a 1 or 11 span, or 2 or 10? Maybe if they are using an ultrawide aspect monitor?

andrewazores:
```diff
diff --git a/src/app/Dashboard/Dashboard.tsx b/src/app/Dashboard/Dashboard.tsx
index 836929f..e84cc30 100644
--- a/src/app/Dashboard/Dashboard.tsx
+++ b/src/app/Dashboard/Dashboard.tsx
@@ -226,8 +226,8 @@ export const Dashboard: React.FC<DashboardProps> = (_) => {
   );
 
   const handleResize = React.useCallback(
-    (idx: number) => {
-      dispatch(dashboardCardConfigResizeCardIntent(idx, cardConfigs[idx].span === 12 ? 6 : 12));
+    (idx: number, span: gridSpans) => {
+      dispatch(dashboardCardConfigResizeCardIntent(idx, span));
     },
     [dispatch, cardConfigs]
   );
@@ -245,7 +245,7 @@ export const Dashboard: React.FC<DashboardProps> = (_) => {
                   idx={idx}
                   defaultSpan={getConfigByName(cfg.name).defaultSpan}
                   onRemove={() => handleRemove(idx)}
-                  onResize={() => handleResize(idx)}
+                  onResize={(span: number) => handleResize(idx, span)}
                 />,
               ],
             })}
diff --git a/src/app/Dashboard/DashboardCardActionMenu.tsx b/src/app/Dashboard/DashboardCardActionMenu.tsx
index 04cee70..febd0c2 100644
--- a/src/app/Dashboard/DashboardCardActionMenu.tsx
+++ b/src/app/Dashboard/DashboardCardActionMenu.tsx
@@ -37,7 +37,16 @@
  */
 import { CardConfig } from '@app/Shared/Redux/Configurations/DashboardConfigSlicer';
 import { RootState } from '@app/Shared/Redux/ReduxStore';
-import { Dropdown, DropdownItem, gridSpans, KebabToggle } from '@patternfly/react-core';
+import {
+  Dropdown,
+  DropdownItem,
+  gridSpans,
+  KebabToggle,
+  Popover,
+  Slider,
+  Text,
+  TextVariants,
+} from '@patternfly/react-core';
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
@@ -46,12 +55,13 @@ export interface DashboardCardActionProps {
   idx: number;
   defaultSpan: gridSpans;
   onRemove: () => void;
-  onResize: () => void;
+  onResize: (span: gridSpans) => void;
 }
 
 export const DashboardCardActionMenu: React.FunctionComponent<DashboardCardActionProps> = (props) => {
   const cardConfigs: CardConfig[] = useSelector((state: RootState) => state.dashboardConfigs.list);
   const [isOpen, setOpen] = React.useState(false);
+  const [span, setSpan] = React.useState(props.defaultSpan || 12);
 
   const [t] = useTranslation();
 
@@ -62,9 +72,23 @@ export const DashboardCardActionMenu: React.FunctionComponent<DashboardCardActio
     [setOpen]
   );
 
+  const handleSpanChange = React.useCallback(
+    (v) => {
+      if (v < 1) {
+        v = 1;
+      }
+      if (v > 12) {
+        v = 12;
+      }
+      setSpan(v);
+    },
+    [setSpan]
+  );
+
   return (
     <>
       <Dropdown
+        onBlur={() => props.onResize(span)}
         isPlain
         isFlipEnabled
         menuAppendTo={'parent'}
@@ -76,13 +100,11 @@ export const DashboardCardActionMenu: React.FunctionComponent<DashboardCardActio
           <DropdownItem key="Remove" onClick={props.onRemove}>
             {t('REMOVE', { ns: 'common' })}
           </DropdownItem>,
-          props.defaultSpan == 12 ? null : (
-            <DropdownItem key="Resize" onClick={props.onResize}>
-              {cardConfigs[props.idx].span === props.defaultSpan
-                ? t('EXPAND', { ns: 'common' })
-                : t('DashboardCardActionMenu.SHRINK', { val: props.defaultSpan })}
-            </DropdownItem>
-          ),
+          <DropdownItem key="Resize">
+            <Text component={TextVariants.p}>{t('EXPAND', { ns: 'common' })}</Text>
+            <Slider value={span} onChange={handleSpanChange} step={1} min={1} max={12} />
+            <Text component={TextVariants.p}>{span}</Text>
+          </DropdownItem>,
         ].filter((item) => item !== null)}
       />
     </>
```

Quick hack on top of your patch here to demonstrate. It's not a great UX, but it gives some easy flexibility to lay things out how the user might like it depending on the content they choose to place and what kind of display and display settings they use.

image

tthvo:
Just a though to extend above concept: maybe we could allow dragging at corners (or at least bottom-right) of any card to resize instead of slider. I guess we can make it "snap" to certain defined span if it reaches a certain size. Not sure how hard it is to implement it :D

andrewazores:
Yea I agree with that, I think that's similar to what we were mentioning above with the "draggable resize". I think in that case the card should always snap to the span width sizes as it is dragged. I also have no idea how hard this would be to implement though, not sure if there is already any support in Patternfly for something like this.

andrewazores:
Anyway, I don't think we need to go all the way to drag-to-resize card borders with this initial PR. Something better than my hack above hopefully, but at least something still that allows the user to pick a `gridSpans` (1-12) value for each card, just with nicer controls than what I did.

maxcao13:
Updated the resizing, thoughts?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+export const DraggableRef: React.FunctionComponent<DraggableRefProps> = (props) => {
+  const dispatch = useDispatch();
+
+  const cardRef = React.useContext(DashboardCardContext);
+  const draggableRef = React.useRef<HTMLDivElement>(null);
+  const isResizing = React.useRef<boolean>(false);
+  const setInitialVals = React.useRef<boolean>(true);
+
+  const SMALLEST_CARD_WIDTH = 126;
```
andrewazores, line 65: Where does this constant come from?

andrewazores (commented):
Nice, this looks pretty decent and works quite well.

src/app/Shared/Redux/Configurations/DashboardConfigSlicer.tsx
```diff
@@ -74,8 +82,19 @@ export const dashboardCardConfigDeleteCardIntent = createAction(DashboardConfigA
...
+    payload: {
+      idx,
+      span,
+    } as DashboardResizeConfigActionPayload,
+  })
+);
+
 export interface CardConfig {
   name: string;
+  span: gridSpans;
```
tthvo, line 97: Since the shape of this state part changes, we should increment `_version` for this slicer?

src/app/Dashboard/DashboardCardActionMenu.tsx
```diff
@@ -64,9 +67,9 @@ export const DashboardCardActionMenu: React.FunctionComponent<DashboardCardActio
         onSelect={onSelect}
         dropdownItems={[
           <DropdownItem key="Remove" onClick={props.onRemove}>
-            Remove
+            {t('REMOVE', { ns: 'common' })}
           </DropdownItem>,
-        ]}
+        ].filter((item) => item !== null)}
```
tthvo, line 72: I think React will ignore `null` for us already? Or is it for any other purposes?

https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

src/app/Dashboard/ResizableCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import * as React from 'react';
+
+export interface DashboardContextProps {}
```
tthvo, line 41: Is this still needed?

src/app/Dashboard/Dashboard.tsx
```diff
@@ -68,6 +73,7 @@ export interface PropControl {
 export interface DashboardProps {}
 
 export interface DashboardCardProps {
+  dashboardIdx: number;
```
tthvo, line 76: Maybe call it `dashboardId`? In the future, we might want to avoid using index as id (i.e. sorting maybe).

src/app/Dashboard/Dashboard.tsx
```diff
@@ -84,20 +90,23 @@ const PlaceholderCard: React.FunctionComponent<
...
+          <Text>title: {props.title}</Text>
+          <Text>message: {props.message}</Text>
+          <Text>count: {props.count}</Text>
+          <Text>toggle: {String(props.toggleswitch)}</Text>
+          <Text>menu: {props.menu}</Text>
+          <Text>asyncmenu: {props.asyncmenu}</Text>
+          <Text>asyncmenus: {props.asyncmenu2}</Text>
+        </CardBody>
+      </Card>
+      <DraggableRef dashboardIdx={props.dashboardIdx} minimumSpan={1} />
```
tthvo, line 108: `minimumSpan` should be taken from descriptor?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,139 @@
...
+    props.minimumSpan,
+    props.dashboardIdx,
+    dispatch,
+  ]);
+  const callbackMouseUp = React.useCallback(handleOnMouseUp, [handleOnMouseUp]);
+
+  return (
+    <div
+      ref={draggableRef}
+      className="card"
```
tthvo, line 129: We can remove this classname?

tthvo: Also, use shortcut `<div ...props />` since this div has no children.

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,139 @@
...
+  else if (val > max) val = max;
+  const ans = Math.round((b - a) * ((val - min) / (max - min)) + a);
+  return _.clamp(ans, a, b) as gridSpans;
+}
+
+export const DraggableRef: React.FunctionComponent<DraggableRefProps> = (props) => {
+  const dispatch = useDispatch();
+
+  const cardRef = React.useContext(DashboardCardContext);
+  const draggableRef = React.useRef<HTMLDivElement>(null);
```
tthvo, line 61: Seem like we are not utilizing this ref anywhere. Can we remove it?

src/app/Dashboard/DraggableRef.tsx
tthvo: Also, we should add some top-right-radius too?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,139 @@
...
+      return;
+    }
+    isResizing.current = false;
+    setInitialVals.current = true;
+    document.body.style.removeProperty('cursor');
+    document.removeEventListener('mousemove', callbackMouseMove);
+    document.removeEventListener('mouseup', callbackMouseUp);
+  };
+
+  const callbackMouseMove = React.useCallback(handleMouseMove, [
```
tthvo, line 118: Probably, bring the `handle*` functions right into their corresponding `callback*` function since now, these `handle*` will get generated/assigned in every re-render.

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,139 @@
...
+  };
+
+  const handleMouseMove = (e: MouseEvent) => {
+    const mousePos = e.clientX;
+    e.stopPropagation();
+    if (!isResizing.current) {
+      return;
+    }
+    if (cardRef.current && window.visualViewport) {
+      cardRef.current.scrollIntoView({ behavior: 'smooth', block: 'center', inline: 'end' });
```
tthvo, line 84: What is the use of scrollIntoView here?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,139 @@
...
+      const cardRect = cardRef.current.getBoundingClientRect();
+      const cardLeft = cardRect.left + SMALLEST_CARD_WIDTH * props.minimumSpan;
+
+      if (setInitialVals.current) {
```
tthvo, line 88: Any specific use of this ref anywhere? Seems to toggle once every resize but not checking anywhere?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,139 @@
...
+import React from 'react';
+import { useDispatch } from 'react-redux';
+import { DashboardCardContext } from './ResizableCard';
+
+export interface DraggableRefProps {
+  dashboardIdx: number;
+  minimumSpan: gridSpans;
+}
+
+function normalizeAsGridSpans(val: number, min: number, max: number, a: number, b: number): gridSpans {
```
tthvo, line 50: `a` and `b` should be typed `gridSpans`?

tthvo (commented):
Wow looks very niceee! Just some comments about src code :D

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const handleMouseMove = (e: MouseEvent) => {
+    const mousePos = e.clientX;
+    e.stopPropagation();
+    if (!isResizing.current) {
+      return;
+    }
+    if (cardRef.current && window.visualViewport) {
+      cardRef.current.scrollIntoView({ behavior: 'smooth', block: 'center', inline: 'end' });
+      const cardRect = cardRef.current.getBoundingClientRect();
+      const cardLeft = cardRect.left + SMALLEST_CARD_WIDTH * props.minimumSpan;
```
tthvo, line 86: Seems like we are looking at the right-side of the rectangle here? Should be renamed?

src/app/Dashboard/DraggableRef.tsx
maxcao13: Add 2 cards, and have both at half size. Then resize the right most card bigger. It's just to keep the screen focused on the card if it doesn't have enough room in a row, and it gets rearranged to the next row.

src/app/Dashboard/DraggableRef.tsx
maxcao13: I think it was used previously, but I forgot to remove it. Thanks.

src/app/Dashboard/DraggableRef.tsx
tthvo: Ahh I see it now! Very nicee. Didn't know it could help like that :D

src/app/Dashboard/DraggableRef.tsx
maxcao13: It's just sort of an eyeball magic number. I believe there is probably a better way to get the boundingClientRect.right of the minimum size of each Card (which is what I'm trying to compute in `minimumCardRight`), so likely this constant will get removed.

src/app/Dashboard/DashboardCardActionMenu.tsx
maxcao13: This was added before because I had a conditional in there that may have added a null component array item. That can't happen anymore so I'll remove it, thanks.

src/app/Dashboard/DraggableRef.tsx
andrewazores: Sounds good. Please just be careful to avoid hardcoding any assumptions that may have to do with screen dimensions like width in pixels. The assumption of 12 grid spans is fine since this is very well established in Bootstrap and there is the whole span-by-screensize css system for responsive layouts, but anything else should be generic or dynamic, or at the very least use relative css units.

src/app/Dashboard/Dashboard.tsx
```diff
@@ -84,7 +90,12 @@ const PlaceholderCard: React.FunctionComponent<
   } & DashboardCardProps
 > = (props) => {
   return (
-    <Card isRounded>
+    <ResizableCard
+      dashboardId={props.dashboardId}
+      minimumSpan={AllPlaceholderCardDescriptor.minimumSpan}
```
tthvo, line 95: Probably extract this minimumSpan into a const and use them for all placeholderCardDescriptor?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,125 @@
...
+    if (!isResizing.current) {
+      return;
+    }
+    isResizing.current = false;
+    document.body.style.removeProperty('cursor');
+    document.removeEventListener('mousemove', callbackMouseMove);
+    document.removeEventListener('mouseup', callbackMouseUp);
+  };
+
+  const callbackMouseMove = React.useCallback(handleMouseMove, []);
```
tthvo, line 111: Need hooks deps. Probably best to pass directly "handle*" implementation here. Also, need to wrap `handleOnMouseDown` in useCallback too.

maxcao13:
I think maybe I will hold off on tests like with the other beta-ish Dashboard features, as this may be changed in the future.

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -0,0 +1,145 @@
...
+      const mousePos = e.clientX;
+      e.stopPropagation();
+      if (!isResizing.current) {
+        return;
+      }
+      if (cardRef.current && window.visualViewport) {
+        cardRef.current.scrollIntoView({ behavior: 'smooth', block: 'center', inline: 'end' });
+        const cardRect = cardRef.current.getBoundingClientRect();
+
+        if (setInitialVals.current) {
```
tthvo, line 85: Do you think we could use only `minWidth` ref but with declaration as:

```ts
const minWidth = React.useRef<number | undefined>(0);
```

Then, we just need to check if this minWith is defined, set if not and reset when done?

locales/en/common.json
```diff
@@ -1,3 +1,5 @@
 {
-  "CRYOSTAT_TRADEMARK": "Copyright The Cryostat Authors, The Universal Permissive License (UPL), Version 1.0"
+  "CRYOSTAT_TRADEMARK": "Copyright The Cryostat Authors, The Universal Permissive License (UPL), Version 1.0",
+  "EXPAND": "Expand",
+  "REMOVE": "Remove"
```
andrewazores, line 4: Are these localizations still used now that there is simply a draggable edge and not a menu item?

src/app/Dashboard/Dashboard.tsx
```diff
@@ -36,19 +36,23 @@
...
 import { Observable, of } from 'rxjs';
 import { AddCard } from './AddCard';
 import { AutomatedAnalysisCardDescriptor } from './AutomatedAnalysis/AutomatedAnalysisCard';
 import { DashboardCardActionMenu } from './DashboardCardActionMenu';
+import { ResizableCard } from './ResizableCard';
 
 export interface DashboardCardDescriptor {
   title: string;
+  minimumSpan: gridSpans;
+  defaultSpan: gridSpans;
```
andrewazores, line 55: Just a thought:

```typescript
interface DashboardCardDescriptor {
  size: {
    span: Sized<gridSpans>
    height: Sized<number>
  }
}

interface Sized<T> {
  minimum: T;
  default: T;
  maximum: T;
}
```

I'm not sure the generic type constraint `T` will work exactly as written, but I think it's illustrative enough.

locales/en/common.json
maxcao13: Nope, I will remove them, thanks.

andrewazores (approved):
Looks great to me.

maxcao13:
Oh, there was actually still a bug with resizing the cards over the right side of the screen that I missed. I will file a quick PR.

Issue #805, opened 2023-01-10
[Bug] Target selection cache with disabled auto-refresh causes outdated jvmIds


tthvo:
## Current Behavior:

When targets' jvmId changes (e.g. via api queries when adding jmx credentials), the displayed target's jvmId should also be updated.

## Expected Behavior:

JvmId stays the same, even after route changes or reloads.

## Steps To Reproduce:

> I think the problem I see has something to do with the React states (or caching?). If I add credentials to a target, and then check the jvmId right after, I will see the old id from pressing the TargetSelect details dropdown . Even if I refresh the page, it will stay like that. I have to explicitly switch targets with the TargetSelect and then switch back, and it will update correctly. I've noticed this even happens across Cryostat builds where, old jvmIds will persist on my webpage from previous builds. This might be another caching problem even, I'm not sure.

_Originally posted by @maxcao13 in https://github.com/cryostatio/cryostat/issues/1325#issuecomment-1377886166_

## Something useful

> ahh it was actually a problem with the way we save selected target to local storage:

> 1. `selectTargetFromCache` is outdated (i.e. no awareness of jvmId). Cached target in local storage will be used instead of fetched one so displayed jvmId is outdated.

 >     https://github.com/cryostatio/cryostat-web/blob/3e2ff76e89cc9b519c2b795d0e45fa41a1eae07a/src/app/TargetSelect/TargetSelect.tsx#L122

> 2. `TargetSelect` has no means of fetching new target information unless auto-refresh is enabled. So, adding credentials to target (i.e. which computes jvmId in some steps of the current queried endpoint) won't update target info on front-end with new jvm id.

> I think a quick way is to save only serviceUrl instead of the whole target in local storage and trigger target info queries whenever `JmxAuthForm` is submitted?

_Originally posted by @tthvo in https://github.com/cryostatio/cryostat/issues/1325#issuecomment-1377996536_

PR #806, opened 2023-01-11
fix(target): update target selection cache

tthvo:
# Welcome to Cryostat! 👋

Partial fixes: #805 
Related to: https://github.com/cryostatio/cryostat/pull/1327 (WS notifications should be sent).

## Description of the change:

Update target cache to save only connectUrl and should update currently selected target with new details.

## Motivation for the change:

See #805

## How to manually test:
2.  Choose any target with credentials. Either enter credentials using Auth modal or add in Security Panel.
3. Expand the target details. See the jvmId being updated.

### DIFF
--- a/src/app/TargetSelect/TargetSelect.tsx
+++ b/src/app/TargetSelect/TargetSelect.tsx
@@ -83,13 +83,14 @@ export const TargetSelect: React.FunctionComponent<TargetSelectProps> = (props) 
   const [warningModalOpen, setWarningModalOpen] = React.useState(false);
 
   const setCachedTargetSelection = React.useCallback(
-    (target, errorCallback?) => saveToLocalStorage('TARGET', target, errorCallback),
+    (targetConnectUrl: string, errorCallback?: () => void) =>
+      saveToLocalStorage('TARGET', targetConnectUrl, errorCallback),
     []
   );
 
   const removeCachedTargetSelection = React.useCallback(() => removeFromLocalStorage('TARGET'), []);
 
-  const getCachedTargetSelection = React.useCallback(() => getFromLocalStorage('TARGET', NO_TARGET), []);
+  const getCachedTargetSelection = React.useCallback(() => getFromLocalStorage('TARGET', NO_TARGET.connectUrl), []);
 
   const resetTargetSelection = React.useCallback(() => {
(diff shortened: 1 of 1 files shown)


andrewazores:
I think it might be better to just ensure that the Targets model is fully synchronized with the backend by using our usual pattern of WS notifications.

On the backend, when new Credentials are added, the DiscoveryStorage observes this event and checks which targets it applies to. It maintains a collection of observed targets that it previously was unable to connect to, so it tries to connect to each of those again using the new credentials, and tries to populate the `jvmId` field.

*If that succeeds*, then a WebSocket notification should be emitted with the updated targets' state. The frontend can figure out how to merge that updated state into its current model, or it can simply drop its model and query to replace it entirely.

This way all of the logic to do with how things update (or are created and deleted), and the relationships and dependencies between things, is all implemented wholly on the backend. The frontend should stay as simple and presentational as possible.

There are a few reasons and benefits to doing things that way. First, it keeps things easier to maintain, since the logic is only located in one place instead of being implemented twice, where both implementations need to be kept in sync with each other. Second, it means that it's easier to reason about what the data model displayed by the frontend is since it is always just a prettified rendered view of what the API responses would return for an equivalent `curl` request.

On that second note, I think there is actually a nefarious bug in this PR's implementation. Since the target refresh query is triggered by the auth modal submission, the frontend instance's Target data model will only be refreshed and synchronized with the backend when *this* frontend instance is the one that added the new credentials. If there are multiple frontend instances, or if there is a `curl` API request etc. that adds the credentials, then the data model remains desynchronized until some other event occurs that triggers a query and full resync.

tthvo:
Ahh that would be better to send notification from backend then. I will remove the 'tap' to query targets but we might need to fix local storage cache. Will check backend to add fixes. Thanks for explaining :D

andrewazores:
Should the cached selection actually be the `jvmId` perhaps?

tthvo:
> Should the cached selection actually be the jvmId perhaps?

Was just concerned that some has null jvmId (like this scenario) so selecting any of them will take the first matched. So connectUrl might be a better option?

andrewazores:
Ah, that's true.

andrewazores (approved):

PR #829, opened 2023-01-20
feat(dashboard): allow drag-and-drop reordering of dashboard cards

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #730 

## Description of the change:
This change allows cards to be re-orderable with drag-and-drop swapping functionality. 

Again, no tests since beta dashboard functionality is likely to be changed in the future.

When create DashboardCards, you must pass in a `CardHeader` into the `cardHeader` prop.

## Motivation for the change:
See #730

## How to manually test:
1. Run normal smoketest and add some cards, and rearrange them by holding the top draggable bar and drag and dropping them around on top of each other to swap the cards places. You can also drag cards between other cards to rearrange them.

### DIFF
--- a/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
+++ b/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
@@ -750,24 +750,26 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
       id="automated-analysis-card"
       isCompact
       isExpanded={isCardExpanded}
+      cardHeader={
+        <CardHeader
+          onExpand={onCardExpand}
+          toggleButtonProps={{
+            id: 'automated-analysis-toggle-details',
+            'aria-label': 'Details',
+            'aria-labelledby': 'automated-analysis-card-title toggle-details',
+            'aria-expanded': isCardExpanded,
+          }}
+        >
+          <CardActions>{...props.actions || []}</CardActions>
+          <Level hasGutter>
+            <LevelItem>
+              <CardTitle component="h4">Automated Analysis</CardTitle>
+            </LevelItem>
+            <LevelItem>{reportSource}</LevelItem>
+          </Level>
+        </CardHeader>
+      }
     >
-      <CardHeader
-        onExpand={onCardExpand}
-        toggleButtonProps={{
-          id: 'automated-analysis-toggle-details',
-          'aria-label': 'Details',
-          'aria-labelledby': 'automated-analysis-card-title toggle-details',
-          'aria-expanded': isCardExpanded,
-        }}
-      >
-        <CardActions>{...props.actions || []}</CardActions>
-        <Level hasGutter>
-          <LevelItem>
-            <CardTitle component="h4">Automated Analysis</CardTitle>
-          </LevelItem>
-          <LevelItem>{reportSource}</LevelItem>
-        </Level>
-      </CardHeader>
       <CardExpandableContent>
         <Stack hasGutter>
           <StackItem>{errorMessage ? null : toolbar}</StackItem>
--- a/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
+++ b/src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
@@ -100,7 +100,7 @@ import * as React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { filter, first, map, tap } from 'rxjs';
 import { DashboardCardDescriptor, DashboardCardProps, DashboardCardSizes } from '../Dashboard';
-import { ResizableCard } from '../ResizableCard';
+import { DashboardCard } from '../DashboardCard';
 import { AutomatedAnalysisConfigDrawer } from './AutomatedAnalysisConfigDrawer';
 import { AutomatedAnalysisConfigForm } from './AutomatedAnalysisConfigForm';
 import {
@@ -744,12 +744,10 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
   }, [usingArchivedReport, usingCachedReport, report, isLoading, errorMessage]);
 
   return (
-    <ResizableCard
+    <DashboardCard
       dashboardId={props.dashboardId}
       cardSizes={AutomatedAnalysisCardSizes}
-      className="dashboard-card"
       id="automated-analysis-card"
-      isRounded
       isCompact
       isExpanded={isCardExpanded}
     >
@@ -784,7 +782,7 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
           </StackItem>
         </Stack>
       </CardExpandableContent>
-    </ResizableCard>
+    </DashboardCard>
   );
 };
 
--- a/src/app/Dashboard/Dashboard.tsx
+++ b/src/app/Dashboard/Dashboard.tsx
@@ -43,15 +43,7 @@ import {
 import { dashboardCardConfigDeleteCardIntent, RootState, StateDispatch } from '@app/Shared/Redux/ReduxStore';
 import { FeatureLevel } from '@app/Shared/Services/Settings.service';
 import { TargetView } from '@app/TargetView/TargetView';
-import {
-  CardActions,
-  CardBody,
-  CardHeader,
-  Grid,
-  GridItem,
-  gridSpans,
-  Text,
-} from '@patternfly/react-core';
+import { CardActions, CardBody, CardHeader, Grid, GridItem, gridSpans, Text } from '@patternfly/react-core';
 import * as React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { Observable, of } from 'rxjs';
@@ -273,7 +265,7 @@ export const Dashboard: React.FC<DashboardProps> = (_) => {
 
   return (
     <TargetView pageTitle="Dashboard" compactSelect={false}>
-      <Grid id={"dashboard-grid"} hasGutter>
+      <Grid id={'dashboard-grid'} hasGutter>
         {cardConfigs.map((cfg, idx) => (
           <GridItem span={cfg.span} key={idx} order={{ default: idx.toString() }}>
             {React.createElement(getConfigByName(cfg.name).component, {
@@ -290,7 +282,7 @@ export const Dashboard: React.FC<DashboardProps> = (_) => {
           </GridItem>
         ))}
         <GridItem key={cardConfigs.length} order={{ default: cardConfigs.length.toString() }}>
-          <AddCard/>
+          <AddCard />
         </GridItem>
       </Grid>
     </TargetView>
--- a/src/app/Dashboard/Dashboard.tsx
+++ b/src/app/Dashboard/Dashboard.tsx
@@ -272,7 +272,7 @@ export const Dashboard: React.FC<DashboardProps> = (_) => {
     <TargetView pageTitle="Dashboard" compactSelect={false}>
       <Grid id={'dashboard-grid'} hasGutter>
         {cardConfigs.map((cfg, idx) => (
-          <GridItem span={cfg.span} key={idx} order={{ default: idx.toString() }}>
+          <GridItem span={cfg.span} key={cfg.id} order={{ default: idx.toString() }}>
             {React.createElement(getConfigByName(cfg.name).component, {
               ...cfg.props,
               dashboardId: idx,
--- a/src/app/Dashboard/Dashboard.tsx
+++ b/src/app/Dashboard/Dashboard.tsx
@@ -38,7 +38,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import {
   CardConfig,
-  dashboardCardConfigReorderCardIntent,
   dashboardCardConfigResizeCardIntent,
 } from '@app/Shared/Redux/Configurations/DashboardConfigSlicer';
 import { dashboardCardConfigDeleteCardIntent, RootState, StateDispatch } from '@app/Shared/Redux/ReduxStore';
@@ -48,10 +47,6 @@ import {
   CardActions,
   CardBody,
   CardHeader,
-  DragDrop,
-  Draggable,
-  DraggableItemPosition,
-  Droppable,
   Grid,
   GridItem,
   gridSpans,
@@ -62,8 +57,8 @@ import { useDispatch, useSelector } from 'react-redux';
 import { Observable, of } from 'rxjs';
 import { AddCard } from './AddCard';
 import { AutomatedAnalysisCardDescriptor } from './AutomatedAnalysis/AutomatedAnalysisCard';
-import { DashboardCardActionMenu } from './DashboardCardActionMenu';
 import { DashboardCard } from './DashboardCard';
+import { DashboardCardActionMenu } from './DashboardCardActionMenu';
 
 export interface Sized<T> {
   minimum: T;
@@ -278,7 +273,7 @@ export const Dashboard: React.FC<DashboardProps> = (_) => {
 
   return (
     <TargetView pageTitle="Dashboard" compactSelect={false}>
-      <Grid hasGutter>
+      <Grid id={"dashboard-grid"} hasGutter>
         {cardConfigs.map((cfg, idx) => (
           <GridItem span={cfg.span} key={idx} order={{ default: idx.toString() }}>
             {React.createElement(getConfigByName(cfg.name).component, {
--- a/src/app/Dashboard/Dashboard.tsx
+++ b/src/app/Dashboard/Dashboard.tsx
@@ -38,19 +38,20 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import {
   CardConfig,
+  dashboardCardConfigReorderCardIntent,
   dashboardCardConfigResizeCardIntent,
 } from '@app/Shared/Redux/Configurations/DashboardConfigSlicer';
 import { dashboardCardConfigDeleteCardIntent, RootState, StateDispatch } from '@app/Shared/Redux/ReduxStore';
 import { FeatureLevel } from '@app/Shared/Services/Settings.service';
 import { TargetView } from '@app/TargetView/TargetView';
-import { CardActions, CardBody, CardHeader, Grid, GridItem, gridSpans, Text } from '@patternfly/react-core';
+import { CardActions, CardBody, CardHeader, DragDrop, Draggable, DraggableItemPosition, Droppable, Grid, GridItem, gridSpans, Text } from '@patternfly/react-core';
 import * as React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { Observable, of } from 'rxjs';
 import { AddCard } from './AddCard';
 import { AutomatedAnalysisCardDescriptor } from './AutomatedAnalysis/AutomatedAnalysisCard';
 import { DashboardCardActionMenu } from './DashboardCardActionMenu';
-import { ResizableCard } from './ResizableCard';
+import { DashboardCard } from './DashboardCard';
 
 export interface Sized<T> {
   minimum: T;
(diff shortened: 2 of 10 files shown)


andrewazores:
Very cool and seems to work nicely.

Dropping a card directly onto another to swap their positions is pretty good, but I think it would also be useful if a card can be dropped in between two others, or dropped at the beginning/end.

tthvo:
Looks also very nicee to me!

I was also thinking if we allow the drag on everywhere on the card instead of the "top extension"? Something like this:

https://material.angular.io/cdk/drag-drop/overview#cdk-drag-drop-overview

Styling can be similar and we can snap the card to its place.

andrewazores:
That might get a little bit "interesting" depending on the card's contents. Cards like the Automated Analysis have a bunch of whitespace that could be seen as part of the draggable card body in the middle of the card content, whereas other cards we've talked about like a Grafana card would have none of that. If there were a card that presents target details in some nested table or tree view for example, would the "empty" parts within the table allow dragging the card too?

This idea matches the https://www.patternfly.org/v4/components/card#selectable-cards , but those examples don't have interactable content within the cards.

https://www.patternfly.org/v4/components/drag-and-drop/ also looks something like that "drop between or at the ends" idea somewhat.

I think having a whole draggable header as a larger target (the region around the card title) might be nice at least, and maybe without the bright blue colouration.

maxcao13:
Sounds good, thanks for the feedback everyone!

andrewazores:
I think just draggable white spaces within the card headers is a good target. I'm not sold on the idea of the entire white body of the card being draggable.

maxcao13:
There's some colors that will be removed as well as the large red border around the colors, those are just for debugging and visualization for now. Now working to refactor so that only CardHeader is draggable.

maxcao13:
Should be ready now.

src/app/Shared/Redux/Configurations/DashboardConfigSlicer.tsx
```diff
@@ -65,6 +66,12 @@ export interface DashboardResizeConfigActionPayload {
   span: number;
 }
 
+export interface DashboardOrderConfigActionPayload {
+  idx: number;
+  order: number;
```
tthvo, line 71: Maybe name these 2 fields:  `previousOrder` and `nextOrder`? Also, need to update `_version` for this slicer.

andrewazores:
Very nice. How difficult would it be to make the drop zone targets a little wider? Dropping a card onto another to swap them has a nice large target since it seems like it happens if my cursor is anywhere over the "bottom" card, but dropping a card between two other means I have to get my cursor into the little gutter space. Likewise, dropping a card onto either end means I have to get my cursor out to the side of the cards, which leaves a fairly narrow margin between the card and either the navigation drawer or the edge of the screen. Maybe if my cursor is toward the outer edge of a bottom card, that can be considered the same as being beyond that edge? ex. to drop in between two cards, rather than my cursor needing to be in the grey gutter in between, it could be anywhere from the left card's right edge minus 20px to the right card's left edge plus 20px, or something like that.

andrewazores:
^ actually, being very careful about it, it seems like this might already be exactly how it is working, so maybe those buffer values on the drop zone target widths just need to be a little larger?

maxcao13:
> ^ actually, being very careful about it, it seems like this might already be exactly how it is working, so maybe those buffer values on the drop zone target widths just need to be a little larger?

That is indeed already how it works. Right now it was at 20px and it is configured by the `translateX` const in the file. We can try 30px or if that feels better?

andrewazores:
Cool, I'll try that out too, or maybe run a local dev build and suggest some other value if I think it feels better.

Haven't looked at the code yet but I think the functionality is super.

Can the dark grey borders on the edges and header be toned down a little? Maybe thinner and a lighter shade?

andrewazores:
I think even 60-80px feels better to me, it (60px here) seems to work even with small cards like this:

image

andrewazores:
https://user-images.githubusercontent.com/3787464/214719595-c331643f-cf1e-4aa1-94a7-332915bf3e0a.mp4

This is just a minor visual bug, but it seems like if I initiate a card drag and place it between two cards and hover there, the two other cards will move apart to open space and indicate that the card I'm holding will go in between. But if I don't drop the card and instead move it back to its original position, the two other cards remain spread apart. Maybe the animation should reverse and the two cards close the gap again.

maxcao13:
> I think even 60-80px feels better to me, it (60px here) seems to work even with small cards like this:
> 
With 60px I don't think I'm able to get the swap functionality when I try swapping on hovering on a small card. Is that okay? Unless it's just my problem or maybe a viewport issue.

andrewazores:
> > I think even 60-80px feels better to me, it (60px here) seems to work even with small cards like this:
> 
> With 60px I don't think I'm able to get the swap functionality when I try swapping on hovering on a small card. Is that okay? Unless it's just my problem or maybe a viewport issue.

Hmm. How small is a small card? 1 span?

Can the drop target width be calculated based on the adjacent cards' widths rather than being a hardcoded constant?

maxcao13:
> > > I think even 60-80px feels better to me, it (60px here) seems to work even with small cards like this:
> > 
> > 
> > With 60px I don't think I'm able to get the swap functionality when I try swapping on hovering on a small card. Is that okay? Unless it's just my problem or maybe a viewport issue.
> 
> Hmm. How small is a small card? 1 span?
> 
> Can the drop target width be calculated based on the adjacent cards' widths rather than being a hardcoded constant?

I've corrected the drop width to be calculated to 0.33 of each of the adjacent card widths dynamically. The scale can be corrected and tested with different values with the `offsetScale` variable. `translateX` now only works to adjust the transition distances when hovering between cards.

andrewazores:
Nice, I really like how this feels to use now.

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -744,32 +744,32 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
...
       id="automated-analysis-card"
-      isRounded
       isCompact
       isExpanded={isCardExpanded}
+      cardHeader={
+        <CardHeader
+          onExpand={onCardExpand}
+          toggleButtonProps={{
+            id: 'automated-analysis-toggle-details',
+            'aria-label': 'Details',
```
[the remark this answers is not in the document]
andrewazores: Or that can be handled separately, but since the foundation is laid and you're in here already, it could be rolled in. Up to you.

src/app/app.css
```diff
@@ -245,6 +245,42 @@ html, body, #root {
   overflow: auto;
 }
 
+.dashboard-card-resizable-wrapper { 
+  display: flex;
+  justify-content: space-between
```
tthvo, line 250: I think we can remove this. We just need the `display: flex` to make the resize bar expanded.

src/app/app.css
```diff
@@ -245,6 +245,42 @@ html, body, #root {
   overflow: auto;
 }
 
+.dashboard-card-resizable-wrapper { 
+  display: flex;
+  justify-content: space-between 
+}
+
+.pf-c-card.pf-m-rounded.dashboard-card {
```
tthvo, line 253: Probably remove this? Card has default border-radius 3px. This should already match with that we use for resize bar.

src/app/app.css
```diff
@@ -245,6 +245,42 @@ html, body, #root {
...
+
+.resizable-ref {
+  cursor: col-resize;
+  position: relative;
+  border-right: var(--pf-global--BorderWidth--lg) solid var(--pf-global--palette--black-500); /* 3px */
+  border-top-right-radius: var(--pf-global--BorderRadius--sm); /* 3px */
+  border-bottom-right-radius: var(--pf-global--BorderRadius--sm);
+}
+
+.draggable-ref__grip {
```
tthvo, line 266: I got an idea for this: Instead of adding a bottom line for header, we can make the card "float" similar so selectable cards when the header is hovered or dragged.

This can be achieved with a custom css class and set/reset this class on the DraggableRef.

```css
.draggable-ref__hover{
  box-shadow: 0 0.5rem 1rem 0 rgba(3, 3, 3, 0.16);
}
```

Notice AA card is having a shadow when hovering the header:

float-card

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -744,32 +744,32 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
   }, [usingArchivedReport, usingCachedReport, report, isLoading, errorMessage]);
 
   return (
-    <ResizableCard
+    <DashboardCard
```
tthvo, line 747: I think we should apply some truncating styles for label texts on header.
Screenshot from 2023-01-26 12-17-54

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -744,32 +744,32 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
...
+          onExpand={onCardExpand}
+          toggleButtonProps={{
+            id: 'automated-analysis-toggle-details',
+            'aria-label': 'Details',
+            'aria-labelledby': 'automated-analysis-card-title toggle-details',
+            'aria-expanded': isCardExpanded,
+          }}
+        >
+          <CardActions>{...props.actions || []}</CardActions>
+          <Level hasGutter>
```
tthvo, line 764: I think we should apply some truncating styles for label texts on header.

Screenshot from 2023-01-26 12-17-54

tthvo: On this note, seems like this happens because the card goes into `scroll mode` when in small spaces.

Probably because of this css class

```
.dashboard-card {
  width: 100%;
  overflow: auto;
}
```

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -35,125 +35,327 @@
...
+  isHovered: boolean;
+}
+
+function resetDroppableItem(droppableItem: DroppableItem) {
+  droppableItem.node.style.transition = transition;
+  droppableItem.node.style.transform = '';
+  droppableItem.isHovered = false;
+}
+
+function setDroppableItem(droppableItem: DroppableItem, transition: string, transform: string, isHovered?: boolean) {
```
tthvo, line 117: Try setting its `opacity: 0.5`. This will allow cards "under" to fade a bit on hovered.

tthvo: Screenshot from 2023-01-26 14-05-14

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -35,125 +35,327 @@
...
+  return [singleRowBetween, 'else'];
+}
+
+function getOrder(el: HTMLElement): number {
+  return parseInt(el.getAttribute(dashboardCardOrderAttribute) as string);
+}
+
+const initStyle = {};
+
+const transition = 'transform 0.5s cubic-bezier(0.2, 1, 0.1, 1) 0s';
```
tthvo, line 86: Maybe extra these (i.e. `transition` `delayedTransition`, `translateX` and `overlapTranslateY`)  to css rules. Then, we can add/remove from `Element.classList` as needed.

For  `translateX` and `overlapTranslateY`, we can specify a fix transform rule as we only have translation animation.

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -35,125 +35,327 @@
...
-        borderRight: '4px outset black',
-        borderRadius: '16px',
-      }}
-    />
+      ref={wrapperRef}
+      className={css(`${draggableRefKlazz}-wrapper`)}
+      onDragStart={onDragStart}
+      onTransitionEnd={onTransitionEnd}
+      style={{ ...refStyle }}
+      {...variableAttribute}
```
tthvo, line 356: Its best to use `data-*` attribute here https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

This way, we can pull from `Element.dataset` consistently?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -35,125 +35,327 @@
...
+  // different rows -> after: right end
+  const multRowAfter = rect1.top !== rect2.top && ev.clientX >= rect1.right && withinHeightRect1;
+  if (multRowAfter || singleRowAfter) {
+    return [true, 'right'];
+  }
+  // different rows -> between: no ends
+  return [singleRowBetween, 'else'];
+}
+
+function getOrder(el: HTMLElement): number {
```
tthvo, line 80: Any handling when the id is unknown?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -35,125 +35,327 @@
...
-import _ from 'lodash';
+import { dashboardCardConfigReorderCardIntent } from '@app/Shared/Redux/Configurations/DashboardConfigSlicer';
+import { css } from '@patternfly/react-styles';
 import React from 'react';
-import { useDispatch, useSelector } from 'react-redux';
-import { DashboardCardSizes } from './Dashboard';
-import { DashboardCardContext } from './ResizableCard';
+import { useDispatch } from 'react-redux';
+
+type itemPosition = 'left' | 'right' | 'else';
```
tthvo, line 43: Should capitalize this type?

tthvo: Maybe name `else` -> `inBetween`?

src/app/Dashboard/Dashboard.tsx
```diff
@@ -270,9 +270,9 @@ export const Dashboard: React.FC<DashboardProps> = (_) => {
 
   return (
     <TargetView pageTitle="Dashboard" compactSelect={false}>
-      <Grid hasGutter>
+      <Grid id={'dashboard-grid'} hasGutter>
         {cardConfigs.map((cfg, idx) => (
-          <GridItem span={cfg.span} key={idx}>
+          <GridItem span={cfg.span} key={idx} order={{ default: idx.toString() }}>
```
tthvo, line 275: I think we should start using unique keys now. Example issue: Try reordering an AA card (i.e. drag or replace it with other card), you will notice this AA card is being freshly built (i.e. one-time subscription to context.target.authRetry is sub again).

I think this is because React looks at `GridItem` keys and determine that they are the same instances (keys are indexs - reorder does not change). However, children under those GridItem are not different -> unmount entire children + freshly rebuild.

tthvo: We have a dependency on `nanoId` which can use in company with card type name to solve this (need to include in redux store state)

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -35,125 +35,327 @@
...
+//
+// |-----------------| mouse |-----------------|
+// |   rect1.right   ^       ^   rect2.left    |
+// |                 |       |                 |
+// returns [inBetweenTwoRectangles, {insertedOnLeft, insertedOnRight, else}]
+function inBetween(ev: MouseEvent, rect1: DOMRect, rect2: DOMRect): [boolean, itemPosition] {
```
tthvo, line 49: Use `const a = () => {}` to avoid hoisting? Not a problem but just easier to read/debug.

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -35,125 +35,327 @@
...
-      maxWidth.current = undefined;
+  const onMouseMoveWhileDragging = React.useCallback(
+    (ev: MouseEvent, droppableItems: DroppableItem[]) => {
+      hoveringDroppable.current = null;
+      hoveringIndex.current = null;
+      const currDragged = wrapperRef.current;
+      if (currDragged) {
+        isMouseDown.current = true;
+        const dragOrder = getOrder(currDragged);
+        droppableItems.forEach((di, idx) => {
```
tthvo, line 196: Why not use the index field of the DroppableItem?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -35,125 +35,327 @@
...
+              resetDroppableItem(_item);
+            });
+          } else {
+            // mouse is hovering between two cards
+            const nextItem = droppableItems[(idx + 1) % droppableItems.length];
+            const [betweenTwoRects, draggedPosition] = inBetween(ev, di.rect, nextItem.rect);
+            if (betweenTwoRects && droppableItems.length > 1) {
+              const hover = (idx + 1) % droppableItems.length;
+              // if the dragged card has a greater order than the hovered card
+              if (dragOrder > hover) {
```
tthvo, line 217: Probably revise `move` function to remove duplicate case. And remove nested loops

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
tthvo: Been playing around with this, there seems not to be a neat way to fix unless [Container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) is supported.

I think the best way now is to keep it scrollable :D Or maybe other ways^^

src/app/app.css
maxcao13: The card border radius is only acting on the card part but the resizable bar is another div that extends beside it meaning that it isn't part of the card. See here if I turn it off: 
image

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
maxcao13: I'm not sure about other ways for now, scrollable seems to be the best fit for now as a user would probably always have the choice to reduce their viewport tiny, but I'm not sure why they would... 

As for the automated analysis card itself, maybe the best thing to do is to have the card itself to handle its own overflow style, not too sure.

src/app/Dashboard/DraggableRef.tsx
maxcao13: Is toggling classes with specific css rules more performant than modifying the styles directly to the react ref? I can't find much about it.

andrewazores: I would think the main performance cost of the animation is for the browser to actually do the animation, ie repaint frames to execute it. Which way the animation style is applied sounds like it shouldn't make much difference, so long as neither method is causing some additional re-renders or anything like that.

Putting the transition(s) into css and adding/removing from the class list might just be a little harder to muck up with typos, and it sounds like it's more suited to being easily extended with more potential animations in the future. To me that's the potential value I'm seeing from refactoring in this way.

andrewazores:
https://user-images.githubusercontent.com/3787464/215214981-11a1c769-a499-47fc-b6da-03287ff01d91.mp4

Something not resetting fully again with the new greyed/faded animation?

andrewazores:
Anything else @tthvo ?

src/app/Dashboard/DraggableRef.tsx
tthvo: Agreed^^ also on this note, I found React docs say this:

https://reactjs.org/docs/faq-styling.html#are-inline-styles-bad

src/app/app.css
```diff
@@ -240,11 +240,72 @@ html, body, #root {
...
+  cursor: move;
+  cursor: grab;
+  cursor: -moz-grab;
+  cursor: -webkit-grab;
+  border-bottom: var(--pf-global--BorderWidth--sm) solid var(--pf-global--palette--black-500); /* 1px */
```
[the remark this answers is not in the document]
tthvo: I would probably increase the opacity to maybe 0.3-0.5?

src/app/Dashboard/DraggableRef.tsx
```diff
@@ -35,125 +35,365 @@
...
+  const multRowAfter = rect1.top !== rect2.top && ev.clientX >= rect1.right && withinHeightRect1;
+  if (multRowAfter || singleRowAfter) {
+    return [true, 'right'];
+  }
+  // different rows -> between: no ends
+  return [singleRowBetween, 'inBetween'];
+};
+
+const initStyle = {};
+const offsetScale = 0.33;
```
tthvo, line 121: Probably, uppercase this constant. Same for initStyle too.

tthvo (approved):
Looks great to me! Awesomeee work!

andrewazores:
Wonderful. Great job.

PR #832, opened 2023-01-23
chore(settings): localize setting page

tthvo:
# Welcome to Cryostat! 👋

Related to #793 
Related to #773 

## Description of the change:
- [x] Localize strings in setting page.
- [x] Safeguard for unsupported language from language detection.
- [x] Restyle helper texts
- [x] Add more test case for Language Setting. 

Also, a follow-up PR for #773:
- [x] Allow only ISO8601 input
- [x] AfterDate check should take into account same date. 
- [x] Update unit tests 

## Motivation for the change:
- Localization support for setting page.
- Strict parsing for datetime filter input.


src/app/Recordings/Filters/DatetimeFilter.tsx
```diff
@@ -152,17 +152,17 @@ export const DateTimeFilter: React.FunctionComponent<DateTimeFilterProps> = ({ o
...
+            timezone: undefined,
             validation: ValidatedOptions.error,
           };
         }
 
         const extractDatetime = parts.slice(0, parts.length - 1).join(' ');
         const d = dayjs(
           extractDatetime,
-          `${dayjs.Ls[dayjs.locale()].formats.L} ${dayjs.Ls[dayjs.locale()].formats.LTS}`,
+          `${dayjs.localeData().longDateFormat('L')} ${dayjs.localeData().longDateFormat('LTS')}`,
```
tthvo, line 165: @andrewazores Back on this topic again, I ran across this and it seems `dayjs` cannot reliably parse localized meridiem. For example, it could not parse `PM` in Korean (throwing invalid) but `AM` is recognized. And formatting/displaying is working perfectly. This is quite common to other locales too (e.g. Bangladesh, ...)

Sample fail :https://runkit.com/63cf2e7d1343950008a5db24/63cf2e7ea6283e00086eeee6

I am think we should restrict to support only English variants since that what we have for Languages as now? Or maybe a more standard input format (i.e. ISO8601)?

tthvo: Looks there are a bunch of reports about this: Looks like https://github.com/iamkun/dayjs/issues?q=is%3Aissue+is%3Aopen+meridiem

andrewazores: I think using ISO8601 for the input format is probably the best bet.

tthvo:
Ready for reviews :D This PR mostly localize the setting page and fixing unit tests. 

I ran into a bit difficulty with snapshot tests and i18next. In tests, we are using real implementation of `i18next` with language-detector plugin, while snapshots now are in `en`. I am thinking either setting language to `en` by default in `Common.tsx` (must be imported in test file) or probably just mock out the `i18next` module?

tthvo:
Seems like option 2 is working nicely so I will put it here unless we prefer stubs or other suggestions :D

src/app/Settings/AutoRefresh.tsx
```diff
@@ -104,9 +106,13 @@ const Component = () => {
...
+  descConstruct: 'SETTINGS.AUTO_REFRESH.DESCRIPTION',
   content: Component,
-  category: 'Connectivity',
+  category: 'SETTINGS.CATEGORIES.CONNECTIVITY',
 };
+
+/**
+ * t('SETTINGS.AUTO_REFRESH.TITLE')
+ * t('SETTINGS.AUTO_REFRESH.DESCRIPTION')
+ */
```
andrewazores, line 118: Remove comment?

andrewazores (commented):
Looks fine to me, just wondering about those block comments in a few files.

src/app/Settings/AutoRefresh.tsx
tthvo: Ahhh that was some limitation of the i18next parser that we are using. It just could not parse dynamic keys (must be explicit). If we remove these, it will remove them from the json file next time we run `yarn localize`.

https://github.com/i18next/i18next-parser#caveats

So, I have been including them...Not very nice tho.

tthvo: Ohh yes good idea! Maybe we can just use `i18n.ts` for all files under the same top-level directory under `/src`?

andrewazores: So in this case a `src/app/Settings/i18n.ts`? That would work too, though I think it might be more easily overlooked when changes are made to specific component files if it's done that way.

tthvo: Yes!! I am just a bit concerned that if we have new separate files for all component files that has a couple comment lines, for example under `/Settings`..., it might be quite clustered?

andrewazores (approved):
Looks good.

maxcao13 (approved):
Everything looks good, I don't see anything missing or wrong, and code looks good.

Thanks for doing this!

Issue #828, opened 2023-01-19
[Task] Collapse toolbar buttons


tthvo:
## Describe the feature

In small screen width, the buttons need to be collapsed so that it does not take it spaces. Another reason is to avoid overflowing. 

Currently, toolbar items are overflowing hidden, which is not very ideal. We can either make it scrollable or use flex to break lines.

## Screenshots

![Screenshot from 2023-01-19 14-21-36](https://user-images.githubusercontent.com/68053619/213539979-36d0b677-fd6d-4b9d-a68b-942afbac6f98.png)

![Screenshot from 2023-01-19 14-20-17](https://user-images.githubusercontent.com/68053619/213539981-b18c760c-75de-4e65-8c03-2ed690556f5d.png)

andrewazores:
Related on the topic of rendering on narrow viewports, the Automated Rules table also doesn't do so well.

tthvo:
Looks like even in regular or large viewports, overflow on Rule table still occurs if there is not enough space for entire row.

Issue #665, opened 2022-11-17
[Task] Need for better unit test coverage


tthvo:
Unit tests are still missing for come components and most utilities (i.e. shared api modules). We might not be able to cover 100% but as much as possible. Also need to figure out what to test for utilities.

Related to #661

PR #846, opened 2023-02-02
fix(layout): fix content overflow

tthvo:
# Welcome to Cryostat! 👋

Fixes: #828 

## Description of the change:

- [x] Enable horizontal scroll for rule table.
- [x] Collapse toolbar action buttons
  - [x] Active Recording Table
  - [x] Archive Recording Table. Note that PF advises against using overflow menu with only 2 buttons. So I add the breakpoint to lowest so that unless spaces are very limited, buttons are still shown.
  - [x]  Update tests. Got much complications with screen size testing.

## Motivation for the change:

The issue is quite severe for the Rule table as regular viewport also make it overflow. This PR addresses that and original issue with overflowing toolbar.

## How to manually test:
2. Go to Automated Rule tab and create a rule if not any. See the table is not scrollable.
3. Go to Recording Tab. Reduce the screen viewport. See the buttons collapsing into a dropdowm

## Screenshots

**Collapsed toolbar:**

Screenshot from 2023-02-02 16-02-45

**Collapsed Toolbar action buttons:**

Screenshot from 2023-02-02 16-02-28

**Scrollable Automated table:**

Screenshot from 2023-02-02 16-02-09

### DIFF
--- a/src/app/Recordings/Recordings.tsx
+++ b/src/app/Recordings/Recordings.tsx
@@ -61,7 +61,7 @@ export const Recordings: React.FunctionComponent<RecordingsProps> = (_) => {
 
   const cardBody = React.useMemo(() => {
     return archiveEnabled ? (
-      <Tabs id="recordings" activeKey={activeTab} onSelect={onTabSelect}>
+      <Tabs id="recordings" activeKey={activeTab} onSelect={onTabSelect} unmountOnExit>
         <Tab id="active-recordings" eventKey={0} title={<TabTitleText>Active Recordings</TabTitleText>}>
           <ActiveRecordingsTable archiveEnabled={true} />
         </Tab>
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -738,7 +738,7 @@ const ActiveRecordingsToolbar: React.FunctionComponent<ActiveRecordingsToolbarPr
         <ToolbarGroup style={{ alignSelf: 'start' }} variant="button-group">
           <ToolbarItem variant="overflow-menu">
             <OverflowMenu
-              breakpoint="xl"
+              breakpoint="lg"
               breakpointReference={
                 props.toolbarBreakReference ||
                 (() => document.getElementById('active-recordings-toolbar') || document.body)
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -63,6 +63,14 @@ import {
   Drawer,
   DrawerContent,
   DrawerContentBody,
+  Dropdown,
+  KebabToggle,
+  OverflowMenu,
+  OverflowMenuContent,
+  OverflowMenuControl,
+  OverflowMenuDropdownItem,
+  OverflowMenuGroup,
+  OverflowMenuItem,
   Text,
   Timestamp,
   TimestampTooltipVariant,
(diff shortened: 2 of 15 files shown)


tthvo:
Tests with recording tables are now only testing large viewports. We really should test the case where the toolbar is collapsed in #665.

maxcao13:
image

I would think a viewport size like this, the toolbar wouldn't shrink but it does here.

maxcao13:
Also when the view port is shrinked even more, there is a weird shadowy line below the filters toolbar vs. when its not shrinked. Is this intended?

image

tthvo:
> Also when the view port is shrinked even more, there is a weird shadowy line below the filters toolbar vs. when its not shrinked. Is this intended?
> 
> image

I think this is Patternfly table behaviour. I think there is a design decision behind (probably to distinguish from toolbar). Try their example:

Screenshot from 2023-02-02 16-09-23

tthvo:
> I would think a viewport size like this, the toolbar wouldn't shrink but it does here.

This is controlled by the `breakReference`, which I set to the toolbar and the `breakpoint`, which is now `xl`. If I set them lower, there is a weird edge case...

Screenshot from 2023-02-02 16-14-38

andrewazores:
I think that looks alright. If the viewport is even smaller then it still collapses it all down into the three-dots menu button and the filter, uh, martini glass button?

tthvo:
> If the viewport is even smaller then it still collapses it all down into the three-dots menu button and the filter, uh, martini glass button?

Yes, it should collapse them all into only 2 icons. Then, we would prefer to set the breakpoint to lower then `xl`, maybe `lg` but tolerate that edge case? 

Screenshot from 2023-02-02 16-41-30

src/app/Recordings/Recordings.tsx
```diff
@@ -61,7 +61,7 @@ export const Recordings: React.FunctionComponent<RecordingsProps> = (_) => {
 
   const cardBody = React.useMemo(() => {
     return archiveEnabled ? (
-      <Tabs id="recordings" activeKey={activeTab} onSelect={onTabSelect}>
+      <Tabs id="recordings" activeKey={activeTab} onSelect={onTabSelect} unmountOnExit>
```
maxcao13, line 64: What does this `unmountOnExit` do?

tthvo: This means we only build the active tab and remove any inactive out of DOM. This helps in avoiding "jump" while rendering when depending on the toolbar size instead of the entire viewport.

Without this option, you will originally see the Kebab first then the buttons on large viewport.

tthvo: Kinda unfortunate but I think it has an advantage that first render should be faster as it only needs to build one table.

andrewazores: https://www.patternfly.org/v4/components/tabs#tabs

> Unmounts tab children (removes them from the DOM) when they are no longer visible

Seems to make sense.

andrewazores:
Looking good. Do you want to also tackle this issue in this PR?

image

tthvo:
> Looking good. Do you want to also tackle this issue in this PR?

Sure! I will likely adjust the span of the target select though. 

I am thinking of removing those action buttons and let those actions handled in Topology view. At that point, this `TargetSelect` could be kept to be used a local selections.

andrewazores:
That makes sense to me. I imagine there could be some `+`-node when looking at Topology and selecting the Custom Targets realm. And within the Custom Targets topology "tree" (one realm node with a list of children), each child could have some action on it somewhere to delete it, perhaps with something like a little `x` button in the corner.

tthvo:
Exactly^^ Working on that now :D Quite difficulty with limited docs but lets see.

andrewazores (approved):

andrewazores:
Any other planned changes?

tthvo:
Ohh I think thats all changes I have :D

Issue #847, opened 2023-02-02
[Task] Grafana Cards


andrewazores:
Some good groundwork has been done:

- [x] https://github.com/cryostatio/jfr-datasource/issues/152
- [x] https://github.com/cryostatio/cryostat-grafana-dashboard/issues/11
- [x] https://github.com/cryostatio/cryostat/pull/1354

The net change to the datasource and backend means that the web-client can select any recording file to be displayed in the dashboard and not need to worry about cleaning up those files later to ensure files are not duplicated across containers and wasting storage. The relative URLs to each chart within the dashboard are now predictable and the URL base is already known (already queried from the Cryostat backend). There is a standardized `<iframe>` embeddable chart format that Grafana exposes, which can be wrapped into a Dashboard card. The card configuration wizard should display a limited selection of metrics that the user may be interested in, for example: CPU usage, memory usage, allocations, exception count. When the user chooses one of these selections the card is created by mapping that selection to a chart ID and creating the card config. The card component uses the chart ID to construct the `<iframe>` element with the rest of the known/computed Grafana URL and simply displays it.

How to manage loading JFR data into the datasource is the next biggest question. This should only be done if the user already has at least one Grafana card on their dashboard. It should also be done at most once, regardless of how many Grafana cards are present. If the selected target has no recordings to load into the datasource then the user should be informed and prompted. If there are active recordings then those should be used for the source data, otherwise fall back to the most recent archived data. This looks very similar to the way the Automated Reports analysis card already handles the same situation, so perhaps the same logic should be implemented or reused here.

PR #856, opened 2023-02-03
feat(dashboard): implement embedded chart cards

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #847
- [x] Depends on https://github.com/cryostatio/cryostat-grafana-dashboard/pull/20
- [x] Depends on https://github.com/cryostatio/cryostat-grafana-dashboard/pull/21
- [x] Depends on https://github.com/cryostatio/cryostat/pull/1359

## Description of the change:
This adds a dashboard card implementation that can embed various Grafana metrics charts.

## Motivation for the change:
Users do not need to use the `View in Grafana...` action on a specific recording to view metrics. Instead, they can simply configure a dashboard with the metrics that interest them, and select a target application. The metrics layout is as the user wishes, the source recording is started as needed, and the data is reloaded on a schedule to provide near-realtime metrics display.

image


andrewazores:
Ready for review.

tthvo:
Looks aweesomeee! I notice there are some issue with cards that has white texts in light theme. For example:

Screenshot from 2023-02-08 19-07-43

I think its an option in grafana dashboard json that sets this...

tthvo:
I think you mentioned this before there seems to be an issue with drag-drop and resize for these cards. Notice how the title is clickable. Though there is some space above it to click and drag, its still pretty hard. Also, for plots, resizing is quite odd as dragging "jump into" the time range selection.

Screenshot from 2023-02-08 19-12-18

maxcao13:
image
 
Looks a little strange to have the create recording message have overflow. 

Also on that note, should the Automated Analysis Card be refactored to have the same walkthrough as these create recording steps?

maxcao13:
image

Also I can't seem to get Firefox to unblock this. I tried manually using the `grafana.ini` changes and rebuilding latest `cryostat-grafana-dashboard` and changing smoketest but that still doesn't work. Not sure if I am missing something...

tthvo:
> Also I can't seem to get Firefox to unblock this. I tried manually using the `grafana.ini` changes and rebuilding latest `cryostat-grafana-dashboard` and changing smoketest but that still doesn't work. Not sure if I am missing something...

I think you need some plugin to pass by that constraint. Maybe try Chrome. I think it should work?

maxcao13:
> > Also I can't seem to get Firefox to unblock this. I tried manually using the `grafana.ini` changes and rebuilding latest `cryostat-grafana-dashboard` and changing smoketest but that still doesn't work. Not sure if I am missing something...
> 
> I think you need some plugin to pass by that constraint. Maybe try Chrome. I think it should work?

Right, I used Chromium with Ignore X - Frame headers extension and it renders correctly now, thanks!

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,360 @@
...
+        maxAge: 120, // seconds
+        maxSize: 100 * 1024 * 1024, // bytes
+      } as CreateRecordingProps,
+    });
+  }, [history]);
+
+  return (
+    <>
+      <DashboardCard
+        key={idx}
```
maxcao13, line 264: Not really sure why, but having this key re-renders the card every time you resize the card...

src/app/Settings/ChartCardsConfig.tsx
```diff
@@ -0,0 +1,100 @@
...
+      setMinRefresh(value);
+      context.settings.setChartControllerConfig({ minRefresh: value });
+    },
+    [minRefresh, setMinRefresh, context.settings]
+  );
+
+  const handleVisibleStep = React.useCallback(
+    (delta: number) => () => {
+      const v = minRefresh + delta;
+      context.settings.setChartControllerConfig({ minRefresh: v });
```
tthvo, line 68: Missing setState? Clicking `-` or `+` does not update input :D

src/app/Settings/ChartCardsConfig.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const Component = () => {
+  const [t] = useTranslation();
+  const context = React.useContext(ServiceContext);
+
+  const [minRefresh, setMinRefresh] = React.useState(context.settings.chartControllerConfig().minRefresh);
+
+  const handleChange = React.useCallback(
+    (evt) => {
+      let value = isNaN(evt.target.value) ? minRefresh : Number(evt.target.value);
```
tthvo, line 55: Since we always change state, we should wrap out this part into `setMinRefresh((old) => ...)` to access current state safely.

src/app/CreateRecording/CustomRecordingForm.tsx
```diff
@@ -397,6 +419,15 @@ export const CustomRecordingForm: React.FC<CustomRecordingFormProps> = ({ prefil
             onChange={handleRecordingNameChange}
             validated={nameValid}
           />
+          <Checkbox
+            label="Restart if recording already exists"
```
tthvo, line 423: Localize? Also its aria-label too?

src/app/Dashboard/AddCard.tsx
```diff
@@ -301,6 +301,8 @@ const PropsConfigForm = ({ onChange, ...props }: PropsConfigFormProps) => {
               onChange={handleNumeric(ctrl.key)}
               onPlus={handleNumericStep(ctrl.key, 1)}
               onMinus={handleNumericStep(ctrl.key, -1)}
+              min={ctrl?.extras?.min}
```
tthvo, line 304: There is no need to check undefined on `ctrl`? Just on `extra` is fine I think.

src/app/Dashboard/Charts/ChartController.tsx
```diff
@@ -0,0 +1,190 @@
...
+    private readonly _notifications: NotificationChannel,
+    private readonly _settings: SettingsService
+  ) {
+    this._subscriptions.push(
+      this._refCount$
+        .pipe(
+          map((v) => v > 0),
+          pairwise()
+        )
+        .subscribe((v) => {
```
tthvo, line 80: Looks like we can use array deconstruct to be more intuitive

```typescript
.subscribe(([prev, curr]) => {
        if (prev && !curr) {
          // last subscriber left
          this._tearDown();
        }
        if (!prev && curr) {
          // first subscriber joined
          this._init();
        }
      })
```

src/app/Dashboard/Charts/ChartController.tsx
```diff
@@ -0,0 +1,190 @@
...
+  switchMap,
+  throttleTime,
+} from 'rxjs';
+
+export const RECORDING_NAME = 'dashboard_metrics';
+
+export class ChartController {
+  private readonly _refCount$ = new BehaviorSubject<number>(0);
+  private readonly _updates$ = new ReplaySubject<void>(1);
+  private readonly _hasRecording$ = new ReplaySubject<boolean>(1);
```
tthvo, line 65: Wondering any reason to use `RelaySubject(1)` over `BehaviorSubject`? Maybe I am wrong but they are pretty behaving the same?

tthvo: Ahh nvmind it does not need an initial value and emits on close XD

src/app/Dashboard/Charts/ChartController.tsx
```diff
@@ -0,0 +1,190 @@
...
+        map((resp) => {
+          const nodes = resp?.data?.targetNodes;
+          if (nodes.length === 0) {
+            return false;
+          }
+          const count = nodes[0]?.recordings?.active?.aggregate?.count;
+          return count > 0;
+        }),
+        catchError((_) => of(false)),
+        first()
```
tthvo, line 142: I think we can remove `first` here as `context.api.graphql` internally calls it already?

src/app/Dashboard/Charts/ChartController.tsx
```diff
@@ -0,0 +1,190 @@
...
+                }
+              }
+            }
+            }
+          }`,
+        { connectUrl: target.connectUrl }
+      )
+      .pipe(
+        map((resp) => {
+          const nodes = resp?.data?.targetNodes;
```
tthvo, line 134: Following `CountResponse` interface, do we need to check undefined here?

src/app/Dashboard/Dashboard.tsx
```diff
@@ -270,26 +310,33 @@ export const Dashboard: React.FC<DashboardProps> = (_) => {
...
-                  onRemove={() => handleRemove(idx)}
-                  onResetSize={() => handleResetSize(idx)}
-                />,
-              ],
-            })}
+      <ChartContext.Provider value={{ controller: chartController.current }}>
+        <Grid id={'dashboard-grid'} hasGutter>
+          {cardConfigs.map((cfg, idx) => (
+            <GridItem span={cfg.span} key={cfg.id} order={{ default: idx.toString() }}>
+              {hasConfigByName(cfg.name) ? (
```
tthvo, line 317: Maybe this check should be applied the outer GridItem since this will add an empty GridItem that has some padding causing larger gaps :D

tthvo: This comes from messing around with local storage.

Screenshot from 2023-02-09 00-33-14

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,360 @@
...
+          ) : (
+            <Bullseye>
+              <EmptyState variant={EmptyStateVariant.large}>
+                <EmptyStateIcon icon={DataSourceIcon} />
+                <Title headingLevel="h2" size="md">
+                  No source recording
+                </Title>
+                <EmptyStateBody>
+                  Metrics cards display data taken from running flight recordings with the label{' '}
+                  <code>origin={RECORDING_NAME}</code>. No such recordings are currently available.
```
tthvo, line 284: We can replace `<code />` with PF label here.

```typscript
<Label color='blue' isCompact>origin={RECORDING_NAME}</Label>
```

Screenshot from 2023-02-09 01-08-49

locales/en/public.json
```diff
@@ -55,6 +55,11 @@
       "LANGUAGE_REGION": "Language & Region",
       "NOTIFICATION_MESSAGE": "Notifications & Messages"
     },
+    "CHARTS_CONFIG": {
+      "DESCRIPTION": "Set configuration options for the behaviour of Metrics cards on the Dashboard.",
```
tthvo, line 59: Wondering if we should set this to empty? Right now, it looks a little odd. I guess the title is already self-explained and description is already available for each config group?

Screenshot from 2023-02-09 01-20-36

src/app/Dashboard/Charts/ChartController.tsx
```diff
@@ -0,0 +1,190 @@
...
+          this._notifications.messages(NotificationCategory.ActiveRecordingStopped)
+        ).pipe(switchMap((_) => this._target.target().pipe(first()))),
+        this._target.target()
+      )
+        .pipe(concatMap((t) => this._hasRecording(t)))
+        .subscribe((v) => {
+          this._hasRecording$.next(v);
+          if (v) {
+            this._api
+              .uploadActiveRecordingToGrafana(RECORDING_NAME)
```
tthvo, line 162: I am running into a bunch of error messages when switching between target while the metrics cards refresh. I think because internally `uploadActiveRecordingToGrafana` subscribe to `context.target.target` but the incoming value for previous target.

Screenshot from 2023-02-09 01-41-26

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,360 @@
...
+    }
+  }, [chartSrc, dashboardUrl]);
+
+  const cardStyle = React.useMemo(() => {
+    return {
+      height: getHeight(props.chartKind),
+    };
+  }, [props.chartKind]);
+
+  const expandButton = React.useMemo(() => {
```
tthvo, line 183: `syncButton`?

src/app/Dashboard/Dashboard.tsx
```diff
@@ -270,26 +310,33 @@ export const Dashboard: React.FC<DashboardProps> = (_) => {
...
-              dashboardId: idx,
-              actions: [
-                <DashboardCardActionMenu
-                  key={`${cfg.name}-actions`}
-                  onRemove={() => handleRemove(idx)}
-                  onResetSize={() => handleResetSize(idx)}
-                />,
-              ],
-            })}
+      <ChartContext.Provider value={{ controller: chartController.current }}>
```
tthvo, line 313: Context value here should be wrapped in `React.useMemo`. Right now, every time the card is re-order or resize, context value is always a different object -> lots of observable is being re-subscribed.

I think this might be the issue for https://github.com/cryostatio/cryostat-web/pull/856#discussion_r1100888788.

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+  const actions = React.useMemo(() => {
+    const a = props.actions || [];
+    return [expandButton, popoutButton, ...a];
+  }, [props.actions, expandButton, popoutButton]);
+
+  const header = React.useMemo(() => {
+    const isWide = props.span > 4;
+    const style = {
+      marginBottom: isWide ? '-2.8em' : '',
```
tthvo, line 221: Maybe a smaller number to leave more space for dragging? `2em` seems like a better fit?

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+  const popout = React.useCallback(() => {
+    if (chartSrc && dashboardUrl) {
+      window.open(chartSrc, '_blank');
+    }
+  }, [chartSrc, dashboardUrl]);
+
+  const cardStyle = React.useMemo(() => {
+    return {
+      height: getHeight(props.chartKind),
```
tthvo, line 179: On this note: https://github.com/cryostatio/cryostat-web/pull/856#issuecomment-1423456099

Not sure if I am correct but quick look into their stylings, the embedded Grafana panels seems to have an intrinsic width and height. This causes scrolling in small parent container.

Later we have height-resizable card, then this wont be a problem. But for now, I am thinking of this:

```typescript

function getHeight(kind: string): number {
  switch (kind) {
    case 'Core Count':
      return 250; // Increase this to 250 seems to work pretty well
    default:
      return 380;
  }
}

....

const cardStyle = React.useMemo(() => {
    return hasRecording? { 
      height:  getHeight(props.chartKind),
    }: {}; // Let the empty text decides its height
  }, [props.chartKind, hasRecording]);
```

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,360 @@
...
+    };
+  }, [props.chartKind]);
+
+  const expandButton = React.useMemo(() => {
+    return (
+      <>
```
tthvo, line 185: React fragment is missing a key (React is throwing a warning). Here, we can just remove it. Same for `popoutButton`.

src/app/Dashboard/AddCard.tsx
tthvo: Also, is it possible that we can remove a `None` selection or maybe make a field as required? Right now, select any option and then reselect the placeholder really does not do anything...

tthvo (commented):
Looking good! Just couple comments above. 

Also, i noticed that `iframe`are being reinitialized when reordering. Came across this one

https://github.com/facebook/react/issues/858

Not sure if we can do anything about that. I tried React.useMemo but still not really helping.

andrewazores:
> Also on that note, should the Automated Analysis Card be refactored to have the same walkthrough as these create recording steps?

I was wondering the same thing. I'm not sure which approach is a better UX. I think your approach in the analysis card is nice because it allows more explicit settings for the defaults used when creating a source recording, but the approach I used here reuses the existing UI flow and has less code to maintain.

src/app/Dashboard/Charts/ChartController.tsx
andrewazores: Huh. I thought I had solved that. I'll try to reproduce it again.

src/app/Dashboard/AddCard.tsx
andrewazores: > Also, is it possible that we can remove a `None` selection or maybe make a field as required? Right now, select any option and then reselect the placeholder really does not do anything...

That seems like a good idea. There's a lot we can do to these configs to make the generated form behave better. Probably better for a separate PR.

src/app/Dashboard/Charts/ChartController.tsx
andrewazores: Are you still seeing this with the latest changes?

maxcao13:
> 

I realized after going that I wasn't removing the `pull=always` variable in `smoketest` it looks as intended now. Nothing wrong just me.

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,367 @@
...
+      id={props.chartKind + '-chart-card'}
+      dashboardId={props.dashboardId}
+      cardSizes={ChartCardSizes}
+      isCompact
+      style={cardStyle}
+      cardHeader={header}
+    >
+      <CardBody>
+        {hasRecording ? (
+          <iframe key={key} style={{ height: '100%', width: '100%' }} src={chartSrc} />
```
maxcao13, line 277: Maybe have a `LoadingView` component render during the card chart init when a state `isLoading` is true? I think this is possible with the `onLoad` prop on the `iframe` component, the only problem is that I believe the onload requires the iframe to be rendered first so we cant do `isLoading ? <LoadingView /> : <iframe ... >`. Maybe hiding the iframe with display none may be better until it is finished loading.

andrewazores: The `onLoad` of an `iframe` fires when the source document has loaded, but since the Grafana chart component inside the frame has to also execute a bunch of JS and load data async from the datasource, I think the `<LoadingView>` would only appear for a brief flash until the embedded HTML has been loaded by the browser. I can give it a try and see how it feels in practice - I guess for slower connections between the browser and the Grafana dashboard in the cluster, this might still be visible for a while.

maxcao13 (commented):
I'm not sure what to do about resizing. Because when you resize to the left, the mouse then hovers on the iframe contents and overrides the resizing. I'm thinking maybe implementing that the dragging bar actually sticks to your mouse but when you let go, it snaps back to the gridSpan size. I think that would solve this issue.

maxcao13:
Also, maybe a stop and start button may be nice. So that the user can stop the graph if they so please if they notice something.

andrewazores:
css z-index might also be helpful for the resizing drag/drop thing, maybe

src/app/Dashboard/Charts/ChartController.tsx
tthvo: Ohhh I dont see it anymore. Nicee! Not sure what happened tho.

maxcao13:
> css z-index might also be helpful for the resizing drag/drop thing, maybe
z-index does not quite work, but using `pointer-events: none` solves the issue, but then users cannot react with the card by itself? Worst case is maybe just not allow any interactions with the iframe from the overhead view, and only if you open it in a new tab or fullscreen.

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,367 @@
...
+      let next = prev + 1;
+      if (next >= 10) {
+        next = 0;
+      }
+      return next;
+    });
+  }, [setKey]);
+
+  React.useEffect(() => {
+    addSubscription(serviceContext.target.target().subscribe((_) => updateKey()));
```
tthvo, line 128: Any bugs that cause us to force listen and force iframe reload since we already listen to this in ChartController?

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,367 @@
...
+  }, [props.actions, resyncButton, popoutButton]);
+
+  const header = React.useMemo(() => {
+    const isWide = props.span > 4;
+    const style = {
+      marginBottom: isWide ? '-2em' : '',
+    };
+    if (hasRecording) {
+      return (
+        <>
```
tthvo, line 230: Remove Fragment here? we probably wont have multiple headers but just for safety on the key issue.

andrewazores:
Tests are broken in the last commit because I'm having a hard time getting the `iframe onLoad` to trigger within the unit tests, so the actual card content is rendered rather than just the loading spinner. I've pushed it as it is regardless so you two can test out how the spinner works/feels before I dump any more time into figuring out the tests.

andrewazores:
> > css z-index might also be helpful for the resizing drag/drop thing, maybe
> 
> z-index does not quite work, but using `pointer-events: none` solves the issue, but then users cannot react with the card by itself? Worst case is maybe just not allow any interactions with the iframe from the overhead view, and only if you open it in a new tab or fullscreen.
> 
> EDIT: I think I can still keep trying however, there are some stackoverflow questions about exactly this: https://stackoverflow.com/questions/5261328/receive-mousemove-events-from-iframe-too

Disabling mouse events on the iframe is probably okay too. The sync button is mostly there right now just for providing a way to reset the data range after clicking and zooming in anyway, so it's already clunky. Removing the mouse event handling would remove the need for that as well as make the drag/resize behave nicely.

But, then we might have a weird difference where the mini iframe version has the play/pause button you suggested and the maximized version in a new tab does not...

src/app/Dashboard/Charts/ChartCard.tsx
andrewazores: The latest changes have a `<LoadingView>` but it triggers and is removed separately from the iframe loading cycle. It now actually appears when the controller is in the middle of determining whether the selected target has a suitable source recording. If the controller determines that it does then the card component should just about immediately mount the iframe and remove the loading spinner, so there may possibly be a lag time in between the loading spinner disappearing and any content actually rendering within the iframe. But this solution was also nicely unit testable, so I'm going with it for now. I can piece something together later to bridge that gap longer if we think it's needed and I figure out that onload event for tests.

tthvo:
Loading state is quite quick to see but by simulating low network, I am seeing a grafana loading screen already built into the iframe. So, we can just use that?

Screenshot from 2023-02-09 15-53-50

maxcao13:
> Tests are broken in the last commit because I'm having a hard time getting the `iframe onLoad` to trigger within the unit tests, so the actual card content is rendered rather than just the loading spinner. I've pushed it as it is regardless so you two can test out how the spinner works/feels before I dump any more time into figuring out the tests.

Seems like sometimes it either takes an excessively long time to load, or doesn't load at all...

> Tests are broken in the last commit because I'm having a hard time getting the `iframe onLoad` to trigger within the unit tests, so the actual card content is rendered rather than just the loading spinner. I've pushed it as it is regardless so you two can test out how the spinner works/feels before I dump any more time into figuring out the tests.

I guess since Grafana already does some sort of loading state, then we don't really need this then. Just looks a bit clunky but the react loading doesn't really play nice with the iframe loading.

andrewazores:
> Loading state is quite quick to see but by simulating low network, I am seeing a grafana loading screen already built into the iframe. So, we can just use that?
> 
> Screenshot from 2023-02-09 15-53-50

Yea, but I think there would still be some time delay in between our loading spinner disappearing and this appearing if the connection is really super slow and the iframe document takes a long time to load. But if the connection is that slow then probably most things about the web-client will feel unusably unresponsive anyway, so this is probably not a big concern.

src/app/Dashboard/Charts/ChartCard.tsx
tthvo: Oopps I miss this comment. Make sense to me to keep the loading state.

src/test/Dashboard/Charts/ChartCard.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  it('renders empty state with information and action button', async () => {
+    jest.spyOn(mockController, 'attach').mockReturnValue(of(ControllerState.NO_DATA));
+
+    renderChartCard(
+      <ChartCard theme={'light'} chartKind={'CPU Load'} duration={120} period={10} span={6} dashboardId={0} />
+    );
+
+    expect(screen.getByText('CPU Load')).toBeInTheDocument();
+    expect(screen.getByText('CHART_CARD.NO_RECORDING.TITLE')).toBeInTheDocument();
```
tthvo, line 146: Hmmm a little odd here. I thought the actual value must be showing instead of keys?

src/test/Dashboard/Charts/ChartCard.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  it('navigates to recording creation with prefilled state when empty state button clicked', async () => {
+    jest.spyOn(mockController, 'attach').mockReturnValue(of(ControllerState.NO_DATA));
+
+    const { user } = renderChartCard(
+      <ChartCard theme={'light'} chartKind={'CPU Load'} duration={120} period={10} span={6} dashboardId={0} />
+    );
+
+    expect(history.location.pathname).toBe('/');
+    await user.click(screen.getByRole('button', { name: /create/i }));
```
tthvo, line 159: We should use the localized version of the button text to search? Same for other places.

src/test/Dashboard/Charts/ChartCard.test.tsx
tthvo: Same other aria-labels too...

maxcao13:
> > > css z-index might also be helpful for the resizing drag/drop thing, maybe
> > 
> > 
> > z-index does not quite work, but using `pointer-events: none` solves the issue, but then users cannot react with the card by itself? Worst case is maybe just not allow any interactions with the iframe from the overhead view, and only if you open it in a new tab or fullscreen.
> > EDIT: I think I can still keep trying however, there are some stackoverflow questions about exactly this: https://stackoverflow.com/questions/5261328/receive-mousemove-events-from-iframe-too
> 
> Disabling mouse events on the iframe is probably okay too. The sync button is mostly there right now just for providing a way to reset the data range after clicking and zooming in anyway, so it's already clunky. Removing the mouse event handling would remove the need for that as well as make the drag/resize behave nicely.
> 
> But, then we might have a weird difference where the mini iframe version has the play/pause button you suggested and the maximized version in a new tab does not...

Figured a good solution.

Set an id to the iframe
```js
<iframe id="some-id" key={key} style={{ height: '100%', width: '100%' }} src={chartSrc} onLoad={handleIFrameLoaded} />
```

In both the `draggableRef` class mouse move callbacks (`callbackMouseMove` and `onDragMove`)  have this:
```js
   if (cardRef.current) {
        const a = document.getElementById('some-id');
        if (a !== null) {
          a.setAttribute('style', `pointer-events: none; height: 100%; width: 100%;`);
        }
```
or some sort of iteration of that where we just set the style of the iframe to `pointer-events: none` during drag moving or resize moving. Probably better to do this with adding classNames as well.

src/test/Dashboard/Charts/ChartCard.test.tsx
tthvo:
```log
    console.warn
      react-i18next:: You will need to pass in an i18next instance by using initReactI18next

      106 |
      107 | export const ChartCard: React.FC<ChartCardProps> = (props) => {
    > 108 |   const [t] = useTranslation();
          |                             ^
      109 |   const serviceContext = React.useContext(ServiceContext);
      110 |   const controllerContext = React.useContext(ChartContext);
      111 |   const history = useHistory();

      at warn (node_modules/react-i18next/dist/commonjs/utils.js:22:31)
      at warnOnce (node_modules/react-i18next/dist/commonjs/utils.js:35:8)
      at useTranslation (node_modules/react-i18next/dist/commonjs/useTranslation.js:44:25)
      at ChartCard (src/app/Dashboard/Charts/ChartCard.tsx:108:29)
      at renderWithHooks (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:5552:18)
      at mountIndeterminateComponent (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:7918:13)
      at beginWork (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:9019:16)
      at performUnitOfWork (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:12649:12)

    console.warn
      react-i18next:: You will need to pass in an i18next instance by using i18nextReactModule
```

src/test/Dashboard/Charts/ChartCard.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+  defaultServices.api,
+  defaultServices.target,
+  defaultServices.notificationChannel,
+  defaultServices.settings
+);
+const mockChartContext = {
+  controller: mockController,
+};
+
+describe('<ChartCard />', () => {
```
tthvo, line 77: I think we should explicitly calling `cleanup` in `afterEach` and reset `history` too.

src/test/Dashboard/Charts/ChartCard.test.tsx
tthvo: Ahh I figured out that `Common.tsx` initialized the test i18n module. So, without importing it, this happens. Maybe we can move the `renderChartCard` to `Common.tsx`?

src/test/Dashboard/Charts/ChartCard.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+    chartContext = mockChartContext,
+    preloadState = {},
+    store = setupStore(preloadState),
+    user = userEvent.setup(),
+    ...renderOptions
+  } = {}
+) => {
+  const Wrapper = ({ children }: PropsWithChildren<unknown>) => {
+    return (
+      <ChartContext.Provider value={chartContext}>
```
tthvo, line 219: We should probably wrap this around children only to keep the order that appears in the actually app.

src/test/Dashboard/Charts/ChartCard.test.tsx
andrewazores: Makes sense, though I don't think `renderChartCard` should go directly into `Common`. Maybe there is some more generic refactoring that I can apply first so that the function can be reused and is actually "common". It is pretty much like all those others, except with the chart controller added.

tthvo (commented):
I am wondering how we are going to handle multiple accesses from different web console instances (2 tabs or 2 people looking at cryostat).  For example, one person is looking at `9093` and another person enters dashboard and choose another target `9091`. This will override datasource data. 

This will make the first person dashboard outdated and rendering empty data at some point?

andrewazores:
> I am wondering how we are going to handle multiple accesses from different web console instances (2 tabs or 2 people looking at cryostat). For example, one person is looking at `9093` and another person enters dashboard and choose another target `9091`. This will override datasource data.
> 
> This will make the first person dashboard outdated and rendering empty data at some point?

Correct. This is just a limitation of the stateful jfr-datasource model we have right now. In the feature documentation we will need to call this out and explain it to the user. Maybe for this reason it also makes sense to downgrade the feature level to BETA.

In the future when we have a smarter jfr-datasource that can handle multiple source recordings and integration with Grafana panels (or custom charts) that know how to perform queries across different sources, then this problem should basically just disappear as a consequence.

tthvo:
Oh cool make sense! Do you plan to fix the drag and resize issue here too?

andrewazores:
> Oh cool make sense! Do you plan to fix the drag and resize issue here too?

Yes, I'll do that next after I finish up with trying to genericize and refactor out the `renderChartCard` function.

tthvo (approved):
Looks very niceee to me! The resize and drag is working as expected :D

locales/en/public.json
```diff
@@ -12,6 +12,23 @@
     "OPEN_SOURCE_LICENSE": "Open Source License",
     "VERSION": "version"
   },
+  "CHART_CARD": {
+    "BUTTONS": {
+      "CREATE": {
+        "LABEL": "Create"
```
maxcao13, line 18: I notice these aria-labels have a different scheme from some other aria labels in the file
e.g.
```json
  "DateTimePicker": {
    "ARIA_LABELS": {
      "DISPLAY_SELECTED_DATETIME": "Displayed selected datetime",
      "TABS": "Select a date or time tab"
    },
    "SELECTED_DATETIME": "Selected DateTime"
  },
  "MeridiemPicker": {
    "ARIA_LABELS": {
      "LISTBOX": "Select AM or PM"
    }
  },
```
Maybe either change this or we should change the other ones to keep consistency?

tthvo: We probably need to document key format somewhere too

src/app/Dashboard/Charts/ChartCard.tsx
```diff
@@ -0,0 +1,358 @@
...
+  },
+  height: {
+    // TODO: implement height resizing
+    minimum: Number.NaN,
+    default: Number.NaN,
+    maximum: Number.NaN,
+  },
+};
+
+export const ChartCardDescriptor: DashboardCardDescriptor = {
```
maxcao13, line 310: Should all of this be localized as well?

tthvo:
>  In the feature documentation we will need to call this out and explain it to the user. 

Not for this PR, I am thinking of something like a banner to show on web console when in BETA mode like this:

Screenshot from 2023-02-09 20-52-44

Or restyle with inline alert: https://www.patternfly.org/v4/components/alert#inline-alerts-variants

maxcao13:
> Looking good! Just couple comments above.
> 
> Also, i noticed that `iframe`are being reinitialized when reordering. Came across this one
> 
> [facebook/react#858](https://github.com/facebook/react/issues/858)
> 
> Not sure if we can do anything about that. I tried React.useMemo but still not really helping.

It is indeed slightly annoying, but only because the re-ordering functionality is tied in with the order that the card elements are being attached to the DOM. It doesn't have to actually be this way since Patternfly grid uses the grid-order for its ordering which is just a css attribute, however I implemented the drag drop keeping the JS list order in mind which it is just a bit more intuitive. There will just need to a bit of refactoring with how the drag and drop works, but its probably not horrible. If you guys think this is important, I will look into a patch.

src/app/Dashboard/Charts/ChartCard.tsx
andrewazores: Probably, but I thought that made sense to do as a separate refactoring since it applies to these cards as well as the analysis card and the two placeholder cards.

locales/en/public.json
andrewazores: You mean just change the way that the key is split up? Or change the actual localized text?

andrewazores:
> > In the feature documentation we will need to call this out and explain it to the user.
> 
> Not for this PR, I am thinking of something like a banner to show on web console when in BETA mode like this:
> 
> Screenshot from 2023-02-09 20-52-44
> 
> If the description is too long, we can just link to upstream docs or dev posts.

Sounds good. That's more along the lines of how the old OSIO project did feature flagging IIRC, which is what inspired the feature flags here too.

andrewazores:
> > Looking good! Just couple comments above.
> > Also, i noticed that `iframe`are being reinitialized when reordering. Came across this one
> > [facebook/react#858](https://github.com/facebook/react/issues/858)
> > Not sure if we can do anything about that. I tried React.useMemo but still not really helping.
> 
> It is indeed slightly annoying, but only because the re-ordering functionality is tied in with the order that the card elements are being attached to the DOM. It doesn't have to actually be this way since Patternfly grid uses the grid-order for its ordering which is just a css attribute, however I implemented the drag drop keeping the JS list order in mind which it is just a bit more intuitive. There will just need to a bit of refactoring with how the drag and drop works, but its probably not horrible. If you guys think this is important, I will look into a patch.

File an issue for it and we'll keep it as a lower priority item to tackle if there is time for this release.

locales/en/public.json
maxcao13: I mean the way the key is split up. Maybe we should have 
```json
"DateTimePicker": {
    "TABS": {  
          "LABEL": "Select a date or time tab"
     }
```
in the same format in this PR that you have done here. If so, then that's okay to change in a separate refactor PR.

Just keeping a consistent key format.

andrewazores: Sure, I don't really mind either way. I think there probably are/will be instances where the `aria-label` matches some text that is visually rendered, so specifying `ARIA_LABELS` as part of the key maybe doesn't make the most sense.

maxcao13 (approved):
Everything looks good to me then! Great work.

andrewazores:
Will merge as soon as unblocked on https://github.com/cryostatio/cryostat-grafana-dashboard/pull/20 .

Issue #839, opened 2023-01-31
[Task] Improve user experience for Automated Analysis Card


maxcao13:
## Describe the feature
It was mentioned in a meeting that these things would be nice for the AutomatedAnalysis Card to have.

![image](https://user-images.githubusercontent.com/69827283/215833562-ceddd60f-1a21-4445-93ac-28692bbae089.png)

> Automated Analysis Card with one result showing

Dynamic layout of scores:
- When the number of visible event results drops below a number (hardcoded number? 5? 10?) , show inline description of the important results 
- Localize result category names
- Localize analysis card strings
- Others?...

Issue #751, opened 2022-12-13
[Task] AutomatedAnalysisConfigForm should be flexible and reusable


tthvo:
## Describe the task

> That's part of what I meant about this component not being properly set up right now. There should be a switch to hide or reveal these controls, but they also need to be refactored so that they can be used separately/outside of a `<Form>`. The `Create` button probably shouldn't be there and the `Save Configuration` should probably just happen every time a change occurs rather than requiring the user to click a button so that they aren't having to save within the config screen and then also separately press `Finish` on the Wizard. Or, the `Finish` button from the wizard needs to be hooked into the `Save Configuration` action as well. Seems out of scope of this particular PR - more to do with the AA card than the dashboard/wizard.

## Any other information?
_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-web/pull/747#discussion_r1046388820_

PR #862, opened 2023-02-07
chore(aa): automated analysis card refactoring

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #751
Fixes: #839

## Description of the change:
Changes include:
* [x] Refactoring `AutomatedAnalysisConfigForm` such that it no longer needs buttons to save the config to local storage
* [x] Adding translation keys to various needed areas of `AutomatedAnalysisCard` and related components
* [x] Adding an inline view of Automated Analysis results
* [x] Fixed rendering bug with the AA Card where filters and toolbar render and flicker before analysis is fetched

Misc:
* [x] Temporary default localized keys now default with the value `cryostat_tmp`. This is so it's easier to search for the string and find the default keys in the file to fix them.  
* [x] Change missing import eslint rule:  `error` -> `warn`
* [x] Exclude `dist` from jest testing

In draft mode. Full changes and tests have not been fixed yet.

## How to manually test:
1. Run cryostat normally with the github action.
2. Add the AutomatedAnalysisCard with the Dashboard AddCard feature and try the configuration editing as well.
3. Create an automated analysis recording and switch the switch to list view to see the new view.


maxcao13:
Not sure the CI fails because of a snapshot update... `EventTypes.test.tsx` does not need an update on my local.

src/test/Events/__snapshots__/EventTypes.test.tsx.snap
```diff
@@ -399,7 +399,7 @@ Array [
             aria-label="Details"
             aria-labelledby="simple-node0 expandable-event-type-row-00"
             className="pf-c-button pf-m-plain"
-            data-ouia-component-id="OUIA-Generated-Button-plain-1"
+            data-ouia-component-id="OUIA-Generated-Button-plain-5"
```
tthvo, line 402: Maybe revert this?

maxcao13: Interestingly, this was changed in https://github.com/cryostatio/cryostat-web/pull/858, but this PR reverts it back. I'm really not sure why.

tthvo: hmm all dependencies are up-to-date? like `yarn install` and all?

andrewazores:
While we're in here, I think the default settings could use updating:

image

Unlimited time but size-based probably makes sense, but 2048b is way too small to be meaningful. I think this should be more like 5 or 10 MiB at least.

andrewazores:
The number input also only works with whole numbers, but scores don't have to be.

image

In particular, notice that there are results with scores between 0 and 1, so I'm unable to use the score filter to remove exact 0s and leave in everything else.

maxcao13:
> The number input also only works with whole numbers, but scores don't have to be.
> 
> image
> 
> In particular, notice that there are results with scores between 0 and 1, so I'm unable to use the score filter to remove exact 0s and leave in everything else.

I dont know if using the value input thing on the right of the slider is able to increase and decrease by specified steps (instead of 1), there doesnt seem to be support for it. Instead you can change fine grained values by using the left and right arrow keys after focusing the slider button.

andrewazores:
What if `Math.ceil` is applied to the scores when checking if they meet the threshold set by the slider? `Math.ceil(0) === 0` but `Math.ceil(0.1) === 1`, so this way if I set the slider value to 1 then it will basically capture `(0, 100]` instead of `[1, 100]`. Maybe that's close enough to the same thing to enable that "filter out the 0s" use case.

maxcao13:
> What if `Math.ceil` is applied to the scores when checking if they meet the threshold set by the slider? `Math.ceil(0) === 0` but `Math.ceil(0.1) === 1`, so this way if I set the slider value to 1 then it will basically capture `(0, 100]` instead of `[1, 100]`. Maybe that's close enough to the same thing to enable that "filter out the 0s" use case.

Hmm... I still think it may be better to keep it this way because then users can explicitly set filters with 1 precision floating point which matches what the scores are rounded to. Doing it with what you suggested may not be very intuitive and user friendly in my opinion. We can just say that using the left and right arrow keys will change the values by 1 decimal point if its hard to specifically drag the slider to the correct number you want. You can also type the exact number you want in the value input element as well.

I could also just round the scores themselves to integers to remove this problem, which is an option.

andrewazores:
> I could also just round the scores themselves to integers to remove this problem, which is an option.

Yes, but I think it's worthwhile to keep the score values the same as what the user will see if they load the same recording file into JMC (of the same version that backs the report generation), even if the rounding would only seem like a small change in value.

andrewazores:
Anyway, that said, I just tested out the latest changes and I think how it works is good enough.

andrewazores (commented):
Looks and feels good. I skimmed the patch and it looks reasonable.

locales/en/public.json
```diff
@@ -12,6 +12,56 @@
...
+    "CHECKBOX_LABEL_SHOW_NA_SCORES": "Show N/A scores",
+    "CLEAR_FILTERS": "Clear all filters",
+    "CRITICAL_RESULTS_one": "{{count}} Critical Result",
+    "CRITICAL_RESULTS_other": "{{count}} Critical Results",
+    "ERROR_TEXT": "Cryostat was unable to generate an automated analysis report.",
+    "ERROR_TITLE": "Automated Analysis Error",
+    "GOOD_RESULTS": "No problems",
+    "NO_RESULTS": "No Results Found",
+    "NO_RESULTS_BODY": "No results match this filter criteria. Try removing filters, or resetting the severity score filter to 0 to show results",
+    "RESULT": "Name",
```
tthvo, line 33: Is this key used anywhere?

locales/en/public.json
```diff
@@ -12,6 +12,56 @@
...
+    "ARIA_LABELS": {
+      "DELETE_AA": "Delete automated analysis",
+      "REFRESH_AA": "Refresh automated analysis",
+      "TOOLBAR": "Automated Analysis Toolbar"
+    },
+    "CARD_DESCRIPTION": "Assess common application performance and configuration issues",
+    "CARD_DESCRIPTION_FULL": "Creates a recording and periodically evaluates various common problems in application configuration and performance. Results are displayed with scores from 0-100 with colour coding and in groups. This card should be unique on a dashboard",
+    "CARD_TITLE": "Automated Analysis",
+    "CHECKBOX_LABEL_SHOW_NA_SCORES": "Show N/A scores",
+    "CLEAR_FILTERS": "Clear all filters",
```
tthvo, line 25: Maybe to `common.js` so we can reuse in other tables?

src/app/app.css
```diff
@@ -153,6 +173,12 @@ html, body, #root {
   text-overflow: ellipsis;
 }
 
+.pf-c-label.clickable-label-hovered {
+  cursor: pointer;
+  --pf-c-label__content--before--BorderWidth: 2.5px;
+  --pf-c-label__content--before--BorderColor: #8a8d90;
```
tthvo, line 179: Wondering which visual effect for this css rule?

tthvo: On this note, maybe we can just remove it as their natural border color with with 2.5px looks better than the grey border :D

src/app/Dashboard/AutomatedAnalysis/ClickableAutomatedAnalysisLabel.tsx
```diff
@@ -42,18 +42,19 @@ import { InfoCircleIcon } from '@patternfly/react-icons';
...
-  isSelected: boolean;
 }
 
 export const clickableAutomatedAnalysisKey = 'clickable-automated-analysis-label';
 
 export const ClickableAutomatedAnalysisLabel: React.FunctionComponent<ClickableAutomatedAnalysisLabelProps> = ({
   label,
-  isSelected,
 }) => {
+  const { t } = useTranslation();
```
tthvo, line 56: I think either way is fine but maybe we can have to consistent to either use:

```typescript
const { t } = useTranslation();
```

or 

```typescript
const [t] = useTranslation();
```

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisConfigDrawer.tsx
```diff
@@ -186,7 +189,7 @@ export const AutomatedAnalysisConfigDrawer: React.FC<AutomatedAnalysisConfigDraw
                 splitButtonVariant="action"
                 toggleVariant="default"
               >
-                Create Recording &nbsp;&nbsp;
+                {t('AutomatedAnalysisConfigDrawer.CREATE_RECORDING')} &nbsp;&nbsp;
```
tthvo, line 192: One idea here is that we can use margin with span so we can control how much space to add. Wdu think?

```typescript
<span style={{ marginRight: '0.5em'}}>
  {t('AutomatedAnalysisConfigDrawer.CREATE_RECORDING')}
</span>
```

src/app/Dashboard/AutomatedAnalysis/Filters/AutomatedAnalysisScoreFilter.tsx
```diff
@@ -111,39 +113,51 @@ export const AutomatedAnalysisScoreFilter: React.FC<AutomatedAnalysisScoreFilter
       return 'automated-analysis-score-filter-slider automated-analysis-score-filter-slider-ok';
     }
   }, [currentScore]);
+  //       // 'Severity scores are calculated based on the number of JFR events that were triggered by the application in the time the report was generated.'
```
tthvo, line 116: Remove this comment?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -86,16 +94,41 @@ import {
...
+  OuterScrollContainer,
+  TableComposable,
+  Tbody,
+  Td,
+  Th,
+  Thead,
+  ThProps,
+  Tr,
+} from '@patternfly/react-table';
+import { t } from 'i18next';
```
tthvo, line 130: Should import `t` from `useTranslation`?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisConfigForm.tsx
```diff
@@ -65,60 +61,43 @@ import {
...
 } from '@patternfly/react-core';
-import { CogIcon } from '@patternfly/react-icons';
 import * as React from 'react';
-import { Link } from 'react-router-dom';
+import { useTranslation } from 'react-i18next';
 import { concatMap, filter, first } from 'rxjs';
 interface AutomatedAnalysisConfigFormProps {
-  onCreate?: () => void;
-  onSave?: () => void;
   isSettingsForm: boolean;
```
tthvo, line 68: I think we can rename this to `noTitle` or something similar. Then, form embeded in the wizard can pass also pass `false`, which remove the title. How about that?

**Before**

Screenshot from 2023-02-15 04-54-37

**After**
Screenshot from 2023-02-15 04-52-37

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisConfigForm.tsx
```diff
@@ -65,60 +61,43 @@ import {
...
-  onCreate,
-  onSave,
-  ...props
-}) => {
+export const AutomatedAnalysisConfigForm: React.FC<AutomatedAnalysisConfigFormProps> = ({ ...props }) => {
   const context = React.useContext(ServiceContext);
   const addSubscription = useSubscriptions();
+  const [t] = useTranslation();
 
+  const parseEventString = (): [string | undefined, TemplateType | undefined] => {
```
tthvo, line 76: Memo this?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisConfigForm.tsx
```diff
@@ -65,60 +61,43 @@ import {
...
+    return [templateName, templateType];
+  };
+  const [recordingConfig, setRecordingConfig] = React.useState<AutomatedAnalysisRecordingConfig>(
+    context.settings.automatedAnalysisRecordingConfig()
+  );
   const [templates, setTemplates] = React.useState([] as EventTemplate[]);
-  const [templateName, setTemplateName] = React.useState<string | undefined>(undefined);
-  const [templateType, setTemplateType] = React.useState<TemplateType | undefined>(undefined);
-  const [maxAge, setMaxAge] = React.useState(context.settings.automatedAnalysisRecordingConfig().maxAge);
+  const [templateName, setTemplateName] = React.useState<string | undefined>(parseEventString()[0]);
```
tthvo, line 93: Since we already have the `recordingConfig` which holds all the config. We can remove these states (i.e. `templateName`, `maxAge`, `maxSize`? We will still need to keep the unit tho.

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisConfigForm.tsx
```diff
@@ -65,60 +61,43 @@ import {
...
+  const [templateType, setTemplateType] = React.useState<TemplateType | undefined>(parseEventString()[1]);
+  const [maxAge, setMaxAge] = React.useState(recordingConfig.maxAge);
   const [maxAgeUnits, setMaxAgeUnits] = React.useState(1);
-  const [maxSize, setMaxSize] = React.useState(context.settings.automatedAnalysisRecordingConfig().maxSize);
+  const [maxSize, setMaxSize] = React.useState(recordingConfig.maxSize);
   const [maxSizeUnits, setMaxSizeUnits] = React.useState(1);
-  const [isLoading, setIsLoading] = React.useState(false);
```
tthvo, line 93: Maybe we can still keep this to show when the templates are being fetched?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -636,28 +799,37 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
...
                 onChange={handleNAScoreChange}
                 id="show-na-scores"
                 name="show-na-scores"
               />
             </ToolbarItem>
+            <ToolbarItem>
+              <Switch
+                label={t('AutomatedAnalysisCard.SWITCH_LABEL_LIST_VIEW')}
+                isChecked={showListView}
+                onChange={() => setShowListView(!showListView)}
```
tthvo, line 821: Access to current state should use factory function syntax? `setState((state) => !state)`;

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
tthvo: Ahh this is for the exported const. We probably should start exporting only translation keys and consumers somewhere can call `t`.

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -605,11 +651,128 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
...
+          <EmptyStateBody>{t('AutomatedAnalysisCard.NO_RESULTS_BODY')}</EmptyStateBody>
+          <EmptyStateSecondaryActions>
+            <Button variant="link" onClick={handleClearFilters}>
+              {t('AutomatedAnalysisCard.CLEAR_FILTERS')}
+            </Button>
+          </EmptyStateSecondaryActions>
+        </EmptyState>
+      );
+    }
+    if (showListView) {
```
tthvo, line 672: Extract to a component?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -580,6 +595,41 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
...
+    ) : (
+      <span className={css('pf-m-red', 'pf-c-label__icon')}>
+        <ExclamationCircleIcon />
+      </span>
+    );
+  }, []);
+
+  const getSortParams = React.useCallback(
+    (columnIndex: number): ThProps['sort'] => ({
+      sortBy: {
```
tthvo, line 620: I think this type is `ISortBy` from `@patternfly/react-table`. 

So, we can just keep a state like below instead of 2 separate state for each field (i.e. activeSortIndex and activeSortDirection`.

https://github.com/cryostatio/cryostat-web/blob/23c440b242c30e4d23a40d960374e062e0eae72f/src/app/Agent/AgentProbeTemplates.tsx#L100

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -591,11 +641,7 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
             reportStalenessTimer > 1 ? 's' : ''
           } ago.`}
           &nbsp;
```
tthvo, line 643: Wrap in a span and add margin right?

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -605,11 +651,128 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
...
+    if (filtered.length === 0) {
+      return (
+        <EmptyState>
+          <EmptyStateIcon icon={SearchIcon} />
+          <Title headingLevel="h4" size="lg">
+            {t(`AutomatedAnalysisCard.NO_RESULTS`)}
+          </Title>
+          <EmptyStateBody>{t('AutomatedAnalysisCard.NO_RESULTS_BODY')}</EmptyStateBody>
+          <EmptyStateSecondaryActions>
+            <Button variant="link" onClick={handleClearFilters}>
```
tthvo, line 665: How about the case where `Show N/A` is not checked and filtered results have N/A scores? Maybe add another clause for it?

Screenshot from 2023-02-15 06-04-18

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
```diff
@@ -788,6 +972,69 @@ export const AutomatedAnalysisCard: React.FC<AutomatedAnalysisCardProps> = (prop
...
+
+export const AutomatedAnalysisHeaderLabel: React.FC<AutomatedAnalysisHeaderLabelProps> = (props) => {
+  const dispatch = useDispatch();
+
+  const [isHoveredOrFocused, setIsHoveredOrFocused] = React.useState(false);
+
+  const handleHoveredOrFocused = React.useCallback(() => setIsHoveredOrFocused(true), [setIsHoveredOrFocused]);
+  const handleNonHoveredOrFocused = React.useCallback(() => setIsHoveredOrFocused(false), [setIsHoveredOrFocused]);
+
+  let onClick: () => void;
```
tthvo, line 990: This block can be memo with deps on `props.type`?

locales/en/common.json
```diff
@@ -1,21 +1,44 @@
...
+  "SECOND_other": "Seconds",
   "SELECT": "Select",
   "SHOW_LESS": "Show less",
   "SHOW_MORE": "Show more",
   "TEMPLATE": "Template",
-  "TIME": "Time"
+  "TEMPLATE_HELPER_TEXT_INVALID": "Template must be selected",
+  "TIME": "Time",
+  "WARNING": "WARNING"
 }
```
tthvo, line 44: Should we also localize `N/A`?

src/test/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.test.tsx
```diff
@@ -215,15 +215,15 @@ describe('<AutomatedAnalysisCard />', () => {
...
+    renderWithServiceContextAndReduxStore(<AutomatedAnalysisCard dashboardId={0} />, {
       preloadState: preloadedState,
     });
 
     expect(screen.getByText('Automated Analysis Error')).toBeInTheDocument(); // Error view
     expect(screen.getByText(FAILED_REPORT_MESSAGE)).toBeInTheDocument(); // Error message
     expect(screen.getByText('Cryostat was unable to generate an automated analysis report.')).toBeInTheDocument(); // Error details
     expect(screen.getByRole('button', { name: /retry loading report/i })).toBeInTheDocument(); // Retry button
-    expect(screen.queryByLabelText('automated-analysis-toolbar')).not.toBeInTheDocument(); // Toolbar
+    expect(screen.queryByLabelText('Automated Analysis Toolbar')).not.toBeInTheDocument(); // Toolbar
```
tthvo, line 226: Should use `testTranslate` to pull translated strings? Same for other tests.

src/test/Settings/AutomatedAnalysisConfig.test.tsx
```diff
@@ -72,11 +72,8 @@ describe('<AutomatedAnalysisConfig/>', () => {
...
     renderWithServiceContext(React.createElement(AutomatedAnalysisConfig.content, null));
-
-    Object.values(defaultAutomatedAnalysisRecordingConfig).forEach((v) => {
-      const currentConfig = screen.getByText(v);
-      expect(currentConfig).toBeInTheDocument();
-      expect(currentConfig).toBeVisible();
-    });
+    const template = screen.getByText('Current configuration');
+    expect(template).toBeInTheDocument();
+    expect(template).toBeVisible();
```
tthvo, line 77: Need to check current configuration displays too?

tthvo (commented):
Looks awesomee! Just some questions/comments. 

One thing I notice is that the AA form now removes the functionality to create the recording. There could be 2 issues:

- After creating an AA card, no analysis is found even though the user steps via the wizard and approve the current config (clicking Finish).
- If the user delete the current AA content, they can choose to set another config by clicking the dropdown and select `Custom`. However, now there is no button to  create.

locales/en/public.json
maxcao13: Yes, when the Automated Analysis shows no warnings or critical results. https://github.com/cryostatio/cryostat-web/pull/862/files#diff-b48b9c2299bbdbf197673ec97eae7e57c2e00305d1e135c0da01777c150ec300R1015

maxcao13: Oh if you are talking about the RESULT, then its here https://github.com/cryostatio/cryostat-web/pull/862/files#diff-b48b9c2299bbdbf197673ec97eae7e57c2e00305d1e135c0da01777c150ec300R718

Now that I look at it, I should rename it to NAME  and put it in public.js.

src/app/Dashboard/AutomatedAnalysis/ClickableAutomatedAnalysisLabel.tsx
maxcao13: Yeah I have been flipping between the two. I think I prefer the first over the latter, but what do you think?

tthvo: Same here! I think we can go with the first one then :D

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.tsx
maxcao13: Are you saying I should change the wording of the text?

src/test/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.test.tsx
maxcao13: Not sure, the tests work fine without it, probably because the component gets rendered with i18n correctly before being tested? What should the right behaviour be here?

maxcao13:
> Looks awesomee! Just some questions/comments.
> 
> One thing I notice is that the AA form now removes the functionality to create the recording. There could be 2 issues:
> 
>     * After creating an AA card, no analysis is found even though the user steps via the wizard and approve the current config (clicking Finish).
> 
>     * If the user delete the current AA content, they can choose to set another config by clicking the dropdown and select `Custom`. However, now there is no button to  create.

For the first point, I think I would have to hook into the AddCard finish button to create a recording or some other way.

src/test/Dashboard/AutomatedAnalysis/AutomatedAnalysisCard.test.tsx
tthvo: I think because `Common.tsx` already intialize the i18n module by side-effect importing it. This mock i18n module is always set to english so it passed. I think we might need to more explicit by calling `testTranslate`?

locales/en/public_old.json
```diff
@@ -0,0 +1 @@
+{}
```
tthvo, line 1: Remove this?

src/app/Dashboard/AddCard.tsx
```diff
@@ -194,7 +196,7 @@ export const AddCard: React.FC<AddCardProps> = (_) => {
                   </Select>
                   <Text>
                     {selection
-                      ? getConfigByTitle(selection).descriptionFull
+                      ? t(getConfigByTitle(selection, t).descriptionFull as string)
```
tthvo, line 199: Any issue that we need to type assertion here?

tthvo: Ahh maybe we should check if description is string => its a key and pass to `t`. Or else just display it.

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisConfigDrawer.tsx
```diff
@@ -125,10 +116,9 @@ export const AutomatedAnalysisConfigDrawer: React.FC<AutomatedAnalysisConfigDraw
     drawerRef.current && drawerRef.current.focus();
   }, [drawerRef]);
 
-  const onOptionSelect = React.useCallback(() => {
-    setIsDropdownOpen(false);
+  const onCogSelect = React.useCallback(() => {
     setIsExpanded(!isExpanded);
```
tthvo, line 120: Should use factory function notation :D Just a React practice thing.

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisConfigForm.tsx
```diff
@@ -324,125 +271,83 @@ export const AutomatedAnalysisConfigForm: React.FC<AutomatedAnalysisConfigFormPr
...
+  if (isLoading) {
+    return <LoadingView />;
+  }
   return (
     <>
-      {props.isSettingsForm ? (
-        formContent
-      ) : (
+      {useTitle ? (
         <Form>
```
tthvo, line 346: I think we will still need to `Form` to be common parent here. Just remove the FormSection if `useTitle` is false.

andrewazores:
> Or it could just be like this where the user has to explicitly start the recording after adding the card. Thoughts?

I think this is best, at least for now. The user might be trying to set up a dashboard but not actually be interested in seeing the dashboard data for the specific target they happen to currently have selected. Since recordings do have some amount of runtime overhead on the target, it should probably be something that the user explicitly performs.

andrewazores:
I say "at least for now" because we should probably actually handle the `NO_TARGET` selection case nicely and allow the dashboard to be created and a skeleton or empty state version of it rendered when the user hasn't selected any target. But even after we have that I think that starting a recording should still be an explicit opt-in action.

src/app/Dashboard/AutomatedAnalysis/AutomatedAnalysisConfigDrawer.tsx
maxcao13: I don't think I can since this is `onClick` on a button where the first parameter is the event.

maxcao13:
There is one final issue where menus and dropdowns stay within the card container such that if you resize the container small enough, the dropdown will stick within the container and not pop out like it should. I will file this as a separate issue since it applies to multiple issues. e.g. the action kebab of the cards. 

One caveat is that we cannot just append the element higher than the container in the DOM because then clicking on the menu and then dragging the card will result in the card sticking in place. So there is more thought and handling required.

Filter menu does not appear on top:
image

What it should look like:
image

Kebab menu does not appear on top:
image

tthvo (approved):
Looks awesomee to me! Good workkk!

Issue #863, opened 2023-02-07
[Bug] Rendering odd whitespace at end of page


andrewazores:
![image](https://user-images.githubusercontent.com/3787464/217264290-303204e1-f99e-47d9-949d-24addf4eb879.png)

Notice the whitish band of space at the bottom of the image. I notice this when using my workstation with a 1920x1200 display.

It seems to be caused by https://github.com/cryostatio/cryostat-web/pull/858 . If I check out a commit before that, or locally revert that patch, then the same page looks like this (normal):

![image](https://user-images.githubusercontent.com/3787464/217264883-f05cbf6c-0c0c-468e-9a4a-5a2935c47149.png)

tthvo:
@andrewazores I also see this when using PF topology extension. My issue was with the `skip-to-content` component in `AppLayout` and removing it fixed the issue. Is this the same?

tthvo:
I guess we can remove that component for now (might lose some functionality for keyboard users and accessibility support) and keep the react-table version bump?

andrewazores:
Needs more investigation, I think. If the `react-table` version bump isn't bringing us anything on its own then I would prefer to keep the SkipToContent component. But, it's not necessarily clear what exactly was changed in that `react-table` update. The [release notes](https://github.com/patternfly/patternfly-react/releases?q=react-table&expanded=true) might be helpful. Otherwise, maybe we should do a bisect of that `[6, 39]` patch version range and see which particular version introduced the problem. If we can narrow that down we can try to prepare a separate small reproducer for the observed problem. If we can get it to happen again then we file a bug with PF, otherwise there is some other problem in our setup that should be fixable.

tthvo:
https://github.com/patternfly/patternfly-react/pull/8521. seems to be pulling in `4.112.31`?

tthvo:
I think this is the issue:

This line specifies that `SkipToContent` applies styles from button, button-primary, and skipToContent.

https://github.com/patternfly/patternfly-react/blob/25b70978ef1d779ff98260036ce83b94bff5c34e/packages/react-core/src/components/SkipToContent/SkipToContent.tsx#L37

But when rendering, notice how the `position: absolute` from skipToContent style is being override by `position: relative` from button.

![Screenshot from 2023-02-07 13-45-18](https://user-images.githubusercontent.com/68053619/217337461-1db1c766-d256-4bc1-9ecd-5b4845d082c3.png)

tthvo:
[...]

PR #864, opened 2023-02-07
build(deps): revert @patternfly/react-table from 4.112.39 to 4.112.6

andrewazores:
This reverts commit 943460c53b4c5865c34abfa65c3b8528c03e0dd5.

Signed-off-by: Andrew Azores <aazores@redhat.com>

# Welcome to Cryostat! 👋

Related to #863

## Description of the change:
This reverts #858 , which introduced some styling change that causes views containing tables to render strangely on 1920x1200 displays (or anything not 16:9 perhaps).

## Motivation for the change:
This is a short-term fix for the bug referenced. Ideally we can upgrade to that `@patternfly/react-table` version or a later one. There may be some css fix or similar that we need to apply locally on top of the version upgrade.

## How to manually test:
Simply resizing the viewport doesn't seem to trigger the bug, so I'm not sure how to test this without having access to a non-16:9 display. With such a display, just open the web-client and go to ex. `Recordings` and check if the negative space at the bottom of the page is present.


maxcao13 (approved):
Something weird going on. I have a problem where the yarn tests fail because my PR revert the snapshot change to `"OUIA-Generated-Button-plain-5"`, just like this revert does. I will approve, it may solve my issue as well.

tthvo:
maybe we should hold off to see if we can resolve this https://github.com/cryostatio/cryostat-web/issues/863#issuecomment-1421571328. The Rule table needs this version in order for the ActionColumns to have menuAppendTo added...

tthvo:
@andrewazores Hey, looks like this react-table `4.112.39` is requiring core `^4.276.6`.

https://github.com/patternfly/patternfly-react/blob/e2d787cd453afc07405d7f5f10eb6b64a215b064/packages/react-table/package.json#L42

While quickstarts only uses [`^4.267.6`](https://github.com/patternfly/patternfly-quickstarts/blob/ef2662f35fac6ed6f0a10eea68ef32d58bef0b65/packages/module/package.json#L58). This causes https://github.com/cryostatio/cryostat-web/pull/881#pullrequestreview-1307964458 to fail.

Not sure how long it takes for quickstarts to bump core so we really need to revert this?

tthvo:
> This causes https://github.com/cryostatio/cryostat-web/pull/881#pullrequestreview-1307964458 to fail.

On some closer look, I am not even sure why the `core` is automatically pumped to `2.276.6` (checked with `yarn info`). I just suspect this react-table cuz it seems to be the only place that requires this core version.

andrewazores:
We still have some time to make the call of whether to revert this or not, so let's wait a while longer and see what happens with those different Patternfly packages and see if their interdependencies sort themselves out. Or we can just ask the Patternfly team what's going on and what the release cadence is for things.

tthvo:
Ahh I am seeing some peer deps are being pumped to later versions what requires `4.276.6`. On a second thought, we might need to bump our PF core (https://github.com/cryostatio/cryostat-web/pull/851) and keep this PR.

The only problem is that quickstart package is not yet bumping their core. I will open an issue on their repo to see.

andrewazores:
Awesome, thanks for following up on that.

tthvo:
Maybe we can close this since the css order is now fixed?

Issue #840, opened 2023-01-31
[Bug] Dashbard should hide cards with unmatching feature level


tthvo:
## Current Behavior:

Noticed this today that cards belong to lower `FeatureLevel` still persist if switching from lower to higher feature level.

## Expected Behavior:

Dashboard only shows cards matching current feature level.

## Screenshots

Notice how console is in `PRODUCTION` mode but a None card is still shown.

![Screenshot from 2023-01-31 12-43-05](https://user-images.githubusercontent.com/68053619/215840800-6de52ded-ffc3-4586-a0c6-42799eb91946.png)

## Steps To Reproduce:
1. Go into DEVELOPMENT mode
2. Create any dev card
3. Switch back to PRODUCTION/BETA
4. Those cards still persist in dashboard.

tthvo:
Fixes seem to complicate reordering and resizing. Any thoughts? I guess it only affect releases that has BETA cards.

andrewazores:
I don't think we would want to make any permanent modifications to the user's dashboard config when they change feature levels. Since the feature level is included in the card configuration properties, maybe the whole card component can just be wrapped in a feature flag at that level. This way it can still be included in the persisted serialized layout, but it will only be rendered if the correct feature level is selected. Or even simpler, maybe there should just be a feature level badge on the card headers similar to what is displayed on the application masthead bar. This way it's explicit to the user if the cards they are using are part of the beta feature flag level and they can decide on their own what to do with that.

maxcao13:
I think this issue has been resolved, thoughts?

Issue #739, opened 2022-12-09
[Task] Allow enlarged view of individual Dashboard cards


tthvo:
## Describe the feature

> I am thinking if we can allow users to see dashboard cards individually in 'full-screen' mode similar to Grafana. Then, the user can hit `Go back` (i.e. breadcrumbs) to go back main dashboard. I think we can implement this in the card action menu.

![Screenshot from 2022-12-08 18-13-35](https://user-images.githubusercontent.com/68053619/206586626-6c2324ae-89e7-463c-95a3-9549a2438f0c.png)

## Any other information?

_Originally posted by @tthvo in https://github.com/cryostatio/cryostat-web/issues/731#issuecomment-1343495401_

PR #870, opened 2023-02-08
feat(dashboard): wrap cards in feature flag

andrewazores:
# Welcome to Cryostat! 👋

Related to #840

## Description of the change:
Wrap dashboard cards in a `<FeatureFlag/>` so that they are only rendered if their configuration defines them part of the current feature level or higher. The dashboard layout is not actually modified in storage, the elements are only skipped when the component tree is being built, so if the user returns back to a lower feature level then the cards will reappear.

TODO later: add some visual indicator, like a badge in the card header matching the one in the application masthead, or a colour-coded border matching that badge, on each card so that it is obvious to the user that the cards they are using may not be production features. In this PR the user can only discover this by toggling the feature level and checking what disappears. This step entails more refactoring to extract the card itself to the higher component level and leave only the inner content as part of the card component, which is also useful for #739.

### DIFF
--- a/src/app/Dashboard/AddCard.tsx
+++ b/src/app/Dashboard/AddCard.tsx
@@ -73,7 +73,7 @@ import * as React from 'react';
 import { useContext } from 'react';
 import { useDispatch } from 'react-redux';
 import { Observable, of } from 'rxjs';
-import { getDashboardCards, getConfigByTitle, PropControl } from './Dashboard';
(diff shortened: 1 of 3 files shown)


tthvo (approved):
Looks good to me! Worked well :D

maxcao13 (approved):
Agreed, works well!

PR #875, opened 2023-02-11
feat(dashboard): MBean numeric metrics card

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #874
Depends on https://github.com/cryostatio/cryostat/pull/1361

## Description of the change:
Adds a dashboard card type which displays charts of JMX numeric metrics.

## Motivation for the change:
These are a lighter-weight alternative to JFR metrics but which offer much less fine-grained information. For a high-level
monitoring overview user story, if the data required is available via this mechanism then it should be preferred to be
gathered this way rather than by collecting JFR files.

## How to manually test:
2. *...*

image


andrewazores:
I'm intentionally leaving out the non-numeric, or even generally non-timeseries, MBean data from this PR. Data that are strings or numbers that don't typically change for the lifetime of the target application instance should be represented/visualized differently and probably won't have much implementation in common with these charts.

src/app/Dashboard/Charts/JFRMetricsChartController.tsx
```diff
@@ -126,7 +126,7 @@ export class ChartController {
     this._stop();
     this._attach = merge(
       merge(
-        this._updates$.pipe(throttleTime(this._settings.chartControllerConfig().minRefresh)),
+        this._updates$.pipe(throttleTime(this._settings.chartControllerConfig().minRefresh * 1000)),
```
tthvo, line 129: This `minRefresh` is already in seconds?

src/app/Dashboard/Charts/MBeanMetricsChartController.tsx
```diff
@@ -0,0 +1,220 @@
...
+    );
+  }
+
+  requestRefresh(): void {
+    this._loading$.next(true);
+    this._updates$.next();
+  }
+
+  loading(): Observable<boolean> {
+    return this._loading$.asObservable();
```
tthvo, line 121: We could use `distinctUntilChanged` here?

src/app/Dashboard/Charts/MBeanMetricsChartController.tsx
```diff
@@ -0,0 +1,220 @@
...
+            updated.splice(idx, 1);
+          }
+        });
+        this._metrics.get(category)?.push(...updated);
+      })
+    );
+  }
+
+  requestRefresh(): void {
+    this._loading$.next(true);
```
tthvo, line 116: Seems like in the `_attach` obs chain we have a tap to set this already. We can remove it here?

src/app/Dashboard/Charts/MBeanMetricsChartController.tsx
```diff
@@ -0,0 +1,220 @@
...
+              mbeanMetrics {
+                ${q.join('\n')}
+              }
+            }
+          }`,
+        { connectUrl: target.connectUrl }
+      )
+      .pipe(
+        map((resp) => {
+          const nodes = resp?.data?.targetNodes;
```
tthvo, line 176: We can remove undefined checking here based on `MBeanMetricsResponse`?

src/app/Dashboard/Charts/MBeanMetricsChartController.tsx
```diff
@@ -0,0 +1,220 @@
...
+        concatMap((t) => this._queryMetrics(t)),
+        tap((_) => this._loading$.next(false))
+      )
+      .subscribe((v) => this._state$.next(v));
+  }
+
+  private _queryMetrics(target: Target): Observable<MBeanMetrics> {
+    const q: string[] = [];
+    this._metrics.forEach((fields, category) => {
+      // TODO deduplicate this constructed query
```
tthvo, line 157: Still a TODO?

src/app/Dashboard/Charts/JFRMetricsChartController.tsx
andrewazores: Right, the stored value is in seconds, but the `throttleTime` function takes milliseconds.

tthvo: ohhh didnt notice that! Thanks :D

maxcao13:
image

Is there a way to fix the cards so that the charts dont have these cutoff and overlapping labels? Or is this just a limitation of the patternfly charts framework?

Also it may be useful to always have unit labels? e.g. the "average" cards maybe should have **%** label and the threads card, just to make it consistent, maybe should have a **Threads** label

andrewazores:
I can add the `"threads" unit to that chart, sure. By the "average" charts, are you referring to the CPU/System Load (Average) charts?

https://docs.oracle.com/en/java/javase/17/docs/api/jdk.management/com/sun/management/OperatingSystemMXBean.html#getSystemCpuLoad()

I can add the '%' as a unit for the process and system CPU load metrics, but the Load Average is actually not a percentage value. I don't think I've ever seen a load average expressed with a unit either, though I suppose one could be defined given how it's calculated, but would a user be familiar with that unit definition?

maxcao13:
> I can add the '%' as a unit for the process and system CPU load metrics, but the Load Average is actually not a percentage value. I don't think I've ever seen a load average expressed with a unit either, though I suppose one could be defined given how it's calculated, but would a user be familiar with that unit definition?

Oh I see, these values are defined like that... hmm then I guess I'm not sure because since I was not familiar with those conventions I instantly assumed those `0.002` or `0.01` values were being expressed as percentages written in decimal form (0.2% or 1%). But I assume a Cryostat user may be familliar with the Process and system CPU values as being between between `[0.0,1.0]`. As for load average, I leave it to you since the docs don't really have their own definition of [SystemLoadAverage](https://docs.oracle.com/en/java/javase/17/docs/api/java.management/java/lang/management/OperatingSystemMXBean.html#getSystemLoadAverage()) units other than just being defined the "system load average".

andrewazores:
The [0.0, 1.0] ones still make sense to express as a percentage. Or maybe I just should multiply the values by 100 so that they actually are percentages, that's fine.

I think load average is a pretty well-understood metric even if it doesn't normally come with units, though.

andrewazores:
Still trying to figure out how to get the charts to render some internal components more nicely:

image

The chart itself is actually all SVG, so getting it to render differently inside of its container is a bit odd. I'd like to not have the bottom legend clipped off like that and it would also be nice to get it to render the Y axis further away from the container edge if the tick labels are either long or are overlapping the Y axis unit label.

andrewazores:
Simply using the `padding` prop on the chart produces really weird looking charts:

image

maxcao13:
Seems like if I set a set a short refresh rate, the chart does not actually refresh that fast. 

image

It has datapoints every 10? seconds.

andrewazores:
Go to settings > dashboard and decrease your minimum refresh interval :-) it defaults to 10 seconds, though these charts try to default to 5. I'll bump that up to also default to 10, but maybe it should dynamically default to whatever is defined by that setting.

maxcao13:
image

Is it possible to round these numbers to 1 or 2 decimal places? They seem a bit long.

tthvo:
> Looks like the tests are failing because of something to do with mock timers/dates or localization with the snapshots. @tthvo what do I need to do here?

Seems like some internal module variables are initialized before mock statement is met. I think this should fix the issue (just mock before importing).

```diff
 
+jest.useFakeTimers('modern').setSystemTime(new Date('14 Feb 2023 00:00:00 UTC'));
 jest.mock('@app/Dashboard/Charts/MBeanMetricsChartController');
 
 import { ChartContext } from '@app/Dashboard/Charts/ChartContext';
@@ -77,10 +78,6 @@ const mockChartContext = {
 const MB = Math.pow(1024, 2);
 
 describe('<MBeanMetricsChartCard />', () => {
-  beforeEach(() => {
-    jest.useFakeTimers('modern').setSystemTime(new Date('14 Feb 2023 00:00:00 UTC'));
-  });
-
   afterEach(() => {
     cleanup();
     jest.useRealTimers();

```

andrewazores:
Hmm... looks like the snapshots are still failing the same way.

maxcao13:
> image
> 
> Is it possible to round these numbers to 1 or 2 decimal places? They seem a bit long.

Also using this screenshot, it seems the dates and times are not synced up with the date/time localization setting. Changing my date locale and/or timezone does not affect the chart.

tthvo:
> Hmm... looks like the snapshots are still failing the same way.

Wow...thats odd. It's passing for me locally...

andrewazores:
Oh, yeah, it passes for me locally too, it's just in the CI environment that it fails like this.

tthvo:
I am seeing these warnings

```
› 9 snapshots obsolete from 1 test suite. To remove them all, re-run jest with `-u`.
   ↳ src/test/Dashboard/Charts/MBeanMetricsChartCard.test.tsx
       • <MBeanMetricsChartCard /> renders correctly: with-content 1
       • <MBeanMetricsChartCard /> renders correctly: with-content 2
       • <MBeanMetricsChartCard /> renders correctly: with-content 3
       • <MBeanMetricsChartCard /> renders correctly: with-content 4
       • <MBeanMetricsChartCard /> renders correctly: with-content 5
       • <MBeanMetricsChartCard /> renders correctly: with-content 6
       • <MBeanMetricsChartCard /> renders correctly: with-content 7
       • <MBeanMetricsChartCard /> renders correctly: with-content 8
       • <MBeanMetricsChartCard /> renders loading state correctly: loading-view 1
```

Seems to remove the snapshots if running with `-u`: `› 1 snapshot file removed from 1 test suite.`

I am wondering if the plots are using any setTimeout or anything that requires time to advance forward?

src/app/Dashboard/Charts/MBeanMetricsChartController.tsx
```diff
@@ -0,0 +1,225 @@
...
+            return {};
+          }
+          return nodes[0]?.mbeanMetrics;
+        }),
+        catchError((_) => of({}))
+      );
+  }
+}
+
+export interface MemoryMetric {
```
maxcao13, line 192: Maybe should be called `MemoryUsage` or `MemoryUsageMetric` to match the java object?

tthvo:
> I am wondering if the plots are using any setTimeout or anything that requires time to advance forward? Because with current mock, time is fixed and won't move forward so all those setTimeout(), setInterval(), clearTimeout(), clearInterval() will all be stuck.

Does not seem to be the issue (i guess charts have their datapoint already). I waited some time later. Seems like the diff in time is always `7:00PM` and `12:00AM`

maxcao13:
One more thing as far as features go, I forgot that #881 includes a "metric card" card config placeholder. I think this PR should include some relevant sample first run MBeanMetric dashboard cards. e.g.
image

All that needs to be done is just to add and replace the relevant card configs to this `CardConfig[]`: https://github.com/cryostatio/cryostat-web/pull/881/files#diff-6b748dd16365d4ced719885fe30ea9018f21a92010df3f6f8d00e0f6ecb09a27R149

maxcao13:
Other than that and the broken snapshot tests, I think it looks great!

src/app/Dashboard/Charts/MBeanMetricsChartCard.tsx
```diff
@@ -0,0 +1,506 @@
...
+      themeColor={themeColor}
+      width={width}
+      padding={{
+        left: 54,
+        right: 30,
+        top: 10,
+        bottom: 60,
+      }}
+    >
+      <ChartAxis tickFormat={(t) => dayjs(t).format('LTS')} fixLabelOverlap />
```
tthvo, line 140: I think this line is the issue. We should apply timezone from user preferences.

```diff
diff --git a/src/app/Dashboard/Charts/MBeanMetricsChartCard.tsx b/src/app/Dashboard/Charts/MBeanMetricsChartCard.tsx
index 4607c135..75ef18a1 100644
--- a/src/app/Dashboard/Charts/MBeanMetricsChartCard.tsx
+++ b/src/app/Dashboard/Charts/MBeanMetricsChartCard.tsx
@@ -95,7 +95,7 @@ const SimpleChart: React.FC<{
   units: string;
   interpolation?: 'linear' | 'step' | 'monotoneX';
 }> = ({ themeColor, style, width, samples, units, interpolation }) => {
-  const [dayjs] = useDayjs();
+  const [dayjs, datatimeFormat] = useDayjs();
 
   const data = React.useMemo(
     () => samples.map((v) => ({ x: v.timestamp, y: v.datapoint.value, name: v.datapoint.name })),
@@ -135,7 +135,7 @@ const SimpleChart: React.FC<{
         bottom: 60,
       }}
     >
-      <ChartAxis tickFormat={(t) => dayjs(t).format('LTS')} fixLabelOverlap />
+      <ChartAxis tickFormat={(t) => dayjs(t).tz(datatimeFormat.timeZone.full).format('LTS')} fixLabelOverlap />
       <ChartAxis
         tickFormat={(t) => (typeof t !== 'number' ? t : t.toPrecision(2))}
         dependentAxis

```

tthvo: For jest-setup, we have it user pref timezone  fixed as `UTC` here:

https://github.com/cryostatio/cryostat-web/blob/c6870e3f7fbc6d5646f2810c2752ce559e63692b/test-setup.js#L29

tthvo: Testing with fork seems to pass:

https://github.com/tthvo/cryostat-web/actions/runs/4248950345/jobs/7388646202

andrewazores:
Snapshots seem to be OK now, I think, but there's this:

```
 FAIL src/test/Dashboard/Charts/MBeanMetricsChartCard.test.tsx
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'component')

      148 |   },
      149 |   {
    > 150 |     id: `${MBeanMetricsChartCardDescriptor.component.name}-1`,
          |                                            ^
      151 |     name: MBeanMetricsChartCardDescriptor.component.name,
      152 |     span: MBeanMetricsChartCardDescriptor.cardSizes.span.default,
      153 |     props: {

      at Object.<anonymous> (src/app/Shared/Redux/Configurations/DashboardConfigSlicer.tsx:150:44)
      at Object.<anonymous> (src/app/Dashboard/DraggableRef.tsx:38:1)
      at Object.<anonymous> (src/app/Dashboard/DashboardCard.tsx:43:1)
      at Object.<anonymous> (src/app/Dashboard/Charts/MBeanMetricsChartCard.tsx:60:1)
      at Object.<anonymous> (src/test/Dashboard/Charts/MBeanMetricsChartCard.test.tsx:44:1)
```

Pretty odd. I'm guessing some kind of circular dependency import issue again...

andrewazores:
https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de

This feels clumsy but maybe this or something like it is worth thinking about as the web-client gets larger and more complex.

maxcao13 (approved):
Everything looks good to me now, great job! I'm wondering about the non-moving data metrics. Something for a future issue, but should those cards be individual or have all the info on one card?

andrewazores:
At least some of that mbean data will be included in #895. The rest can either go into some further details on that same card, ex. as an expandable content or other similar feature, or can go onto a separate card of their own as well.

Issue #678, opened 2022-11-18
[Task] Target Topology View


tthvo:
We could have a separate side-card tab for target view (Topology) as here but with some tweaks? Other pages can just use context selector similar to how we select namespace in OpenShift.

![Screenshot from 2022-11-17 13-44-39](https://user-images.githubusercontent.com/68053619/202533195-0df30e31-d703-4af0-ba51-0b6b023777ed.png)

This way we can have some initial setup for multi-namespace support as we discussed. Also, filters can be supported here too. I guess this would mean all operations on target list (i.e. add, delete, edit) must be done only on this new side-car tab.

tthvo:
For our case, the `/api/v2.2/discovery` endpoint will return a discovery tree and group the targets within correct REALM (i.e. OpenShift, JDP, Custom and Agent push). This means the front-end would just need to render this tree hierarchically and have a selector to filter REALM. Everything else under REALM node is just rendered as it is.

See an example below for the visualization of the returned tree:

![discovery](https://user-images.githubusercontent.com/68053619/203140411-954a3a0e-baee-49dd-9b0c-38c0541bd2e4.svg)

tthvo:
We can implement actions on target (i.e. target creation form using `Drawer`) as this design:

![detail-drawer](https://user-images.githubusercontent.com/68053619/203425499-c99e1331-6707-4ea8-a137-e73dd47bde0e.png)

This way, we can see and edit details directly.

tthvo:
Current issue with TS type checking: https://github.com/patternfly/patternfly-react/issues/8498 after bumping topology package version to `4.91.*`.

Current solution: setting `"strictFunctionTypes": false,` tho strict mode is still enabled with `"strict": true` for `tsconfig.json`.

Issue #752, opened 2022-12-13
[Request] Use Description List for target details


tthvo:
## Describe the feature

We can extract field values from target JSON into a more user-friendly panel like a description list. The JSON content can has its own section, for example,`raw JSON`.

![ocp-example](https://user-images.githubusercontent.com/68053619/207206826-eecddc3e-2ef8-4d49-a740-610cd258ddc0.png)

## Any other information?

Reference: https://www.patternfly.org/v4/components/description-list

Issue #666, opened 2022-11-17
[Task] Target selection should allow filtering


tthvo:
Right now, the target select menu will just show a plain list and will expand as needed. I think we should limit its size and allow filters. Or perhaps a new design for this.

Related to #429

andrewazores:
Sure, I like that idea. It's a good upgrade from where our UX is now at least. I think the question of how to display things when the user wants to select multiple targets, or a higher order node from the tree, is still open ("start a recording on every target below this Deployment"), but at least this way we get a way to visualize those higher order nodes and see some of the relationships between targets.

Issue #594, opened 2022-10-28
[Task] Add "test connection" button to Custom Target creation modal


andrewazores:

tthvo:
@andrewazores I am wondering why endpoint for creating custom target is not accepting labels? All custom targets now have empty labels...

andrewazores:
There are basically three kind of key-value stores attached to targets:

- labels
- annotations.platform
- annotations.cryostat

Labels are implicitly meant to come from the deployment platform, along with the platform annotations. ie. in a k8s environment with targets defined by the KubeApiPlatformClient, those two fields are copied from the k8s Endpoints/Pods objects etc.

Cryostat annotations are the ones that add extra information from the application, or in this case the client/user.

Now that there is the Discovery Plugin API it's possible for a client to register itself in a way that allows it to publish its own labels and platform annotations however it pleases too.

I think it makes sense for the labels and platform annotations to continue to be "immutable" (or at least defined by a plugin that probably keeps them stable), whereas the Cryostat annotations can change arbitrarily including by interactive user intervention.

tthvo:
Ahh right that makes sense! thanks for explaining :D

PR #891, opened 2023-02-21
feat(topology): topology view

tthvo:
# Welcome to Cryostat! 👋

Fixes: #678, #429, #867, #666, #752, #594 

## Description of the change:

This PR includes the new Topology to view discovered targets along with their group/realm. Remaining tasks:

- [ ] Filters (some difficulties to avoid re-initialize model and incorporate both target and group filters - worst case)
  - [ ] Target filters 
    - [ ] Name
    - [ ] Labels
    - [ ] Annotations
    - [ ] JVM ID
  - [ ] Group/Realm filter
    - [ ] Name
    - [ ] Labels
    - [ ] Annotations
    - [ ] Number of children (targets)
 - [ ] Duplicate/Reuse functionalities to listview
 - [ ] Implement a simple catalog view of possible entity that can added to topology view (shown when clicked on left-most icon in the toolbar). Currently, only custom target is allowed. Maybe later we can have others, for examples, notes.
 - [ ] Various chores to fix eslint errors and cleaner look.

**Note**: intentionally left out localization to keep the PR less clustered. Will include those changes as a follow-up PR.

## Motivation for the change:

See #678

## How to manually test:

- Go to Topology Tab under sidecar.
- **Note: Currently supporting local deployment with JDP, Agent and Custom Targets. Need to implement tree-walk function to transform data model.**


tthvo:
Just in draft right now as filtering and tree-walk function needs closer look. This works only in local dev (exploring how it is in OCP now). Let me know how it looks or any functionalities that you want to add in :D

andrewazores:
Wow, this looks amazing. My mind is pretty blown :-)

When I click on the link from the Topology view to add credentials for a target:

image

it would be cool if not only were I directed to the Security view but also if the credentials adding dialog could be drawn up immediately, too. Or maybe the modal can just be drawn up on top of the Topology view without directing to Security at all.

Also, maybe this is just the design guidelines for the PF Topology view, but I find that the connection test on the new custom target creation form is not very obvious. It took me a minute to realize that logo element was clickable. Maybe something like https://rxjs.dev/api/operators/throttleTime / https://rxjs.dev/api/operators/debounceTime on the text input fed into the testing function automatically would be helpful?

image

And if the test fails, the creation button should probably be disabled (until the connection URL input is changed) since we can now predict that the action is going to fail anyway.

tthvo:
> Also, maybe this is just the design guidelines for the PF Topology view, but I find that the connection test on the new custom target creation form is not very obvious. It took me a minute to realize that logo element was clickable. Maybe something like https://rxjs.dev/api/operators/throttleTime / https://rxjs.dev/api/operators/debounceTime on the text input fed into the testing function automatically would be helpful?

Ahhh actually I designed this layout to to replace the `Test` text button...Turned out not to be too obvious then. I will test out with the automatic testing here and maybe mention this in the top alert banner of the form + add a quickstart (if possible) + a blog post on this?

andrewazores:
Sounds good, though I don't think a blog post is necessary just for this aspect. A post covering the topology view as a whole would be great. Ideally the test button should look button-y and be discoverable/obvious to the user just looking at the view, without having to go look at some other documentation to figure out how to operate it.

tthvo:
Okayyy maybe a hint text right under the button would also help + mentioning in the top banner then :D

tthvo:
> it would be cool if not only were I directed to the Security view but also if the credentials adding dialog could be drawn up immediately, too. Or maybe the modal can just be drawn up on top of the Topology view without directing to Security at all

On this one, could u check the latest commit to see if that is what we want?

andrewazores:
> What if the latest input test is throttled (the user stops typing) and the validation state is now shown by the previous one?

The idea is just to prevent too many tests from being performed on the input while it's changing, so once the input has settled for some time there would be a "final" observable emission that would go through the pipeline and end up as a connection test call.

> What if one input test request is slow for some reason and comes after the newer test request. Then, its state will override the newer one? Probably can be solved in switchMap.

Possible yes, but as you say I think there's a way to solve this with the right kind of observable chaining, and by applying an appropriately slow/long debounce on the input changes.

> How is this automatic testing affects the backend load? Seems like we can DoS ourselves?

Well... I hope not. This isn't a very expensive request and if it's debounced so that each individual web-client instance is only performing one request every second or so, DoSing should be unlikely.

andrewazores:
> On this one, could u check the latest commit to see if that is what we want?

This is probably even better than what I was saying, I say stick with what you added here. The helper text for the test button is also probably enough, so no need to worry too much about the automatic testing thing.

tthvo:
@andrewazores Just wanna double check on this one. I added the resource tab for each target. Currently, it only shows the total of certain resource. Anything you think we should add in here? Each link should lead the users the corresponding view.

Screenshot from 2023-02-22 16-49-21

Screenshot from 2023-02-22 17-18-06

andrewazores:
Maybe the number of matching rules and credentials?

The links from that table to the views appear to work nicely, but navigating directly to those views from the main left side nav menu seems to be broken now.

tthvo:
Opps i forgot to check `null` XD

tthvo:
Rebasing causes this [CACHE_CHECKSUM_MISMATCH](https://yarnpkg.com/advanced/error-codes#yn0018---cache_checksum_mismatch):

```
@babel/template@npm:7.18.10: The remote archive doesn't match the expected checksum
```

Not sure what happens to this package checksum (maybe backports? updates?). So I have to clean cache and use `YARN_CHECKSUM_BEHAVIOUR=update` to fix (seems to fix and no visible problems at the moment)...

tthvo:
Testing with 2 sample apps on k8s, seems to work pretty fine. Visually, its a little hilarious but we can add in an option to show only top-level realm and omit an intermediate groups (similar to how we have with local dev).

Screenshot from 2023-02-28 13-42-08

maxcao13 (commented):
I think there is a bug where if you continue clicking on expand and close on the topology view node, it will continue to move farther and farther away: see

[Screencast from 02-28-2023 02:00:04 PM.webm](https://user-images.githubusercontent.com/69827283/221953265-e39da4bc-b02a-4e32-aa73-2755dc369180.webm)

andrewazores:
At least some of the intermediate node types are important to keep. In the k8s example where the realm is the k8s API built-in discovery plugin, the next node level is the Namespace which should definitely be visible. After that I think is the Deployment(Config) which should probably be visible since in practice there will often be many replicas belonging to the same Deployment and it only makes sense to keep that logical grouping visible.

Maybe a toggle option to collapse the model in cases where an intermediate node has only 1 child? If there is a chain like A -> B -> C -> D then it could be rendered only as A -> D. If there are chains like A -> B -> C2 -> D1, A -> B -> C -> D2, A -> B -> C2 -> D3, then only the B node would be collapsed.

I think putting the node type onto those labels might also be useful, but maybe too cluttered. ex. `Namespace: default`, `Deployment: quarkus-test`.

andrewazores:
Anyway, how does it look if you do scale up some of those deployments? With 2, 5, 10 replicas of each? If there are that many, is there any way to collapse them so that rather than 10 individual discs appearing, it can be a single icon that looks like a stack of discs and this can be expanded further if the user wants to drill down into it?

tthvo:
> Maybe a toggle option to collapse the model in cases where an intermediate node has only 1 child? If there is a chain like A -> B -> C -> D then it could be rendered only as A -> D. If there are chains like A -> B -> C2 -> D1, A -> B -> C -> D2, A -> B -> C2 -> D3, then only the B node would be collapsed.
> I think putting the node type onto those labels might also be useful, but maybe too cluttered. ex. Namespace: default, Deployment: quarkus-test.

Great idea! I will look into how to do this :D

> Anyway, how does it look if you do scale up some of those deployments? With 2, 5, 10 replicas of each? If there are that many, is there any way to collapse them so that rather than 10 individual discs appearing, it can be a single icon that looks like a stack of discs and this can be expanded further if the user wants to drill down into it?

I think only groups are allowed to be collapsible so I think we can collapse the deployment level?

> I think there is a bug where if you continue clicking on expand and close on the topology view node, it will continue to move farther and farther away: see

Seems like weird bugs. Event emitted from internal topology package always show changes in position. Maybe its has something to do with node size. I will look into it.

tthvo:
> Maybe a toggle option to collapse the model in cases where an intermediate node has only 1 child? If there is a chain like A -> B -> C -> D then it could be rendered only as A -> D. If there are chains like A -> B -> C2 -> D1, A -> B -> C -> D2, A -> B -> C2 -> D3, then only the B node would be collapsed.

not sure I understand right but this means we want to remove node B?

andrewazores:
Correct, B would be visually removed in that scenario.

I do think this should be a toggleable control that the user has to enable, since it is removing information from the graph that the user might need in order to make sense of what they are seeing. For large graphs where the user already understands what they are looking at then toggling this option on could help to remove some visual clutter and make things a little easier to navigate/locate.

andrewazores:
Actually, hold on, I didn't write the chains properly or look vlosely enough at your ASCII art representation of it...

In the case I wrote above, since B has two children, it would not be eliminated actually. Sorry. It is *C* that would be eliminated.

tthvo:
right thanks! I will work on that! Also, latest commit includes a `test.json` that contains sample k8s discovery when deploying on k8s/ocp. Just saving some times redeploying on a real cluster.

src/app/Topology/Actions/NodeActions.tsx
```diff
@@ -94,7 +94,8 @@ export interface NodeAction {
   action?: NodeActionFunction;
   title?: React.ReactNode;
   isSeparator?: boolean;
-  blackList?: NodeType[];
+  include?: NodeType[]; // Empty means all
+  blackList?: NodeType[]; // Empty means none
```
andrewazores, line 98: https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language

Let's call this something like 'blockList' or 'denyList'

tthvo: oh sorry! thanks for pointing that out...

tthvo:
With latest commit (latest image), I think this PR is stable enough to test on k8s/ocp for both graph view and listview. Tested with crc and it seems fine (will do further checks). But you can now test it out.

In display options, there are 2 mode to choose (Only realm and collapse singles - collapse single-child group).

andrewazores:
Awesome, I'll try it out soon.

andrewazores:
image

image

Could the target be pre-selected in the modal?

andrewazores:
image

After I define credentials and save, the modal closes and the topology node changes colour to indicate the new status. But the JVM ID/credentials warning is still there until I close and reopen that side panel.

andrewazores:
I also feel that the application nav menu should be reordered with Topology first, then the Overview section, and then the Console section.

andrewazores:
I think the "quickstarts blogs" top notification banner thing should be removed - it's a nice idea, but 1) the blogs shouldn't be the "official" source of docs, it should be the [Guides](https://cryostat.io/guides/) 2) the masthead `?` button has a link to the upstream website containing the docs already.

Maybe there can/should be a dashboard quickstart card on the default first run layout that points out things like this to the user? How to access the docs, how to file bugs, how to start discussions, how to change their locale settings. @maxcao13

andrewazores:
Overall this is (still) awesome, I'm very happy with the result and it's quite impressive. Really excited to have this in the next release after so long planning/dreaming about it.

For a follow-up, really not for this PR, it should also be possible to perform certain actions against the group nodes themselves, ie "start a recording on this Deployment". This should be a single GraphQL query:

`environment-nodes-query-1.graphql`
```graphql
query {
    environmentNodes(filter: { name: "JDP" }) {
        descendantTargets {
            doStartRecording(recording: {
                name: "foo",
                template: "Profiling",
                templateType: "TARGET",
                duration: 30
                }) {
                    name
                    state
            }
        }
    }
}
```

This query should be given a query name, and the name filter parameter should be passed as a separate GraphQL query parameter, but that's the gist of it.

image

+

image

tthvo:
> Could the target be pre-selected in the modal?

I was actually thinking of a way for this (same for automatic rules and settings) but I remember running into complications with knowing which value to listen to `context.target` from the early commits. I will look more closely again.

> After I define credentials and save, the modal closes and the topology node changes colour to indicate the new status. But the JVM ID/credentials warning is still there until I close and reopen that side panel.

Opps must be something is being memo with a missing hook deps.

> I think the "quickstarts blogs" top notification banner thing should be removed - it's a nice idea, but 1) the blogs shouldn't be the "official" source of docs, it should be the [Guides](https://cryostat.io/guides/) 2) the masthead ? button has a link to the upstream website containing the docs already.

Oh yehh sure. How about pointing it to the quickstarts page? I am thinking that the `/quickstarts` route could be in PRODUCTION (but remove search bar for now to till quickstart bumps core) while cards can be marked by feature level so that its only shown when needed but the entire view is still accessible.

> For a follow-up, really not for this PR, it should also be possible to perform certain actions against the group nodes themselves, ie "start a recording on this Deployment". This should be a single GraphQL query:

Sure niceeee of course, it should be not too difficult to extend from current actions for targets. Maybe follow-up after this ^^

maxcao13:
image
I think the Connect Url should be `Connect URL` or `ConnectURL`? 

image
I think it should either be `JVM ID` or `JvmID`

Everything looks really good like Andrew said, I'm amazed as well!

maxcao13:
image
This should be in sentence case, or the other popover texts should follow the this popover's title case.

maxcao13:
> Oh yehh sure. How about pointing it to the quickstarts page? I am thinking that the `/quickstarts` route could be in PRODUCTION (but remove search bar for now to till quickstart bumps core) while cards can be marked by feature level so that its only shown when needed but the entire view is still accessible.

I think this makes sense. Eventually at the end of the release, I was planning that the quickstarts and all of the actual content will be in production level. 

The navigating to blogs/docs, filing bugs, starting discussions, changing locale settings idea for a quick start card makes sense as well. I'll file an issue about it.

andrewazores:
image

image

image

image

```
$ https -v --auth-type=bearer --auth=$(oc whoami -t | base64) --
verify=false https://clustercryostat-sample-cryostat.apps-crc.testing/api/v2.1/discovery
GET /api/v2.1/discovery HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Authorization: Bearer c2hhMjU2fmhxQlhvdkFZSVhZWmlyZGpUOHVSRU5ycGhIaktqN19BNDRWV2tsZFBYdFkK
Connection: keep-alive
Host: clustercryostat-sample-cryostat.apps-crc.testing
User-Agent: HTTPie/3.2.1



HTTP/1.1 200 OK
cache-control: private
content-encoding: gzip
content-length: 1128
set-cookie: 8df8ccc1630e6819797dc02b7c042ec4=555eed54811c11786b32b6cd998d0ae9; path=/; HttpOnly; Secure; SameSite=None

{
    "data": {
        "result": {
            "children": [
                {
                    "children": [
                        {
                            "children": [
                                {
                                    "children": [
                                        {
                                            "children": [
                                                {
                                                    "children": [
                                                        {
                                                            "labels": {},
                                                            "name": "service:jmx:rmi:///jndi/rmi://10-217-0-98.cryostat.pod:9091/jmxrmi",
                                                            "nodeType": "Endpoint",
                                                            "target": {
                                                                "alias": "clustercryostat-sample-5748ff8d4c-jg647",
                                                                "annotations": {
                                                                    "cryostat": {
                                                                        "HOST": "10.217.0.98",
                                                                        "NAMESPACE": "cryostat",
                                                                        "POD_NAME": "clustercryostat-sample-5748ff8d4c-jg647",
                                                                        "PORT": "9091",
                                                                        "REALM": "KubernetesApi"
                                                                    },
                                                                    "platform": {
                                                                        "k8s.v1.cni.cncf.io/network-status": "[{\n    \"name\": \"openshift-sdn\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.217.0.98\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]",
                                                                        "k8s.v1.cni.cncf.io/networks-status": "[{\n    \"name\": \"openshift-sdn\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.217.0.98\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]",
                                                                        "kubectl.kubernetes.io/restartedAt": "2023-03-03T12:40:08-05:00",
                                                                        "openshift.io/scc": "restricted-v2",
                                                                        "seccomp.security.alpha.kubernetes.io/pod": "runtime/default"
                                                                    }
                                                                },
                                                                "connectUrl": "service:jmx:rmi:///jndi/rmi://10-217-0-98.cryostat.pod:9091/jmxrmi",
                                                                "jvmId": null,
                                                                "labels": {
                                                                    "app": "clustercryostat-sample",
                                                                    "component": "cryostat",
                                                                    "kind": "cryostat",
                                                                    "pod-template-hash": "5748ff8d4c"
                                                                }
                                                            }
                                                        }
                                                    ],
                                                    "labels": {
                                                        "app": "clustercryostat-sample",
                                                        "component": "cryostat",
                                                        "kind": "cryostat",
                                                        "pod-template-hash": "5748ff8d4c"
                                                    },
                                                    "name": "clustercryostat-sample-5748ff8d4c-jg647",
                                                    "nodeType": "Pod"
                                                }
                                            ],
                                            "labels": {
                                                "app": "clustercryostat-sample",
                                                "component": "cryostat",
                                                "kind": "cryostat",
                                                "pod-template-hash": "5748ff8d4c"
                                            },
                                            "name": "clustercryostat-sample-5748ff8d4c",
                                            "nodeType": "ReplicaSet"
                                        }
                                    ],
                                    "labels": {
                                        "app": "clustercryostat-sample",
                                        "app.kubernetes.io/name": "cryostat",
                                        "component": "cryostat",
                                        "kind": "cryostat"
                                    },
                                    "name": "clustercryostat-sample",
                                    "nodeType": "Deployment"
                                },
                                {
                                    "children": [
                                        {
                                            "children": [
                                                {
                                                    "children": [
                                                        {
                                                            "labels": {},
                                                            "name": "service:jmx:rmi:///jndi/rmi://10-217-0-92.cryostat.pod:9097/jmxrmi",
                                                            "nodeType": "Endpoint",
                                                            "target": {
                                                                "alias": "quarkus-test-58c98fcb9d-f4g85",
                                                                "annotations": {
                                                                    "cryostat": {
                                                                        "HOST": "10.217.0.92",
                                                                        "NAMESPACE": "cryostat",
                                                                        "POD_NAME": "quarkus-test-58c98fcb9d-f4g85",
                                                                        "PORT": "9097",
                                                                        "REALM": "KubernetesApi"
                                                                    },
                                                                    "platform": {
                                                                        "k8s.v1.cni.cncf.io/network-status": "[{\n    \"name\": \"openshift-sdn\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.217.0.92\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]",
                                                                        "k8s.v1.cni.cncf.io/networks-status": "[{\n    \"name\": \"openshift-sdn\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.217.0.92\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]",
                                                                        "openshift.io/generated-by": "OpenShiftNewApp",
                                                                        "openshift.io/scc": "restricted-v2",
                                                                        "seccomp.security.alpha.kubernetes.io/pod": "runtime/default"
                                                                    }
                                                                },
                                                                "connectUrl": "service:jmx:rmi:///jndi/rmi://10-217-0-92.cryostat.pod:9097/jmxrmi",
                                                                "jvmId": "kScIVo35NrRZVyHcb22nMlEtSjvGkysL4Dg2XqCcosA=",
                                                                "labels": {
                                                                    "deployment": "quarkus-test",
                                                                    "pod-template-hash": "58c98fcb9d"
                                                                }
                                                            }
                                                        }
                                                    ],
                                                    "labels": {
                                                        "deployment": "quarkus-test",
                                                        "pod-template-hash": "58c98fcb9d"
                                                    },
                                                    "name": "quarkus-test-58c98fcb9d-f4g85",
                                                    "nodeType": "Pod"
                                                }
                                            ],
                                            "labels": {
                                                "deployment": "quarkus-test",
                                                "pod-template-hash": "58c98fcb9d"
                                            },
                                            "name": "quarkus-test-58c98fcb9d",
                                            "nodeType": "ReplicaSet"
                                        }
                                    ],
                                    "labels": {
                                        "app": "quarkus-test",
                                        "app.kubernetes.io/component": "quarkus-test",
                                        "app.kubernetes.io/instance": "quarkus-test"
                                    },
                                    "name": "quarkus-test",
                                    "nodeType": "Deployment"
                                }
                            ],
                            "labels": {},
                            "name": "cryostat",
                            "nodeType": "Namespace"
                        },
                        {
                            "children": [
                                {
                                    "children": [
                                        {
                                            "children": [
                                                {
                                                    "children": [
                                                        {
                                                            "labels": {},
                                                            "name": "service:jmx:rmi:///jndi/rmi://10-217-0-90.myns1.pod:9097/jmxrmi",
                                                            "nodeType": "Endpoint",
                                                            "target": {
                                                                "alias": "quarkus-test-646ccb84d6-qbwn4",
                                                                "annotations": {
                                                                    "cryostat": {
                                                                        "HOST": "10.217.0.90",
                                                                        "NAMESPACE": "myns1",
                                                                        "POD_NAME": "quarkus-test-646ccb84d6-qbwn4",
                                                                        "PORT": "9097",
                                                                        "REALM": "KubernetesApi"
                                                                    },
                                                                    "platform": {
                                                                        "k8s.v1.cni.cncf.io/network-status": "[{\n    \"name\": \"openshift-sdn\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.217.0.90\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]",
                                                                        "k8s.v1.cni.cncf.io/networks-status": "[{\n    \"name\": \"openshift-sdn\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.217.0.90\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]",
                                                                        "openshift.io/generated-by": "OpenShiftNewApp",
                                                                        "openshift.io/scc": "restricted-v2",
                                                                        "seccomp.security.alpha.kubernetes.io/pod": "runtime/default"
                                                                    }
                                                                },
                                                                "connectUrl": "service:jmx:rmi:///jndi/rmi://10-217-0-90.myns1.pod:9097/jmxrmi",
                                                                "jvmId": "DDfsi3QpM_uXxZrmMggEZY4txuRY4Nhe-kqI95GAORE=",
                                                                "labels": {
                                                                    "deployment": "quarkus-test",
                                                                    "pod-template-hash": "646ccb84d6"
                                                                }
                                                            }
                                                        }
                                                    ],
                                                    "labels": {
                                                        "deployment": "quarkus-test",
                                                        "pod-template-hash": "646ccb84d6"
                                                    },
                                                    "name": "quarkus-test-646ccb84d6-qbwn4",
                                                    "nodeType": "Pod"
                                                }
                                            ],
                                            "labels": {
                                                "deployment": "quarkus-test",
                                                "pod-template-hash": "646ccb84d6"
                                            },
                                            "name": "quarkus-test-646ccb84d6",
                                            "nodeType": "ReplicaSet"
                                        }
                                    ],
                                    "labels": {
                                        "app": "quarkus-test",
                                        "app.kubernetes.io/component": "quarkus-test",
                                        "app.kubernetes.io/instance": "quarkus-test"
                                    },
                                    "name": "quarkus-test",
                                    "nodeType": "Deployment"
                                }
                            ],
                            "labels": {},
                            "name": "myns1",
                            "nodeType": "Namespace"
                        },
                        {
                            "children": [
                                {
                                    "children": [
                                        {
                                            "children": [
                                                {
                                                    "children": [
                                                        {
                                                            "labels": {},
                                                            "name": "service:jmx:rmi:///jndi/rmi://10-217-0-91.myns2.pod:9091/jmxrmi",
                                                            "nodeType": "Endpoint",
                                                            "target": {
                                                                "alias": "cryostat-sample-6dc66dc-47hw5",
                                                                "annotations": {
                                                                    "cryostat": {
                                                                        "HOST": "10.217.0.91",
                                                                        "NAMESPACE": "myns2",
                                                                        "POD_NAME": "cryostat-sample-6dc66dc-47hw5",
                                                                        "PORT": "9091",
                                                                        "REALM": "KubernetesApi"
                                                                    },
                                                                    "platform": {
                                                                        "k8s.v1.cni.cncf.io/network-status": "[{\n    \"name\": \"openshift-sdn\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.217.0.91\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]",
                                                                        "k8s.v1.cni.cncf.io/networks-status": "[{\n    \"name\": \"openshift-sdn\",\n    \"interface\": \"eth0\",\n    \"ips\": [\n        \"10.217.0.91\"\n    ],\n    \"default\": true,\n    \"dns\": {}\n}]",
                                                                        "openshift.io/scc": "restricted-v2",
                                                                        "seccomp.security.alpha.kubernetes.io/pod": "runtime/default"
                                                                    }
                                                                },
                                                                "connectUrl": "service:jmx:rmi:///jndi/rmi://10-217-0-91.myns2.pod:9091/jmxrmi",
                                                                "jvmId": null,
                                                                "labels": {
                                                                    "app": "cryostat-sample",
                                                                    "component": "cryostat",
                                                                    "kind": "cryostat",
                                                                    "pod-template-hash": "6dc66dc"
                                                                }
                                                            }
                                                        }
                                                    ],
                                                    "labels": {
                                                        "app": "cryostat-sample",
                                                        "component": "cryostat",
                                                        "kind": "cryostat",
                                                        "pod-template-hash": "6dc66dc"
                                                    },
                                                    "name": "cryostat-sample-6dc66dc-47hw5",
                                                    "nodeType": "Pod"
                                                }
                                            ],
                                            "labels": {
                                                "app": "cryostat-sample",
                                                "component": "cryostat",
                                                "kind": "cryostat",
                                                "pod-template-hash": "6dc66dc"
                                            },
                                            "name": "cryostat-sample-6dc66dc",
                                            "nodeType": "ReplicaSet"
                                        }
                                    ],
                                    "labels": {
                                        "app": "cryostat-sample",
                                        "app.kubernetes.io/name": "cryostat",
                                        "component": "cryostat",
                                        "kind": "cryostat"
                                    },
                                    "name": "cryostat-sample",
                                    "nodeType": "Deployment"
                                }
                            ],
                            "labels": {},
                            "name": "myns2",
                            "nodeType": "Namespace"
                        }
                    ],
                    "labels": {
                        "REALM": "f022c64f-935d-48e0-9a36-20d26d7ae335"
                    },
                    "name": "KubernetesApi",
                    "nodeType": "Realm"
                },
                {
                    "children": [],
                    "labels": {
                        "REALM": "315f186e-c237-4c8d-b40e-d4de24bfd2d6"
                    },
                    "name": "Custom Targets",
                    "nodeType": "Realm"
                }
            ],
            "labels": {},
            "name": "Universe",
            "nodeType": "Universe"
        }
    },
    "meta": {
        "status": "OK",
        "type": "application/json"
    }
}
```

andrewazores:
Not sure what a minimal reproducer for this bug is, but the one `quarkus-test` instance is "orphaned" off on the side there.

This looks great though:

image

andrewazores:
These are also awesome:

image
image

Though, I think "Realm Only" already implies "Collapse Singles", right? Maybe toggling "Realm Only" on should also toggle "Collapse Singles" on and disable it, or something.

andrewazores:
This is funny:

image

image

image

image

tthvo:
Ohh thats hilarious. Not sure what causes but if its latest commits now, then probably some testings for filters causes this. But i will look closer thanks^^

> Not sure what a minimal reproducer for this bug is, but the one quarkus-test instance is "orphaned" off on the side there.

Oh this is also an ongoing problem with large number of nodes or if the users drag the node outside view. Ideally, we could pan them into view but maybe we should keep it position to save initializing time (node might "fly" around). The users can always click on control button that fit everything in screen.  There is also a help icon on toolbar to do this.

> Though, I think "Realm Only" already implies "Collapse Singles", right? Maybe toggling "Realm Only" on should also toggle "Collapse Singles" on and disable it, or something.

Yeh suree^^Thanks

tthvo:
> The users can always click on control button that fit everything in screen. There is also a help icon on toolbar to do this.

Actually, i think maybe a tiny warning alert somewhere to say that "Hey, there are some nodes being out of view". I will find a good place to put it.

andrewazores:
It's not that the node was out of view, it's that it doesn't appear to be rendered as part of its groups - like it looks like it doesn't have a deployment or namespace.

tthvo:
> It's not that the node was out of view, it's that it doesn't appear to be rendered as part of its groups - like it looks like it doesn't have a deployment or namespace.

Opps I was looking at the wrong image. But yeh was looking into that too.

tthvo:
I am hoping its not internal package bugs. Here, I only showing `cryostat` namespace. Both deployments are shown here.

Screenshot from 2023-03-03 16-11-44

tthvo:
I have some thoughts for search (`Find by...`) but seems like adding it might make the toolbar too clustered. What do you think?

andrewazores:
Everything looks great, at least in the podman case, right now. What other ideas did you have for search/filter?

tthvo:
>  What other ideas did you have for search/filter

Ahh opps I forgot to mention...

## Right next to filters

Seems like it does not look too bad.

Screenshot from 2023-03-07 11-24-01

## Another level

Taking too much spaces from graph view and chips are rendered above it (odd). Fixable tho.

Screenshot from 2023-03-07 11-28-34

## Hook search into select typeahead

Select component is typeahead while allowing a callback prop on text input that we can hook into like a search input.

Its pretty nice and compact. Tho, this will add quite lots of complexity to filters and won't be using matched expressions.

Screenshot from 2023-03-07 11-29-18

andrewazores:
Options 1 and 3 look good. Can they be combined? Would it make sense to put a "match expression" selection in that second dropdown, which in the screenshots is "Target: Jvm ID", so that the text field becomes a match expression field when that kind of filter is selected?

tthvo:
> Would it make sense to put a "match expression" selection in that second dropdown, which in the screenshots is "Target: Jvm ID", so that the text field becomes a match expression field when that kind of filter is selected?

Not sure I got it correctly but instead of `Target: JVM ID`, it would be `target.jvmId`? How about groupings?

> Can they be combined?

I havent got any ideas to combine but do you have anything in mind? Maybe some sketches?

andrewazores:
Sorry, what I meant was something like this:

image

I just did this by directly editing the page HTML, so I don't have any useful code to share or insight into whether the library code can be made to work this way :-)

tthvo:
Ahh thanks! Make sense to me! Just need to see how the filters turn out because its design right now will break with nested groups...

tthvo:
Just a heads up: Latest commits include implementations for target-related/relevant resource table so it needs backend https://github.com/cryostatio/cryostat/pull/1389 (will throw warning if not).

tthvo:
Screenshot from 2023-03-08 16-00-44

tthvo:
I think this is ready to be reviewed again. Filters are now also considering node types so filter data are no long static and based on a discovery tree at a given time.

Tested with a mock response json from above (i.e. under `@test/resources`). Seems to work fine. Will check more if there are any irregularities.

Let me know what u think :D

andrewazores:
https://github.com/cryostatio/cryostat-web/pull/891#issuecomment-1453890002

> Though, I think "Realm Only" already implies "Collapse Singles", right? Maybe toggling "Realm Only" on should also 
toggle "Collapse Singles" on and disable it, or something.

I think this got skipped, or at least I'm not seeing anything like it in the latest test image. It's pretty minor but I think could help reduce some confusion between the two options.

andrewazores:
image

This "sample" item can be removed now I think, right?

andrewazores:
Screenshot_2023-03-09_13-47-22

Seems like the matchExpression testing API call is probably being called more often than it needs to be?

tthvo:
> This "sample" item can be removed now I think, right?

Ahh each item has its own feature level which show removes it. There is a bug there. I will fix it...

tthvo:
> I think this got skipped, or at least I'm not seeing anything like it in the latest test image. It's pretty minor but I think could help reduce some confusion between the two options.

Ahh must have lost among comments. This should be included now.

maxcao13:
image
Would it make sense to make the `Target` bolded or shown in a different styling/format from the targetId?

tthvo:
Screenshot from 2023-03-09 14-51-25

Like this right?

maxcao13:
image

The filter label says Connection URL but the Filter typeahead text thing says ConnectionUrl. Maybe change them to Connection URL ? In the chip filter group too.

tthvo:
Seems like there is also a mismatching in nodeType identifier (maybe case or spacing) that causes the custom target not being filtered out. Looking into it...

tthvo:
> Seems like the matchExpression testing API call is probably being called more often than it needs to be? This is what the network request log looks like after just switching between selected target nodes once.

There was an extra hook deps that caused this (oddly eslint didnt see this). Could you check again if it still happens?

maxcao13:
When the viewport is smaller and the filter toolbar is open, the dropdown doesnt hover where it should correctly as seen here.
image

maxcao13:
image
Would it make sense to allow these links to open in a new tab? Or have them link to a subroute that has a button to go back where you were in the topology view, but that feature seems a little much.

src/app/Rules/CreateRule.tsx
```diff
@@ -101,6 +101,15 @@ const Comp: React.FC = () => {
   const [errorMessage, setErrorMessage] = React.useState('');
   const [loading, setLoading] = React.useState(false);
 
+  const targetSubject = React.useRef(new BehaviorSubject(NO_TARGET)).current;
```
maxcao13, line 104: I'm wondering what was the reason for using this vs the old `context.target.target()`?

src/app/Shared/Redux/Filters/TopologyFilterSlice.tsx
```diff
@@ -0,0 +1,269 @@
...
+      };
+    };
+  };
+  targetFilters: {
+    category: string;
+    filters: {
+      [nodeType: string]: {
+        Alias: string[];
+        ConnectionUrl: string[];
+        s;
```
maxcao13, line 132: What is this `s` for?

tthvo: Nope thats the typo. I should not be there. Must be during typing XD

src/app/Rules/CreateRule.tsx
tthvo: The `context.target.target` is for global target selection. Here, we just want to see detail without changing that. 

Also, the reason behind using a subject is to ensure there is a single observable to subscribe to to its entire lifecycle. And we can control to ignore previous inner observable when the target node change.

tthvo:
> (Or have them link to a subroute that has a button to go back where you were in the topology view, but that feature seems a little much.)
> 
> EDIT: Maybe not actually now that I think about it as it would create more clients.

I think since we are pushing a new route in. The user can also go back using the browser back button.

src/app/Rules/CreateRule.tsx
maxcao13: Ah, I see thanks for that.

tthvo:
> When the viewport is smaller and the filter toolbar is open, the display options dropdown doesnt hover where it should correctly as seen here.

Screenshot from 2023-03-09 16-50-57

maxcao13:
> > (Or have them link to a subroute that has a button to go back where you were in the topology view, but that feature seems a little much.)
> > EDIT: Maybe not actually now that I think about it as it would create more clients.
> 
> I think since we are pushing a new route in. The user can also go back using the browser back button.

I meant with the selections it was on before, like having the expanded view of whatever node JVM we were looking at before clicking on the resource link. But I think that would require more internal state management that might not be that useful other than for this one feature so it's not that important.

tthvo:
Oh i see! Should be possible using breadcrumbs without history state. But agreed that it adds much complexity than intended.

andrewazores:
> Sorry not a bug tho. I got a little confused. Just a question: Would we want to category target filters based on their nodeType (i.e. JVM, Endpoint, Custom Target) like right now? This would mean if you choose a value in a category (nodeType), that filter is only applied to targets of that type.

Not sure if I'm totally following. Could you give a concrete example of what two such selections would be and what the options for each are?

tthvo:
> Not sure if I'm totally following. Could you give a concrete example of what two such selections would be and what the options for each are?

Oh for example (running in local dev):

1. Create a custom target (localhost:0).
2. Now, in the filter for targets, each select drop-down has 2 groups (i.e. JVM and Custom Target).
3. Let's say we are considering JVM ID.
4. Select a JVM ID of the JVM group.
5. Notice that that matching target is showned along with Custom Target (this is because the filter you just chose does not apply to Custom Target).
6. You can do the same in the opposite way. Choose the new Custom Target JVM ID. See that all other JVMs are still shown.

andrewazores:
Ah, I see. I think what makes sense is that the filter should be applied to targets within that type, and targets of other types should probably just be removed too. Selecting a filter that is specific to a type seems like it should also implicitly filter out things not of that type.

image

image

I think this looks confusing, like both of those nodes have the same JVM ID.

tthvo:
>  I think what makes sense is that the filter should be applied to targets within that type, and targets of other types should probably just be removed too.

Right! Then all targets are treated the same way regardless of their node type then. Tho, this means target filters and group filters will behave differently.

A selection of a group type will only filter groups of that types. Others will be unaffected unless they are parents or children of those nodes. I think this makes sense in the context of groupings.

src/app/Topology/Actions/CreateTarget.tsx
```diff
@@ -0,0 +1,522 @@
...
+          </Alert>
+        </FormAlert>
+      )}
+      <Flex className={css(className)} direction={{ default: 'column' }}>
+        <FlexItem alignSelf={{ default: 'alignSelfCenter' }}>
+          <Tooltip
+            content={
+              _actionEnabled
+                ? `Click to test${validation.option !== ValidatedOptions.default ? ' again' : ''}.`
+                : 'Please provide a valid connection URL.'
```
maxcao13, line 485: Connection URL ?

src/app/Topology/Actions/NodeActions.tsx
```diff
@@ -0,0 +1,149 @@
...
+      services.target.setTarget(targetNode.target);
+      history.push('/rules/create');
+    },
+    title: 'Create Automated Rules',
+  },
+  { isSeparator: true },
+  {
+    action: (element, { services }, track) => {
+      const targetNode: TargetNode = element.getData();
+      track(services.api.deleteTarget(targetNode.target).subscribe(() => console.log('DONE')));
```
maxcao13, line 144: Does this need to be here?

src/app/Topology/GraphView/CustomNode.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  const classNames = React.useMemo(() => {
+    const additional = expression === '' || isTargetMatched(data, expression) ? '' : 'search-inactive';
+    return css('topology__target-node', additional);
+  }, [data, expression]);
+
+  return (
+    <Layer id={contextMenuOpen ? TOP_LAYER : DEFAULT_LAYER}>
+      {/* eslint-disable-next-line @typescript-eslint/no-explicit-any */}
+      <g className={classNames} id={'target-node-visual-group'} ref={hoverRef as any}>
```
maxcao13, line 125: you can do `hoverRef as React.LegacyRef<SVGGElement>` any get rid of the eslint warning I think

tthvo: Nice thanks!

src/app/Topology/Shared/Entity/EntityDetails.tsx
```diff
@@ -0,0 +1,598 @@
...
+        helperTitle: 'Alias',
+        helperDescription: 'Connection Nickname (same as Connection URL if not specified).',
+        content: serviceRef.alias,
+      },
+      {
+        key: 'JVM ID',
+        title: 'JVM ID',
+        helperTitle: 'JVM ID',
+        helperDescription: 'The ID of the current JVM.',
+        content: serviceRef.jvmId || <EmptyText text="No Jvm ID" />,
```
maxcao13, line 197: Jvm -> JVM

tthvo:
Here is the latest filters for target (need to clear local storage). Unfortunately, we can't remove options that belongs to targets in filtered-out group (searching upwards to find parent will need to check many levels, not ideal for performance).

Screenshot from 2023-03-09 19-54-49

src/app/Topology/Shared/Entity/EntityDetails.tsx
```diff
@@ -0,0 +1,598 @@
...
+    services.api,
+    isOwned,
+    targetSubject,
+    resourceType,
+    services.notificationChannel,
+    setResources,
+    setError,
+  ]);
+
+  React.useEffect(() => {
```
maxcao13, line 419: Are there duplicate useEffects here or am I missing something?...

If there is not, there seems to be a lot of duplicated code, so I think there could be some refactoring here...

tthvo: Nope this one is for DELETE events. The one above is for ADD event.

tthvo: But I think there might be a way to refactor it.

src/app/Topology/Shared/Entity/utils.ts
```diff
@@ -0,0 +1,337 @@
...
+        return of([...arr.filter((probe) => !probeIds.includes(probe.id)), ...probes]);
+      };
+    case 'automatedRules':
+      return (arr: Rule[], eventData: any, removed?: boolean) => {
+        const rule: Rule = eventData.message;
+
+        return apiService.isTargetMatched(rule.matchExpression, target).pipe(
+          map((ok) => {
+            if (ok) {
+              // return apiService.isTargetMatched(credential.matchExpression, )
```
maxcao13, line 275: What's this comment for?

tthvo: Another copy and paste but forgot to clean up :D

src/app/Topology/Shared/Entity/EntityDetails.tsx
tthvo: Refactored :D

src/app/Topology/Shared/TranslatedText.tsx
```diff
@@ -0,0 +1,49 @@
...
+ */
+import { Text, TextProps } from '@patternfly/react-core';
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+
+export interface TranslatedTextProps extends TextProps {
+  transKey: string;
+}
+
+export const TranslatedText: React.FunctionComponent<TranslatedTextProps> = ({ transKey, ...props }) => {
```
maxcao13, line 46: Is this used anywhere?

tthvo: Nope but its there in prep for follow-up PR that localize topology.

maxcao13:
image
Would it be nice for the user for this to be validated to be successful or not based on the current text? If the current matchexpression matches 0, maybe have a `ValidatedOptions.error`, and if it matches at least 1 target, have some indication that the expression is successful? 

Right now using the search bar, I'm not exactly sure when any targets are matched and where they are on the topology view when they do.

tthvo:
> Right now using the search bar, I'm not exactly sure when any targets are matched and where they are on the topology view when they do.

Matched targets are shown as normal and unmatched are faded away. Try `false` or something else?

> Would it be nice for the user for this to be validated to be successful or not based on the current text? If the current matchexpression matches 0, maybe have a ValidatedOptions.error, and if it matches at least 1 target, have some indication that the expression is successful?

This bar is only for searching tho...I think its better not to show any validation since its not a form?

src/app/Topology/Shared/TranslatedText.tsx
tthvo: Would you prefer keeping it for removing and adding to the next PR?

maxcao13:
Ah, I see that's fine according to the guideline then, but then in the list view, it seems like the search bar does not do anything?

tthvo:
> but then in the list view, it seems like the search bar does not do anything?

Ohh thanks for pointing that out! Totally forgot to implement searching for listview. Will do it now.

tthvo:
Implemented just now. ListView search behaviour is a different from GraphView:

1. All groups will be expanded on search.
2. Only matching targets are shown (no fading away).

maxcao13:
During the search debouncing, would it make sense to have some sort of loading visual during this time?

maxcao13:
image

Weird. Seems like if the expression is `target.alias == "vertx-fib-demo-1` it will match everything.

tthvo:
> During the search debouncing, would it make sense to have some sort of loading visual during this time?

Hmm maybe not? I added debouncing to reduce re-rendering. Intermediate loading state will add back that workload so its sorta cancel out the optimization. 

> Weird. Seems like if the expression is target.alias == "vertx-fib-demo-1 it will match everything.

Not sure but from the image, every one of them has the alias `vertx-fib-demo-1`?

maxcao13:
Interesting way to recreate:
[Screencast from 03-10-2023 02:48:03 PM.webm](https://user-images.githubusercontent.com/69827283/224413911-86a06fd9-244c-44ee-abe3-cb2a29719911.webm)
You can use any match expression to do this.

tthvo:
Ahh i see it now! Thanks! I think the match expression actually with a single `=` actually assign the value to the target object. Here is the implementation that I pull from match Expression views.

```ts
export const isTargetMatched = ({ target }: TargetNode, matchExpression: string): boolean => {
  try {
    const f = new Function('target', `return ${matchExpression}`);
    const res = f(target);

    if (typeof res === 'boolean') {
      return res;
    }
    return false;
  } catch (err) {
    return false;
  }
};
```

Let me try deep copy the object before comparing...

tthvo:
Seems to fix. Could you try one more time?

maxcao13:
I'm not sure if that's the most safe. It seems like you can do javascript injection with that prompt.
For example typing `alert("hello")` in the prompt will actually send an alert function.

image

andrewazores:
It's an injection, but the runtime is the user's own browser. The same effect could be had by just opening the browser console and writing in the same JS snippet.

andrewazores:
Screenshot_2023-03-10_15-42-39

tthvo:
Right was wondering about this too. I guess its user's own risk then. Not sure if there is any way we can sanitize the match expression in a way?

andrewazores:
It isn't really needed since it's running entirely on the frontend. When a backend resource is created that contains a matchExpression we already do have and apply the MatchExpressionValidator class on the backend which parses the expression and rejects it if it contains anything dangerous like function calls or definitions. This way any malicious script won't be executed on the server side. The user can do it to their own client machine if they want :-)

tthvo:
Right that makes sense! Then I think deepcopy comparing object should solve the above issue. I think we need to change that in match expression evaluator component too.

andrewazores:
Screenshot_2023-03-10_15-50-49

This is an existing "vulnerability" from before this PR anyway, but I don't think it poses any real risk since again this is exactly equivalent to the built-in JS console in the browser.

maxcao13:
Gotcha, makes sense to me then. Thanks for the explanation!

maxcao13:
One last thing: 
image

I think these Details fields should be selectable like text?

tthvo:
> I think these Details fields should be selectable like text?

Unfortunately, the TreeView component seems to wrap that view in a button which causes it not to be selectable...

src/app/Shared/MatchExpressionEvaluator.tsx
```diff
@@ -61,37 +59,34 @@ import {
...
-  React.useEffect(() => {
-    if (!matchExpression || !target?.connectUrl) {
+    if (!matchExpression || !target.connectUrl) {
       setValid(ValidatedOptions.default);
       return;
     }
     try {
       const f = new Function('target', `return ${matchExpression}`);
-      const res = f(target);
+      const res = f(_.cloneDeep(target));
```
andrewazores, line 89: Probably worth extracting this to a utility function since it's used in two places now.

andrewazores:
image

image

Typing in the filter box doesn't seem to do anything?

tthvo:
A little weird that the compareTo function is given different arguments if select Option is not in a group. Fixed now thanks!

maxcao13 (approved):
No other comments, looks great to me! Awesome job on this.

Issue #895, opened 2023-02-23
[Task] Add JVM Version, JVM Start Time, other MBean properties to the JVM Details dashboard card


andrewazores:
Depends on #894

PR #896, opened 2023-02-23
feat(dashboard): add JVM Details card

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #894
Fixes: #895
Depends on https://github.com/cryostatio/cryostat-web/pull/891

## Description of the change:
This adds a simple dashboard card that simply displays basic information about the selected target JVM such as its connection URL, alias, and labels/annotations.

## Motivation for the change:
This provides users with a nicer presentation of the data currently displayed on the `<TargetSelect/>` component expansion, which the user might need to know in order to define matchExpression credentials, automated rules, or to filter/sort through the All Archives table. A follow-up PR will add further MBean-based information to this card as well.

## How to manually test:
2. Select a target JVM
3. Add the JVM Details dashboard card
4. Check that the details are updated with different target selections and that the details match the target

image


tthvo (commented):
Looks great to me! 

Just wondering because I have a similar in #891 that looks like this with a bit styling, handling for empty fields and warnings. Do u want to apply these here or maybe after #891?


Screenshot from 2023-02-23 15-53-05

Screenshot from 2023-02-23 15-53-51

andrewazores:
I'll wait for yours and pull it out to a shared component.

andrewazores:
image

tthvo:
No features for #895 right?

andrewazores:
I could (re)add that into the new shared component.

src/app/Dashboard/JvmDetails/JvmDetailsCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+      labels: {},
+    };
+  }, [target]);
+
+  return (
+    <DashboardCard
+      id={`${props.dashboardId}`}
+      dashboardId={props.dashboardId}
+      cardSizes={JvmDetailsCardSizes}
+      isCompact
```
tthvo, line 80: We need to pass in `isFullHeight`, `isDraggable` and `isResizable` from props. Or just `{...props}` following #905.

src/app/Dashboard/JvmDetails/JvmDetailsCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { DashboardCard } from '../DashboardCard';
+import '@app/Topology/styles/base.css';
+
+export interface JvmDetailsCardProps extends DashboardCardProps {}
+
+export const JvmDetailsCard: React.FC<JvmDetailsCardProps> = (props) => {
+  const context = React.useContext(ServiceContext);
+  const addSubscription = useSubscriptions();
+
+  const [target, setTarget] = React.useState(undefined as Target | undefined);
```
tthvo, line 57: Probably just 

```
const [target, setTarget] = React.useState<Target>();
```

src/app/Dashboard/JvmDetails/JvmDetailsCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  React.useEffect(() => {
+    addSubscription(context.target.target().subscribe(setTarget));
+  }, [addSubscription, context.target, context.api, setTarget]);
+
+  const wrappedTarget = React.useMemo((): TargetNode | undefined => {
```
tthvo, line 63: I think we can do this (memo the entire entity instead of just targetNode).

```ts
const wrappedTarget = React.useMemo(() => {
    if (!target) {
      return undefined;
    }
    return {
      getData: (): TargetNode => ({
        name: target.alias,
        target,
        nodeType: NodeType.JVM,
        labels: {},
      })
    };
  }, [target]);

```

tthvo:
> I could (re)add that into the new shared component.

Sure! I think I have a const here to define them. I guess we can extract into a common function to return this array.

https://github.com/cryostatio/cryostat-web/blob/c704ceb281c937a2466c88a1f41741c37a4d13a2/src/app/Topology/Shared/Entity/EntityDetails.tsx#L176

andrewazores:
image

image

src/app/Dashboard/JvmDetails/JvmDetailsCard.tsx
tthvo: Ping :D

src/app/Topology/Shared/Entity/EntityDetails.tsx
```diff
@@ -214,24 +229,110 @@ export const TargetDetails: React.FC<{
...
-              <DescriptionListTermHelpTextButton>{d.title}</DescriptionListTermHelpTextButton>
-            </Popover>
-          </DescriptionListTermHelpText>
-          <DescriptionListDescription style={{ userSelect: 'text', cursor: 'text' }}>
-            {d.content}
-          </DescriptionListDescription>
-        </DescriptionListGroup>
-      ))}
+      {_transformedData.map(mapSection)}
+      <ExpandableSection
```
tthvo, line 237: This section should be outside the part `<DescriptionList {...props} columnModifier={columnModifier}>` since it will create a nested description list?

src/app/Topology/Shared/Entity/EntityDetails.tsx
```diff
@@ -214,24 +229,110 @@ export const TargetDetails: React.FC<{
...
+        onToggle={onToggle}
+        isExpanded={isExpanded}
+      >
+        <MBeanDetails isExpanded={isExpanded} connectUrl={serviceRef.connectUrl} />
+      </ExpandableSection>
     </DescriptionList>
   );
 };
 
+const MBeanDetails: React.FC<{ isExpanded: boolean; connectUrl: string }> = ({ isExpanded, connectUrl }) => {
```
tthvo, line 248: This should also receive the `columnModifier` from parent and pass down to its `DescriptionList`. This is being used for ListView.

Screenshot from 2023-03-13 22-00-56

src/app/Topology/Shared/Entity/EntityDetails.tsx
```diff
@@ -214,24 +229,110 @@ export const TargetDetails: React.FC<{
...
+                    version
+                    arch
+                    availableProcessors
+                  }
+                }
+              }
+            }`,
+            { connectUrl }
+          )
+          .pipe(map((resp) => resp.data.targetNodes[0].mbeanMetrics))
```
tthvo, line 278: `resp.data` might be null if there is an error. How about adding a `catchError` like this:

```ts
.pipe(
  map((resp) => resp.data.targetNodes[0].mbeanMetrics),
  catchError((_) => of({})) 
)
```

This way, any error will be caught and just set empty.

src/app/Topology/Shared/Entity/EntityDetails.tsx
```diff
@@ -214,24 +229,110 @@ export const TargetDetails: React.FC<{
...
+        content: dayjs(mbeanMetrics?.runtime?.startTime || 0)
+          .tz(dateTimeFormat.timeZone.full)
+          .format('LLLL'),
+      },
+      {
+        key: 'JVM Version',
+        title: 'JVM Version',
+        helperTitle: 'JVM Version',
+        helperDescription: 'The version of the JVM.',
+        content: mbeanMetrics.runtime?.vmVersion,
```
tthvo, line 300: Looks like the content of these configurations can be undefined or null. We should show some indication instead of blank space. For example, 

```ts
 <EmptyText text="No JVM ID" />
```

src/app/Topology/Shared/Entity/EntityDetails.tsx
```diff
@@ -167,11 +170,23 @@ export const EntityDetails: React.FC<EntityDetailsProps> = ({ entity, className,
   return <div className={css(className)}>{viewContent}</div>;
 };
 
+const mapSection = (d) => (
```
tthvo, line 173: Maybe we need to extract a type for d:

```ts
type DescriptionConfig = {
  key: string,
  title: React.ReactNode,
  helperTitle: React.ReactNode,
  helperDescription: React.ReactNode,
  content: React.ReactNode,
},
```

tthvo (commented):
Looks good! Just some more comments :D

src/app/Topology/Shared/Entity/EntityDetails.tsx
```diff
@@ -214,24 +229,110 @@ export const TargetDetails: React.FC<{
...
+  }, [isExpanded, addSubscription, connectUrl, context.api, setMbeanMetrics]);
+
+  const _collapsedData = React.useMemo(() => {
+    return [
+      {
+        key: 'Start Time',
+        title: 'Start Time',
+        helperTitle: 'Start Time',
+        helperDescription: 'The time when this JVM process started.',
+        content: dayjs(mbeanMetrics?.runtime?.startTime || 0)
```
tthvo, line 291: We should handle cases where value is negative? I am not quite what is the meaning behind negative values..

Screenshot from 2023-03-13 22-39-06

maxcao13:
image

I think it should have a `View Topology` action when the EntityDetails is on the Dashboard.

src/app/Topology/Shared/Entity/EntityDetails.tsx
andrewazores: I think that if this value is actually what the MBean interface reports, it should be displayed as-is. It looks odd, but the various fields may have different meanings for negative values. Usually it means something about the JVM or the host OS not supporting that kind of performance metric, which would be useful for the user to know. It may be the case that specific values have specific meanings so I would rather not try to hide the the value with some "friendlier" UI label.

andrewazores:
> I think it should have a View Topology action instead of View Dashboard when the EntityDetails is on the Dashboard.

That's a good point. This will take a little more refactoring to the shared component.

src/app/Dashboard/JvmDetails/JvmDetailsCard.tsx
andrewazores: It's not showing up as changed here since it happened lower down in the patch, but I did add the `{..props}` destructuring in a recent commit.

tthvo: Ah thanks! Saw it just now!

src/app/Topology/Shared/Entity/EntityDetails.tsx
tthvo: Oh okay right! That make sense!

tthvo: Oh yes that was the agent on `9988`!

andrewazores:
> > I think it should have a View Topology action instead of View Dashboard when the EntityDetails is on the Dashboard.
> 
> That's a good point. This will take a little more refactoring to the shared component.

Any ideas on this @tthvo ? It looks like I could hack it together by adding some property to the "View Dashboard" action in `NodeActions` which the `utils actionFactory` can use to filter the action out so it doesn't appear in this context. And is there a good way to add a "View Topology" action that does more than simply link to the Topology view? It would be cool if it could route to that view and also have the corresponding node selected and the sidebar panel opened so that the EntityDetails component is visible in the new context.

tthvo:
> And is there a good way to add a "View Topology" action that does more than simply link to the Topology view? It would be cool if it could route to that view and also have the corresponding node selected and the sidebar panel opened so that the EntityDetails component is visible in the new context.

I think we can add some state to route location or use query parameter. This could be done in `Topology.tsx` but require some refactoring. I will look into it...

andrewazores:
In the case that the Topology view is routed to directly with a preselected entity state like that, is it possible to just clear out any filters? I guess that would mean that whatever fires the history navigation event with the preseleted state should also first fire a Redux intent to clear the filters?

tthvo:
Oh sure! Just in time for my comments XD

src/app/Topology/Shared/Entity/EntityDetails.tsx
```diff
@@ -214,24 +245,122 @@ export const TargetDetails: React.FC<{
...
+    <>
+      <DescriptionList {...props} columnModifier={columnModifier}>
+        {_transformedData.map(mapSection)}
+      </DescriptionList>
+      <ExpandableSection
+        toggleText={isExpanded ? 'Show less' : 'Show more'}
+        onToggle={onToggle}
+        isExpanded={isExpanded}
+      >
+        <MBeanDetails isExpanded={isExpanded} connectUrl={serviceRef.connectUrl} />
```
tthvo, line 260: we need to pass in columnModifier here?

tthvo: Just this one left. Others look very nice!

tthvo (dismissed):
Looks great to me!

andrewazores:
Bah, I need to rebase to sign commits.

maxcao13 (approved):

tthvo (approved):

Issue #765, opened 2022-12-14
[Task] Multiple Dashboard Layouts


andrewazores:
Depends on https://github.com/cryostatio/cryostat-web/issues/764

The user should be able to name their dashboard layout, then create a new dashboard layout that starts as blank and allows the user to graphically build the new layout. There should be some Switch or other selector somewhere that allows the user to swap between dashboard layouts on the fly. Perhaps this should be implemented as a card that can be added to the dashboard itself, or maybe it should be part of the default state (before entering the wizard) of the Add card.

Issue #764, opened 2022-12-14
[Task] Import/export dashboard layouts


andrewazores:
Now that the Dashboard has a serialized form that is loaded to and from localstorage, we should also add some simple controls that allow the user to download that serialized form to their workstation, and to upload those files to the Dashboard. Only a single file upload would be allowed at a time, and the contents would completely replace the current layout state.

PR #899, opened 2023-02-24
feat(dashboard): import/export dashboards and switch between layouts

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #764 
Fixes: #765 
Depends on https://github.com/cryostatio/cryostat-web/pull/891

## Description of the change:
Allows users to:
- Import/export dashboard layouts:
- Create new empty layouts within the UI
- Delete layouts
- Switch between layouts
## Motivation for the change:
See #764, #765 

## How to manually test:
Run normally, add try out the new dashboard configuration toolbar.
Localization/tests will come later when most is approved. 

Looking for UI improvements and feedback.

# WARNING: 
checking out this PR will change your dashboard configuration local storage keys because of some refactoring, the current dashboard is now nested in another level because of the newly added feature of having more than one dashboard configuration persisting at once.


andrewazores:
Nicely done, seems to fit all the requirements I had in mind for this set of features and looks pretty nice already.

image

This looks a little funny where the toolbar meets the tops of the cards with no gap, but then there's a bit of radiusing on the corner intersection and the card grab handles. I'm not sure if a gutter between the toolbar and the cards is best though, since then the toolbar would graphically look like its own standalone element - but maybe it should be? It is also a little funny when there are two or more rows of cards and the toolbar somewhat looks like it's only part of the top row and maybe only would affect the top row.

image

In Thuan's topology PR the target selection widget card is going away and replaced by a top full-width toolbar-like component - maybe it makes sense for this toolbar to be attached to that component on the dashboard, or at least to be placed similarly?

maxcao13:
I will also be waiting on the topology view then, until then, I will just do cleanup. Implementation details and other unrelated components are still ok to be reviewed though.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -1134,6 +1136,33 @@ export class ApiService {
     );
   }
 
+  downloadDashboardLayout(layout: DashboardLayout): void {
+    const serializedLayout = this.getSerializedDashboardLayout(layout);
+    const filename = `${layout.name}${nanoid(5)}.json`;
```
andrewazores, line 1141: This filename with the random ID ends up being a little weird looking. If there is no random ID, I think the browser should (hopefully) handle any on-disk naming collisions for us?

Also I'd suggest putting something like `cryostat-dashboard-` as a prefix so that it's more obvious to the user what this `.json` file is later when they see it in their downloads directory.

andrewazores:
Clicking outside of the select dropdown doesn't close it:

image

So I can do this:

image

Just a minor visual thing.

The new UI looks great when I'm not trying to break it:

image

src/app/Dashboard/DashboardLayoutConfig.tsx
```diff
@@ -0,0 +1,347 @@
...
+        >
+          <MenuContent maxMenuHeight="16em" id="dashboard-layout-menu-content" style={{ overflow: 'unset' }}>
+            {menuGroups(t('DashboardLayoutConfig.MENU.FAVORITES'), true, (l) => favorites.includes(l.name))}
+            <Divider />
+            {menuGroups(t('DashboardLayoutConfig.MENU.OTHERS'), false, (l) => !favorites.includes(l.name))}
+          </MenuContent>
+        </Menu>
+      </Dropdown>
+    );
+  }, [t, onLayoutSelect, onActionClick, onToggle, menuGroups, favorites, isSelectorOpen, currLayout.name]);
```
tthvo, line 323: Missing `isSelectorOpen` as deps here...

tthvo: Seems to be causing

> Clicking outside of the select dropdown doesn't close it:

andrewazores:
I created this dashboard, downloaded it, deleted it, and tried to upload it:

```json
{"list":[{"name":"MBeanMetricsChartCard","span":4,"props":{"chartKind":"Process CPU Load","duration":60,"period":10,"themeColor":"blue"}}],"name":"test"}
```

image

This message might also be a little confusing:

image

Should the error message say something about the name already being in use instead?

andrewazores:
Here's a weird one. Renaming a dashboard by clicking the Create button works, but hitting Enter on my keyboard makes the whole page reload (and the dashboard isn't renamed).

maxcao13:
Funny, I have to prevent default and handle it manually on a form submission because pressing "Enter" on a form submits it and refreshes the page. There was also a query parameter with name="{dashboardName}" after the refresh. :D

andrewazores:
Looking good. Maybe there should be a deletion dialog confirmation prompt for dashboards?

andrewazores:
Favouriting + renaming seems to have some funny interactions. If I favourite a dashboard and then rename it, it loses its favourite status. If I rename it back to the original name then it regains its favourite status.

andrewazores:
Sweet, everything seems to be working perfectly now.

andrewazores (approved):
Implementation all looks good to me.

src/app/TargetView/TargetView.tsx
```diff
@@ -45,6 +46,7 @@ import { NoTargetSelected } from './NoTargetSelected';
 import { TargetContextSelector } from './TargetContextSelector';
 
 interface TargetViewProps {
+  appendDashboard?: boolean;
```
tthvo, line 49: I think this could be a prop like: 

```ts
attachments?: React.ReactNode;
```

This way, parent can supply any attachments needed (easy to extend later if other views want to add something below context selector). Wdu think?

src/app/utils/utils.ts
```diff
@@ -47,17 +47,17 @@ const DAY_MILLIS = 24 * HOUR_MILLIS;
 // [     0,    1,    2,    3     ] array
 //       0     1     2     3       indexes
 // {  0  |  1  |  2  |  3  |  4  } gap indices (drop zones)
-export function moveDashboardCard(arr: CardConfig[], from: number, gapIndex: number) {
+export function moveDashboardCard(cards: CardConfig[], from: number, gapIndex: number) {
```
tthvo, line 50: This function and below seems to work for any array. We can generalize it by removing type assertions here so we don't have to import redux slices.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -1117,6 +1118,34 @@ export class ApiService {
...
+    const filename = `cryostat-dashboard-${layout.name}.json`;
+    const file = new File([serializedLayout], filename);
+    const resourceUrl = URL.createObjectURL(file);
+    this.downloadFile(resourceUrl, filename);
+    setTimeout(() => URL.revokeObjectURL(resourceUrl), 1000);
+  }
+
+  private getSerializedDashboardLayout(layout: DashboardLayout): string {
+    const serialCards: SerialCardConfig[] = [];
+    for (const card of layout.cards) {
```
tthvo, line 1132: Seems like we can use `Array.map` here.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -1117,6 +1118,34 @@ export class ApiService {
     );
   }
 
+  downloadDashboardLayout(layout: DashboardLayout): void {
+    const serializedLayout = this.getSerializedDashboardLayout(layout);
+    const filename = `cryostat-dashboard-${layout.name}.json`;
+    const file = new File([serializedLayout], filename);
+    const resourceUrl = URL.createObjectURL(file);
```
tthvo, line 1125: There is a `createBlobURL ` in `@app/utils/utils` to handle this. We can reuse it here.

src/app/Dashboard/DashboardSolo.tsx
```diff
@@ -51,7 +51,12 @@ const DashboardSolo: React.FC<DashboardSoloProps> = ({ ..._props }) => {
   const { search } = useLocation();
   const history = useHistory();
 
-  const cardConfigs: CardConfig[] = useSelector((state: RootState) => state.dashboardConfigs.list);
+  const dashboardConfigs = useSelector((state: RootState) => state.dashboardConfigs);
+
+  const cardConfigs: CardConfig[] = React.useMemo(
+    () => dashboardConfigs.layouts[dashboardConfigs.current].cards,
```
tthvo, line 57: Since now, cards are grouped by layouts. The index passing here should also be taken from query parameter. Maybe, `?layout=defaut&cardId=cardname-somehash`. Empty view should also mention this parameter.

This helps in the case where the user bookmark this url and comes back later.

src/test/Dashboard/Dashboard.test.tsx
```diff
@@ -65,11 +65,24 @@ jest.mock('@app/Dashboard/AddCard', () => ({
...
+// Mock the local storage such that the first run config is not shown
+jest.mock('@app/utils/LocalStorage', () => ({
+  getFromLocalStorage: jest.fn(() => {
+    return {
+      _version: '0',
+    };
+  }),
+}));
+
+jest
```
tthvo, line 81: This should work the same and cleaner?

```ts
jest.spyOn(defaultServices.target, 'target').mockReturnValue(of(mockFooTarget));
```

src/app/Dashboard/DashboardLayoutUploadModal.tsx
```diff
@@ -0,0 +1,257 @@
...
+  const { t } = useTranslation();
+  const submitRef = React.useRef<HTMLDivElement>(null); // Use ref to refer to submit trigger div
+  const abortRef = React.useRef<HTMLDivElement>(null); // Use ref to refer to abort trigger div
+
+  const [numOfFiles, setNumOfFiles] = React.useState(0);
+  const [allOks, setAllOks] = React.useState(false);
+  const [uploading, setUploading] = React.useState(false);
+
+  React.useEffect(() => {
+    if (allOks && numOfFiles) {
```
tthvo, line 75: This could be sending state dispatch more than needed. For example, after all succeeding, the user might remove a file from the upload list -> dispatch again.

Probably, this should be handle in the callback in `handleSubmit`. If `allOks`, then

```ts
const latest = dashboardConfigs.layouts[dashboardConfigs.layouts.length - 1];
dispatch(dashboardConfigReplaceLayoutIntent(latest.name));
```

src/app/Dashboard/DashboardLayoutConfig.tsx
```diff
@@ -0,0 +1,398 @@
...
+          </MenuContent>
+        </Menu>
+      </Dropdown>
+    );
+  }, [t, onLayoutSelect, onActionClick, onOpenChange, onToggle, menuGroups, isSelectorOpen, currLayout.name]);
+
+  const toolbarContent = React.useMemo(() => {
+    return (
+      <ToolbarContent style={{ paddingLeft: '24px' }}>
+        <ToolbarGroup>
```
tthvo, line 367: How about this?

```jsx
<ToolbarContent style={{ paddingLeft: '24px' }}>
  <ToolbarItem>{newButton}</ToolbarItem>
  <ToolbarGroup>
    <ToolbarItem spacer={{ default: 'spacerNone' }}>{menuDropdown}</ToolbarItem>
    <ToolbarItem>{renameButton}</ToolbarItem>
  </ToolbarGroup>
  <ToolbarGroup variant="button-group">
    <ToolbarItem>{uploadButton}</ToolbarItem>
    <ToolbarItem>{downloadButton}</ToolbarItem>
    <ToolbarItem>{deleteButton}</ToolbarItem>
  </ToolbarGroup>
</ToolbarContent>
```

This way, the groups are clearly separated?

src/app/Dashboard/DashboardLayoutConfig.tsx
```diff
@@ -0,0 +1,398 @@
...
+  const handleDeleteWarningModalOpen = React.useCallback(
+    (_ev: React.MouseEvent<HTMLButtonElement, MouseEvent>, layout: string) => {
+      setSelectDelete(layout);
+      setIsDeleteWarningModalOpen(true);
+    },
+    [setSelectDelete, setIsDeleteWarningModalOpen]
+  );
+
+  const handleDeleteWarningModalClose = React.useCallback(() => {
+    setIsDeleteWarningModalOpen(false);
```
tthvo, line 137: Reset `selectDelete`? Same for `handleDelete`?

src/app/Dashboard/DashboardLayoutConfig.tsx
```diff
@@ -0,0 +1,398 @@
...
+        {t('DELETE', { ns: 'common' })}
+      </Button>
+    ),
+    [t, handleDeleteButton, currLayout.name]
+  );
+
+  const menuGroups = React.useCallback(
+    (label: string, favoriteGroup: boolean) => {
+      const filter = favoriteGroup ? (l: DashboardLayout) => l.favorite : () => true;
+      return (
```
tthvo, line 291: We probably should hide `Favorite` group if there is none. For example, here we have an empty group above.

Screenshot from 2023-03-19 23-53-19

I think we could extract the filtering out and check children array length?

tthvo (commented):
Looks awesome! Just some refactoring comments :D

src/app/Dashboard/DashboardLayoutConfig.tsx
tthvo: Just to be consistent with target selector :))

src/app/Dashboard/DashboardLayoutConfig.tsx
maxcao13: I think it wouldn't matter since selectDelete is always changed to before dispatch is called, so this would be extra calls that are not needed. Thoughts?

tthvo: ahh yehh no worries. Just that intermediate states can be consistent (like if there is nothing to delete, then selectDelete should be undefined)...

maxcao13: Okay, if that's the guideline then I can change it.

tthvo: not exactly guideline haha but its easier to debug :D

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -557,24 +557,19 @@ const ActiveRecordingsToolbar: React.FunctionComponent<ActiveRecordingsToolbarPr
...
-  );
-
   const handleActionToggle = React.useCallback(() => setActionToggleOpen((old) => !old), [setActionToggleOpen]);
 
   const handleWarningModalClose = React.useCallback(() => {
     setWarningModalOpen(false);
   }, [setWarningModalOpen]);
 
   const handleDeleteButton = React.useCallback(() => {
-    if (deletionDialogsEnabled) {
```
maxcao13, line 572: For some reason the deletionDialogsEnabled memo was not being changed whenever a user clicked on "Do Not ask again", but using the context.settings directly works correctly.

tthvo: `context.settings.deletionDialogsEnabledFor` is not an observable rite? its pulling directly from localstorage and depends only on `context.settings` (never changes). Then, it makes sense that is not updating :D 

The latest changes can `context.settings.deletionDialogsEnabledFor` on -demand so it works ^^

maxcao13: Ah, whoops yeah that makes sense.

tthvo (approved):
Looks awesome! Good job!!

Issue #901, opened 2023-03-03
[Task] Actions for group nodes


tthvo:
> For a follow-up, really not for this PR, it should also be possible to perform certain actions against the group nodes themselves, ie "start a recording on this Deployment". This should be a single GraphQL query:

`environment-nodes-query-1.graphql`
```graphql
query {
    environmentNodes(filter: { name: "JDP" }) {
        descendantTargets {
            doStartRecording(recording: {
                name: "foo",
                template: "Profiling",
                templateType: "TARGET",
                duration: 30
                }) {
                    name
                    state
                }
            }
        }
    }
}
```

> This query should be given a query name, and the name filter parameter should be passed as a separate GraphQL query parameter, but that's the gist of it.

![image](https://user-images.githubusercontent.com/3787464/222603433-1bce1ccc-df76-4aef-b908-bc4689568084.png)

![image](https://user-images.githubusercontent.com/3787464/222603511-ec05d2dd-cc1e-4aed-946c-45d48122c4fd.png)

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-web/issues/891#issuecomment-1452777114_

PR #906, opened 2023-03-13
feat(topology): group node actions

tthvo:
# Welcome to Cryostat! 👋

Fixes: #901 
Depends on https://github.com/cryostatio/cryostat/issues/1403
Depends on https://github.com/cryostatio/cryostat/issues/1410
Depends on https://github.com/cryostatio/cryostat/pull/1411
Depends on https://github.com/cryostatio/cryostat/pull/1417

## Description of the change:

- [x] Add an action menu for group nodes. Currently, only support starting recordings for all targets under that group (all leaves).
- [x] Clean up comments.
- [x] Remove subscription tracker for callback. Action menu is closed when selecting, thus unmounting the items => unsubscribe to the observable. However, there is no `setState` available in callback so its safe not to tracked.  

Note: Fail operations on an individual target (e.g. missing credentials) are silent (even notifications are not shown). Any idea to check errors?

## Motivation for the change:

See #901 

## How to manually test:
2.  Go to Topology View

### DIFF
--- a/src/app/Topology/Actions/NodeActions.tsx
+++ b/src/app/Topology/Actions/NodeActions.tsx
@@ -36,58 +36,94 @@
  * SOFTWARE.
  */
 
-import { NotificationsContext } from '@app/Notifications/Notifications';
+import { Notifications, NotificationsContext } from '@app/Notifications/Notifications';
+import { ActiveRecording } from '@app/Shared/Services/Api.service';
+import { NotificationCategory, NotificationMessage } from '@app/Shared/Services/NotificationChannel.service';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
-import { DropdownItem, DropdownItemProps } from '@patternfly/react-core';
+import { Dropdown, DropdownItem, DropdownItemProps, DropdownProps, DropdownToggle } from '@patternfly/react-core';
+import { css } from '@patternfly/react-styles';
 import { ContextMenuItem as PFContextMenuItem, GraphElement } from '@patternfly/react-topology';
 import * as React from 'react';
 import { useHistory } from 'react-router-dom';
-import { ListElement } from '../Shared/utils';
-import { NodeType, TargetNode } from '../typings';
+import { Observable, Subject, switchMap, map, merge, filter, debounceTime } from 'rxjs';
+import { getConnectUrlFromEvent } from '../Shared/Entity/utils';
+import { getAllLeaves, ListElement } from '../Shared/utils';
+import { EnvironmentNode, NodeType, TargetNode } from '../typings';
 import { ActionUtils } from './utils';
 
-export type NodeActionFunction = (
-  element: GraphElement | ListElement,
-  actionUtils: ActionUtils,
-  track: ReturnType<typeof useSubscriptions>
-) => void;
+export type NodeActionFunction = (element: GraphElement | ListElement, actionUtils: ActionUtils) => void;
 
 export type MenuItemVariant = 'dropdownItem' | 'contextMenuItem';
 
+export type MenuItemComponent = React.FC<DropdownItemProps> | React.FC<React.ComponentProps<typeof PFContextMenuItem>>;
+
 export interface ContextMenuItemProps
-  extends Omit<React.ComponentProps<typeof PFContextMenuItem> & React.ComponentProps<typeof DropdownItem>, 'onClick'> {
+  extends Omit<
+    React.ComponentProps<typeof PFContextMenuItem> & React.ComponentProps<typeof DropdownItem>,
+    'onClick' | 'isDisabled'
+  > {
   onClick?: NodeActionFunction;
   element: GraphElement | ListElement;
   variant: MenuItemVariant;
+  isDisabled?: (element: GraphElement | ListElement, actionUtils: ActionUtils) => Observable<boolean>;
 }
 
-export const ContextMenuItem: React.FC<ContextMenuItemProps> = ({ children, element, onClick, variant, ...props }) => {
+export const ContextMenuItem: React.FC<ContextMenuItemProps> = ({
+  children,
+  element,
+  onClick,
+  variant,
+  isDisabled,
+  ...props
+}) => {
   const history = useHistory();
+  const addSubscription = useSubscriptions();
   const services = React.useContext(ServiceContext);
   const notifications = React.useContext(NotificationsContext);
-  const addSubscription = useSubscriptions();
+  const elementSubjRef = React.useRef(new Subject<GraphElement | ListElement>());
+  const elementSubj = elementSubjRef.current;
 
-  const handleOnclick = React.useCallback(() => {
-    onClick && onClick(element, { history, services, notifications }, addSubscription);
-  }, [onClick, history, services, notifications, addSubscription, element]);
+  const [disabled, setDisabled] = React.useState(false);
 
-  let Comp: React.FC<DropdownItemProps> | React.ComponentProps<typeof PFContextMenuItem>;
+  const handleOnclick = React.useCallback(
+    (e: MouseEvent) => {
+      e.stopPropagation();
+      onClick && onClick(element, { history, services, notifications });
+    },
+    [onClick, history, services, notifications, element]
+  );
+
+  React.useEffect(() => {
+    if (isDisabled) {
+      addSubscription(
+        elementSubj
+          .pipe(switchMap((element) => isDisabled(element, { services, notifications, history })))
+          .subscribe(setDisabled)
+      );
+    }
+  }, [addSubscription, elementSubj, isDisabled, services, notifications, history]);
+
+  React.useEffect(() => {
+    elementSubj.next(element);
+  }, [elementSubj, element]);
+
+  let Component: MenuItemComponent;
   switch (variant) {
     case 'contextMenuItem':
-      Comp = PFContextMenuItem;
+      Component = PFContextMenuItem;
       break;
     case 'dropdownItem':
-      Comp = DropdownItem;
+      Component = DropdownItem;
       break;
     default:
-      throw new Error('unknown variant');
+      throw new Error(`Unknown variant: ${variant}`);
   }
 
   return (
-    <Comp onClick={handleOnclick} {...props}>
+    <Component {...props} onClick={handleOnclick} isDisabled={disabled}>
       {children}
-    </Comp>
+    </Component>
   );
 };
 
@@ -97,21 +133,57 @@ export type NodeActionKey =
   | 'CREATE_RECORDINGS'
   | 'CREATE_RULES'
   | 'DELETE_TARGET'
+  | 'GROUP_START_RECORDING'
+  | 'GROUP_STOP_RECORDING'
+  | 'GROUP_DELETE_RECORDING'
+  | 'GROUP_ARCHIVE_RECORDING'
   | '';
 
+export const QUICK_RECORDING_NAME = 'cryostat_topology_action';
+
+export const QUICK_RECORDING_LABEL_KEY = 'cryostat.io.topology-group';
+
+const isQuickRecording = (recording: ActiveRecording) => {
+  return recording.name === QUICK_RECORDING_NAME;
+};
+
+const isQuickRecordingExist = (group: EnvironmentNode, { services }: ActionUtils) => {
+  const svcUrls = new Set(getAllLeaves(group).map((tn) => tn.target.connectUrl));
+  const filterFn = (e: NotificationMessage) => {
+    const targetId = getConnectUrlFromEvent(e);
+    const recording = e.message.recording;
+    return targetId !== undefined && svcUrls.has(targetId) && isQuickRecording(recording);
+  };
+
+  return merge(
+    services.api.groupHasRecording(group, QUICK_RECORDING_NAME),
+    services.notificationChannel.messages(NotificationCategory.ActiveRecordingCreated).pipe(
+      filter(filterFn),
+      map((_) => true)
+    ),
+    services.notificationChannel.messages(NotificationCategory.ActiveRecordingDeleted).pipe(
+      filter(filterFn),
+      debounceTime(500),
+      map((_) => services.api.groupHasRecording(group, QUICK_RECORDING_NAME))
+    )
+  );
+};
+
 export interface NodeAction {
-  key: NodeActionKey;
-  action?: NodeActionFunction;
-  title?: React.ReactNode;
-  isSeparator?: boolean;
-  includeList?: NodeType[]; // Empty means all
-  blockList?: NodeType[]; // Empty means none
+  readonly key: NodeActionKey;
+  readonly isGroup?: boolean;
+  readonly action?: NodeActionFunction;
+  readonly title?: React.ReactNode;
+  readonly isSeparator?: boolean;
+  readonly isDisabled?: (element: GraphElement | ListElement, actionUtils: ActionUtils) => Observable<boolean>;
+  readonly includeList?: NodeType[]; // Empty means all
+  readonly blockList?: NodeType[]; // Empty means none
 }
 
 export const nodeActions: NodeAction[] = [
   {
     key: 'VIEW_DASHBOARD',
-    action: (element, { history, services }, _) => {
+    action: (element, { history, services }) => {
       const targetNode: TargetNode = element.getData();
 
       services.target.setTarget(targetNode.target);
@@ -121,7 +193,7 @@ export const nodeActions: NodeAction[] = [
   },
   {
     key: 'VIEW_RECORDINGS',
-    action: (element, { history, services }, _) => {
+    action: (element, { history, services }) => {
       const targetNode: TargetNode = element.getData();
 
       services.target.setTarget(targetNode.target);
@@ -132,7 +204,7 @@ export const nodeActions: NodeAction[] = [
   { key: '', isSeparator: true },
   {
     key: 'CREATE_RECORDINGS',
-    action: (element, { history, services }, _) => {
+    action: (element, { history, services }) => {
       const targetNode: TargetNode = element.getData();
 
       services.target.setTarget(targetNode.target);
@@ -142,7 +214,7 @@ export const nodeActions: NodeAction[] = [
   },
   {
     key: 'CREATE_RULES',
-    action: (element, { history, services }, _) => {
+    action: (element, { history, services }) => {
       const targetNode: TargetNode = element.getData();
 
       services.target.setTarget(targetNode.target);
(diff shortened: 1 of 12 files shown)


tthvo:
## List View
Screenshot from 2023-03-13 16-18-56


## GraphView Detail Panel

Screenshot from 2023-03-13 16-19-56


## GraphView Context Menu

Screenshot from 2023-03-13 16-20-07

src/app/Topology/Actions/NodeActions.tsx
```diff
@@ -139,11 +141,74 @@ export const nodeActions: NodeAction[] = [
...
+        .graphql(
+          `
+            query StartRecordingForGroup($filter: EnvironmentNodeFilterInput) {
+              environmentNodes(filter: $filter) {
+                  descendantTargets {
+                      doStartRecording(recording: {
+                          name: "${recordingName}",
+                          template: "Profiling",
+                          templateType: "TARGET",
+                          duration: 30
```
andrewazores, line 165: In the absence of any UI widget for specifying a template and duration, I would lean more toward using the continuous template since the potential runtime impact (recording overhead) is lower, and probably with an unlimited duration or perhaps just something fairly long, like 10 or 15 minutes. I think if there were also an action for stopping recordings on the group then it would be fine to let the duration be unlimited.

tthvo: > I think if there were also an action for stopping recordings on the group then it would be fine to let the duration be unlimited.

Sure!! But about this, `Stop Recordings` here means stopping every recordings? Wondering how the users would know which recordings will be stopped? And how do we find the recording name?

src/app/Topology/Actions/NodeActions.tsx
```diff
@@ -139,11 +141,74 @@ export const nodeActions: NodeAction[] = [
...
-      track(services.api.deleteTarget(targetNode.target).subscribe(() => undefined));
+      services.api.deleteTarget(targetNode.target).subscribe(() => undefined);
     },
     title: 'Delete Target',
     includeList: [NodeType.CUSTOM_TARGET],
   },
+  {
+    action: (element, { services }) => {
+      const group: EnvironmentNode = element.getData();
+      const recordingName = `${group.name.replace(/[\s+-]/g, '_')}_${Date.now()}`;
```
andrewazores, line 154: Using the timestamp as part of the name is nice to help the user avoid name collisions but it also makes it harder/costlier for us to undo that operation, since now we need to query for recordings that have names that look like they match this pattern. I think using a fixed constant for the name is okay so long as the name is unlikely to be something the user has already used on their own, ex. `cryostat_topology_action`. `cryostat_group_${group.name.replaced}` might be okay as well but I wonder if there are cases or platforms where the group name may change without that also meaning that all the children to the leaf nodes have been redeployed, in which case a group's rename would also break our ability to undo group operations.

src/app/Topology/Actions/NodeActions.tsx
andrewazores: See other comment that I was writing concurrently

tthvo: > And how do we find the recording name?

I guess we can use well known name but that would allow creating once unless deleting the recordings later?

tthvo: Ahh nicee! Okayyy! I will change the query to this.

andrewazores: So I think there would be:

- a Start Recording group action that uses a well-known name and sets the `restart` setting
- a Stop Recording group action that stops any recording with that name, ignoring any failures due to recording already being stopped
- a Delete/Remove Recording group action that deletes active recordings with that name regardless of running state

There could also hypothetically be a query that checks for the aggregate count of number of recordings with the well-known name that are currently in the Running state, and this can control whether the Stop Recording action is enabled. Might be overkill. Likewise for Remove Recording, checking the aggregate count of recordings with that name in any state.

andrewazores: Oh, another idea, you can set that `${group.name.replace()}` and the timestamp and other details in the metadata when you perform the recording start mutation. The `metadata` should be provided as a (JSON) object with a `labels` field that is just a `Map<String, String>`. I think it goes like this:

```graphql
query {
    environmentNodes(filter: { name: "JDP" }) {
        descendantTargets {
            doStartRecording(recording: {
                name: "foo",
                template: "Profiling",
                templateType: "TARGET",
                duration: 30,
                metadata: {
                    labels: {
                        "cryostat.io/topology-group": "my-group-name"
                    }
                }
                }) {
                    name
                    state
            }
        }
    }
}
```

src/app/Topology/Actions/NodeActions.tsx
```diff
@@ -139,11 +141,74 @@ export const nodeActions: NodeAction[] = [
...
+      services.api
+        .graphql(
+          `
+            query StartRecordingForGroup($filter: EnvironmentNodeFilterInput) {
+              environmentNodes(filter: $filter) {
+                  descendantTargets {
+                      doStartRecording(recording: {
+                          name: "${recordingName}",
```
andrewazores, line 162: I *think* this `recordingName` can also be passed to the server as a GraphQL parameter the same way that `$filter` is being done? If so then it's just a little more optimized because the server only needs to parse this query one time and then it knows how to re-execute it with different values for the variables. Doing client-side string interpolation to build the query means it busts the parsed query cache on the server side.

tthvo:
Ahh I was trying separate some changes for #907 but seems to lose some changes for this PR...

src/app/Topology/Actions/NodeActions.tsx
tthvo: Hey actually, I was trying to get this query to work but seems to throw invalid syntax or casting error

```graphql
query StartRecordingForGroup($filter: EnvironmentNodeFilterInput, $name: String!) {
              environmentNodes(filter: $filter) {
                  descendantTargets {
                      doStartRecording(recording: {
                          name: $name,
                          template: "Continuous",
                          templateType: "TARGET",
                          duration: 0,
                          restart: true,
                          metadata: {
                            labels: {
                                "cryostat.io/topology-group": "my-group-name"
                            }
                          },
                          }) {
                              name
                              state
                      }
                  }
              }
          }
```

tthvo: ### With Labels
```
Invalid Syntax : offending token '\"cryostat.io/topology-group\"' at line 1 column 283
```

### With restart

```
Exception while fetching data (/environmentNodes[0]/descendantTargets[0]/doStartRecording) : class java.lang.Boolean cannot be cast to class java.lang.String (java.lang.Boolean and java.lang.String are in module java.base of loader 'bootstrap')
```

tthvo: > There could also hypothetically be a query that checks for the aggregate count of number of recordings with the well-known name that are currently in the Running state, and this can control whether the Stop Recording action is enabled. Might be overkill. 

Actually very nice idea but right now, the way its set up is not very neat to support this. We will need to go with `regardless of state`.

andrewazores: > ### With Metadata/Labels
> 
> ```
> Invalid Syntax : offending token '\"cryostat.io/topology-group\"' at line 1 column 283
> ```
> 
> ### With restart option
> 
> ```
> Exception while fetching data (/environmentNodes[0]/descendantTargets[0]/doStartRecording) : class java.lang.Boolean cannot be cast to class java.lang.String (java.lang.Boolean and java.lang.String are in module java.base of loader 'bootstrap')
> ```

It looks like I probably just gave a bad example of a label key that isn't acceptable, and I guess the "restart" parameter needs to be supplied as a string value, so just wrapping the `true` boolean literal in double quotes (`"true"`) should fix the casting issue I think.

tthvo: Unfortunately, there is another error when wrapping that boolean in string:

```
"Validation error (WrongType@[environmentNodes/descendantTargets/doStartRecording]) : argument 'recording.restart' with value 'StringValue{value='true'}' is not a valid 'Boolean' - Expected AST type 'BooleanValue' but was 'StringValue'."
```

tthvo: > It looks like I probably just gave a bad example of a label key that isn't acceptable.

Seems to work fine when creating recordings. Is there some constraints on the GraphQL side?

Screenshot from 2023-03-13 19-37-14

andrewazores: What if you also pass the label value in as a query variable? I seem to recall there are some limitations around this on the GraphQL side but I'm not sure if there is a way to escape it here.

For the `restart` thing, it looks like the Mutator code on the backend is just wrongly assuming that it will get a String back from the context object, but GraphQL has already converted iti to a Boolean so the cast fails. The GraphQL schema does specify it should be a Boolean parameter however, so the request validation fails when you wrap it as a String. Should be an easy fix on the backend.

tthvo: Still cant solve this labels but I did some digging and found this PR. Could this be the issue?

https://github.com/cryostatio/cryostat/pull/971#issuecomment-1157956641

tthvo: If it is really, I am thinking we can go with:

```ts
metadata: {
  labels: {
    topology_group: "${group.name.replace(/[\s+-]/g, '_')}"
  }
},
```

tthvo:
Other than the issue with label format, I think its ready to be reviewed. Actions should work now.

andrewazores:
Hmm, yea, looks like the label format thing is just currently broken because the label key has to be something that GraphQL accepts as an object literal key, so basically something that is a valid JS identifier. That should be updated to match the syntax in the examples that Janelle provided which the `doPutMetadata` mutator accepts. ie

```graphql
query {
    targetNodes(filter: { name: "service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi" }) {
        recordings {
            active(filter: { name: "myActiveRecording" }) {
                doPutMetadata(metadata: { labels: [{key:"app",value:"cryostat"}, {key:"template.name",value:"Profiling"},{key:"template.type",value:"TARGET"}] }) {
                    name
                    metadata {
                        labels
                    }
                }
            }
        }
    }
}

{
    "data": {
        "targetNodes": [
            {
                "recordings": {
                    "active": [
                        {
                            "doPutMetadata": {
                                "metadata": {
                                    "labels": {
                                        "app": "cryostat",
                                        "template.name": "Continuous",
                                        "template.type": "TARGET"
                                    }
                                },
                                "name": "myActiveRecording"
                            }
                        }
                    ]
                }
            }
        ]
    }
}
```

`metadata: { labels: [{key:"app",value:"cryostat"}, {key:"template.name",value:"Profiling"},{key:"template.type",value:"TARGET"}] }` is the key thing.

https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/net/web/http/api/v2/graph/StartRecordingOnTargetMutator.java#L163

https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/net/web/http/api/v2/graph/PutActiveRecordingMetadataMutator.java#L121

tthvo:
I actually run into another problem with nodes of the same name. For example, starting a recording on `9988` actually creates one for `9097`.

Screenshot from 2023-03-14 01-09-17

andrewazores:
Ah... shoot. Let me think about this for a while. In this version of the discovery API there are no unique IDs for individual nodes in the tree, so when names are non-unique we're a bit stuck.

Maybe on the backend we can start at the leaf node (as usual), hash it, and then patch the node to add that hash as a property to itself. Then recurse up to the parent and apply the same thing. It's not particularly efficient on read or on write but I don't think it should be too bad, and it gives us a uniqueness property that can be referenced here if we also expose it in the GraphQL schema and add it to the filter inputs.

andrewazores:
Or really, on the backend side in Java, we already have the `Object#hashCode` method that does basically exactly this that we can reuse.

tthvo:
I am trying to have some visual indicators about failed operations on individual target. However, not sure if we can pull any meaningful message from the graphql response. Here is what i see. Any guide on how to get to show something like `Could not start recording on ${group.name}/${target-alias}`?

Screenshot from 2023-03-15 16-10-58


```json
{
   "errors":[
      {
         "message":"Exception while fetching data (/environmentNodes[0]/descendantTargets[0]/doStartRecording) : org.openjdk.jmc.rjmx.ConnectionException caused by java.lang.SecurityException: Authentication failed! Invalid username or password",
         "locations":[
            {
               "line":1,
               "column":142
            }
         ],
         "path":[
            "environmentNodes",
            0,
            "descendantTargets",
            0,
            "doStartRecording"
         ],
         "extensions":{
            "classification":"DataFetchingException"
         }
      },
      {
         "message":"Exception while fetching data (/environmentNodes[0]/descendantTargets[2]/doStartRecording) : org.openjdk.jmc.rjmx.ConnectionException caused by java.lang.SecurityException: Authentication failed! Invalid username or password",
         "locations":[
            {
               "line":1,
               "column":142
            }
         ],
         "path":[
            "environmentNodes",
            0,
            "descendantTargets",
            2,
            "doStartRecording"
         ],
         "extensions":{
            "classification":"DataFetchingException"
         }
      },
      {
         "message":"Exception while fetching data (/environmentNodes[0]/descendantTargets[3]/doStartRecording) : org.openjdk.jmc.rjmx.ConnectionException caused by java.io.IOException: Failed to retrieve RMIServer stub: javax.naming.CommunicationException [Root exception is java.rmi.ConnectIOException: non-JRMP server at remote endpoint]",
         "locations":[
            {
               "line":1,
               "column":142
            }
         ],
         "path":[
            "environmentNodes",
            0,
            "descendantTargets",
            3,
            "doStartRecording"
         ],
         "extensions":{
            "classification":"DataFetchingException"
         }
      }
   ],
   "data":null
}
```

andrewazores:
I don't think what you're asking for can be neatly done right now in a generic way, though it's a very good idea. The exception stack trace in the message is not really machine processable. We can make the backend throw exceptions in a very well structured and consistent way but that may be error prone. The level that the exception is being thrown at doesn't know anything about discovery nodes or the GraphQL request path so the GraphQL fetchers/mutators (or their base class) would need to catch exceptions like this and wrap them into something that is processable. Only the `message` field of that response is actually generated by Cryostat code directly though, the rest of that response format is GraphQL.

Does the frontend component that fires this request have that contextual information about what node it's being invoked on and what operation it's performing? If so then this can be used to generate a message about a partial request failure if the `errors` field of the response is a nonempty array. Rather than having an individualized message for each failure with specific details, I think it's probably much easier to just emit one warning saying "part of that operation failed, part of it succeeded", and that's better than partial silent failures at least.

andrewazores:
Oh, it looks like your last commit did something a lot like what I was describing while I was writing it :-)

tthvo:
Oh yess i just checked the response body and show all the error that it throws :)) Seems like the only probable way now^^ Do you want to file an issue for later?

andrewazores:
> Wondering if from the current message, we can extract anything that identify the target being operated on? Seems to have some indexing but not sure that can be used?

Yea I don't think it really can be, at least not as it is from the client side. The indexing is just based on the list returned by each fetcher stage of the query at runtime. The order within the list may or not be stable so the client shouldn't really depend on it. And, for a query like this one, the client doesn't know what the contents of the list were anyway. Maybe if the query also included other fields like the node name then the partial success response could be used and the error indices used to look up the names of the failed nodes? Actually, if that works, it's probably the best approach...

tthvo:
> Maybe if the query also included other fields like the node name then the partial success response could be used and the error indices used to look up the names of the failed nodes? Actually, if that works, it's probably the best approach...

Sounds good! Just sorry how does this work? Changes to graphql query front-end sending?

src/app/Topology/Actions/NodeActions.tsx
```diff
@@ -153,11 +160,175 @@ export const nodeActions: NodeAction[] = [
...
+    title: 'Start Recordings',
+    isGroup: true,
+    action: (element, { services, notifications }) => {
+      const group: EnvironmentNode = element.getData();
+      services.api
+        .graphql<any>(
+          `
+          query StartRecordingForGroup($filter: EnvironmentNodeFilterInput, $name: String!) {
+            environmentNodes(filter: $filter) {
+              descendantTargets {
```
andrewazores, line 181: Just add the `name` field here, ie.

```
query StartRecordingForGroup($filter: EnvironmentNodeFilterInput, $name: String!) {
  environmentNodes(filter: $filter) {
    *name*
    descendantTargets {
    }
  }
}
```

This way, the responses should contain the name of the target nodes. I *think* it will even include them in the case that the `doStartRecording` mutator on that node fails. The array index of the `doStartRecording` failure should give you the lookup of which array index to look for the corresponding `name`, if so.

tthvo: This is the error returns. I guess maybe its not working as expected?

```json
{
   "errors":[
      {
         "message":"Exception while fetching data (/environmentNodes[0]/descendantTargets[0]/doStartRecording) : org.openjdk.jmc.rjmx.ConnectionException caused by java.lang.SecurityException: Authentication failed! Invalid username or password",
         "locations":[
            {
               "line":1,
               "column":156
            }
         ],
         "path":[
            "environmentNodes",
            0,
            "descendantTargets",
            0,
            "doStartRecording"
         ],
         "extensions":{
            "classification":"DataFetchingException"
         }
      },
      {
         "message":"Exception while fetching data (/environmentNodes[0]/descendantTargets[2]/doStartRecording) : org.openjdk.jmc.rjmx.ConnectionException caused by java.lang.SecurityException: Authentication failed! Invalid username or password",
         "locations":[
            {
               "line":1,
               "column":156
            }
         ],
         "path":[
            "environmentNodes",
            0,
            "descendantTargets",
            2,
            "doStartRecording"
         ],
         "extensions":{
            "classification":"DataFetchingException"
         }
      },
      {
         "message":"Exception while fetching data (/environmentNodes[0]/descendantTargets[3]/doStartRecording) : org.openjdk.jmc.rjmx.ConnectionException caused by java.io.IOException: Failed to retrieve RMIServer stub: javax.naming.CommunicationException [Root exception is java.rmi.ConnectIOException: non-JRMP server at remote endpoint]",
         "locations":[
            {
               "line":1,
               "column":156
            }
         ],
         "path":[
            "environmentNodes",
            0,
            "descendantTargets",
            3,
            "doStartRecording"
         ],
         "extensions":{
            "classification":"DataFetchingException"
         }
      }
   ],
   "data":null
}
```

andrewazores: Hmm.

https://stackoverflow.com/questions/52778096/how-to-return-both-error-and-data-in-a-graphql-resolver

https://stackoverflow.com/questions/52778096/how-to-return-both-error-and-data-in-a-graphql-resolver#comment108893874_52780113

Might be this. Needs some backend investigation but in theory this should be doable.

andrewazores: I guess you can proof-of-concept this by just modifying the `types.graphqls`/`schema.graphqls` and removing some `!` non-nullness signifiers.

tthvo: Uhh niceee I was able to get this when removing 

```diff
diff --git a/src/main/resources/types.graphqls b/src/main/resources/types.graphqls
index 6fe95865..2325fea1 100644
--- a/src/main/resources/types.graphqls
+++ b/src/main/resources/types.graphqls
@@ -75,7 +75,7 @@ type TargetNode implements Node {
     nodeType: NodeType!
     labels: Object!
 
-    doStartRecording(recording: RecordingSettings!): ActiveRecording!
+    doStartRecording(recording: RecordingSettings!): ActiveRecording
     doSnapshot: ActiveRecording!
 }
```

Screenshot from 2023-03-15 18-52-10

tthvo: I will probably do the same for `doStop` and `doDelete`.

tthvo:
Now, we can see what target node failed instead of silencing errors.

Screenshot from 2023-03-15 20-47-14

andrewazores:
Seems to work great. Two more ideas:

1. A node action for archiving recordings
2. Some indication at the Topology level of active recordings' states. An indicator on the target node itself (similar to the warning state badge used now when the node isn't connectable/has no JVM ID) but also perhaps some further details in the sidebar panel just to summarize, ex. `Active recordings: 1 running, 1 stopped`

tthvo:
> A node action for archiving recordings

Is this archiving that special active recordings we create? Or all active recordings? I guess the first case?

andrewazores:
Conceptually, I think archiving any running active recording does basically the same thing. If there are multiple running recordings then JFR basically merges the configurations and records all the configured events, and so recording files taken from any of the running recordings will actually contain the same data.

So on the one hand, always picking the recording that we started from the topology view action works fine.

On the other hand, this means that recordings started in any other way won't be archived by this new action, which could be confusing. I might have an Automated Rule for example, which always starts a recording when a target appears, and either doesn't have a scheduled archiving configuration or the frequency is slow and I want an archival copy taken right now.

Maybe we should have a GraphQL mutator that does archival directly on a target node, and under the hood it just chooses one of the active recordings to archive.

Otherwise, right now on the client side you can make a query that performs a snapshot and then immediately archives it.

tthvo:
Actually, now that we mention the archive. Seems like `stop and delete` also falls into this issue. Maybe stop/delete recordings will also imply stop/delete all?

andrewazores:
Yea, this gets a little complicated.

For deletion I would rather err on the side of caution and only delete the recording that was directly started by the action taken on this same view. Deleting other data that the user didn't expect is really bad.

For stopping I think it's similar, just stop the one that was started from here. If the user created the recording manually then they might have had a reason. If it was started by Automated Rule they can stop it by disabling the rule. If it was started by some other means outside of Cryostat then again they may have some reason for that and we should leave it be.

tthvo:
Ohh just saw ur comments now. Okayy I will work on the archive action then. Later, we might need explain the meaning behind these actions somewhere (i.e. blog posts, quickstarts, etc)...

tthvo:
I will try to see if I can refactor so that action menu can disable certain action (i.e. delete, stop, archive) if that special recording does not exist.

tthvo:
Latest commit includes changes:
- Check if the special recording is present and disable actions if needed (no recording available).
- Reword action display so that it does not imply scope of the actions is outside of the special recording.

Seems to work fine. Let me know what you think :D

tthvo:
Updated to use query parameter for `Start recording` now. I also removed debounceTime for `NotificationCategory.ActiveRecordingCreated` (it just returns true anyways, React can know how to prevent re-rendering if state is the same).

andrewazores (approved):

PR #907, opened 2023-03-13
fix(topology): custom target submission should also send credentials

tthvo:
P# Welcome to Cryostat! 👋

Related to #891 
Depends on https://github.com/cryostatio/cryostat/issues/1404

## Description of the change:

Previously, the custom target creation is not sending credentials when submitting. Testing is working fine tho. Also, added some indications for error similar to testing.


maxcao13:
Currently it seems creating a target even thought there are credentials does not work. I believe it's because previously the web-client sends a TargetPost request without credentials and dryRun=false. 

This would call `getJvmId` and try and use the cache but fail and then call `computeJvmId`. But since there are no credentials present in the original Http request, this call 
```bash
        CompletableFuture<String> future =
                this.targetConnectionManager.executeConnectedTaskAsync(
                        new ConnectionDescriptor(
                                targetId,
                                credentials.isPresent()
                                        ? credentials.get()
                                        : credentialsManager.getCredentialsByTargetId(targetId)),
                        JFRConnection::getJvmId);
```
would result in using the credentials from credentialsManager. HOWEVER, this fails because getCredentialsByTargetId iterates the list of platformClients and targets within them to find a matching credential. But obviously that target does not exist yet since we are creating it and hence it fails.

So it would seem like we would have to somehow use credentials in the original request. Like here in this pr.

tthvo:
Seems like it...

```bash
$ http --form POST localhost:8181/api/v2/targets?dryrun=false \
alias="self" \
connectUrl='service:jmx:rmi:///jndi/rmi://localhost:9091/jmxrmi'  \
username="smoketest" \
password="smoketest" \
--auth user:pass \
--auth-type basic

HTTP/1.1 427 Client Error (427)
content-encoding: gzip
content-length: 126
vary: origin

{
    "data": {
        "reason": "JMX Authentication Failure"
    },
    "meta": {
        "status": "Client Error (427)",
        "type": "text/plain"
    }
}
```

andrewazores:
Looks like a backend bug needs to be filed then?

andrewazores (approved):

src/app/Topology/Actions/CreateTarget.tsx
```diff
@@ -187,20 +189,32 @@ export const CreateTarget: React.FC<CreateTargetProps> = ({ prefilled, ..._props
...
+          } else {
+            setValidation({
+              option: option,
+              errorMessage: body['data']['reason'],
+            });
           }
         })
     );
-  }, [setLoading, addSubscription, context.api, connectUrl, alias, history]);
+  }, [setLoading, addSubscription, context.api, connectUrl, alias, history, credentials]);
```
maxcao13, line 217: Needs setValidation?

tthvo: Hmm eslint misses this again. Thanks!

tthvo:
Sorry latest commit has some patches to add missing hook deps, reuse `createTarget` for testing and removing TargetCreationModal since we don't use it anymore.

andrewazores (approved):

Issue #912, opened 2023-03-20
[Task] Reuse target details component in matchExpression views


andrewazores:
The right sidebar panel component here:

![image](https://user-images.githubusercontent.com/3787464/226461142-9f400187-b824-41a1-b1b0-f35c42739564.png)

represents the same information as the right side split component here:

![image](https://user-images.githubusercontent.com/3787464/226461223-05ba4672-f7dc-4d44-8f40-865fbf5de246.png)

This second old-style serialized JSON component is currently present on the `Automated Rules > Create` and `Security > Store JMX Credentials > Add` views. The new details component is much more pleasing to look at and probably makes it easier for the user to differentiate the different fields they're looking at. With some section renaming or tooltips, the field names required to build a matchExpression can be determined from the new details component pretty easily.

tthvo:
> With some section renaming or tooltips, the field names required to build a matchExpression can be determined from the new details component pretty easily.

Are we putting the field name into the tooltip? 

Also, do we still need to keep the codeblock for hints? This is what I have right now:

![Screenshot from 2023-03-20 04-54-10](https://user-images.githubusercontent.com/68053619/226463316-c053ef8b-0fc9-4117-b73a-fea32161f7ca.png)

andrewazores:
> Are we putting the field name into the tooltip?

I think so, maybe something like the JSON path, ex. `target.connectUrl` as the tooltip on "Connection URL". Or even just replace the "connection URL" section title with `target.connectUrl`.

> Also, do we still need to keep the codeblock for hints? This is what I have right now:

Not necessarily, though I think it's useful to provide a hint of some form somewhere. After #913 it's hopefully self-evident and easy enough that the hint block can be removed.

tthvo:
Make sense thanks! I will add them in in the popover content.

Issue #900, opened 2023-02-28
[Task] Remove explicit mention of JMX for Stored Credentials


andrewazores:
On that note though, the frontend needs to be updated to remove specific references to JMX, since these stored credentials will also be used for HTTP-based Agent connections.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/1377#issuecomment-1448648881_

tthvo:
Seems like this issue can be solved along with #914. What should I be writing here??

PR #914, opened 2023-03-20
feat(matchExpr): match expression visualizer

tthvo:
# Welcome to Cryostat! 👋

Fixes: #911 
Fixes: #912 
Fixes: #900 

## Description of the change:

- [x] Match Expression Visualizer with mini topology
  - [x] Graph View
  - [x] List view (if # of targets are too large). 
  - [x] Fix height constraints -> need to fill viewport instead of overflowing.
  - [x] Update target detail panel to reflect fieldName.

**Note: No tests yet tho. Focusing on flushing out features for now.**

## Motivation for the change:

See https://github.com/cryostatio/cryostat-web/issues/376#issuecomment-1448689272

## How to manually test:
2. Go to Rule and Credential Forms.

### Credential Modal

Screenshot from 2023-03-20 23-30-29

### Create Rule Form

Screenshot from 2023-03-21 01-25-39

### Match Expression Hint

Screenshot from 2023-03-21 01-27-44


tthvo:
Notice how the field path is represented as breadcrumbs. I took this design from OpenShift console. Let me know what you think :D

Screenshot from 2023-03-20 23-36-40

tthvo:
Read for review now :D 

Just a small note for testing with observables that includes `debounceTime`: just need to use `waitFor` to check if some function (e.g. api requests, etc) is called.

andrewazores:
1. Go to Topology
2. Select a target with the warning exclamation
3. Click "Add Credentials"
4. Select a target with the warning exclamation
5. Click "Add Credentials" ...

Can that action be removed from the sidebar panel component when it's in this visualization context?

tthvo:
Yeh sure, of course!

I left it in case the user want to add credentials in Create Rule Form. For example. if the target is matched but could not fetch templates, adding credentials there can help refreshing on the go. What do you think?

tthvo:
I added options to allow hide action buttons. This is applied for credential modal but still kept for Create Rule Form.

Screenshot from 2023-03-21 15-55-15

tthvo:
Views appear not responding to target discovery events. Investigating...

tthvo:
Updated to add some effect to update the views and fixing some array reduce error (didnt know it throws error if the array is emty ^^).

andrewazores:
image

The Actions dropdown within the modal is still present and clickable, but I guess because there are no possible actions here, no menu drops down. Can the Actions menu button just be removed if its action list is empty?

tthvo:
> The Actions dropdown within the modal is still present and clickable, but I guess because there are no possible actions here, no menu drops down. Can the Actions menu button just be removed if its action list is empty?

Hmm was that the dropdown for sidecar panel? I think it should open up some actions. Did it not?

andrewazores:
Also the text at the top of that modal needs updating:

> Creates stored credentials for target JVMs according to various properties. If a Target JVM requires JMX authentication, Cryostat will use stored credentials when attempting to open JMX connections to the target.

Should be more generic and just talk about opening connections, not specifically JMX ones.

tthvo:
Ohhh interesting...there is a bug somewhere I think. It should show up. Let me check!

tthvo:
Never mind, the menu was actually showing but it was mounted in the back of the modal haha. Fixed now :D

tthvo:
Updated texts and remove actions in empty state: asynchronously update input field will cause the cursor jump to the end -> cant use hook -> actions to reset match expression can't be used.

maxcao13:
image
Should the message be different when the expression is valid?

src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx
```diff
@@ -0,0 +1,227 @@
...
+            </Button>
+          </Popover>
+        }
+        isRequired
+        fieldId="match-expression"
+        helperText={
+          matchExpressionValid === ValidatedOptions.warning
+            ? `Warning: Match expression matches no targets.`
+            : `
+        Enter a match expression. This is a Java-like code snippet that is evaluated against each target
```
maxcao13, line 200: Would it make sense to change Java-like to JavaScript like? Since the match expression evaluation is actually evaluating js code.

src/app/Rules/CreateRule.tsx
```diff
@@ -232,60 +234,70 @@ const Comp: React.FC = () => {
...
+          throw new Error('Invalid match expression');
+        });
+        validation = matches.length ? ValidatedOptions.success : ValidatedOptions.warning;
+      } catch (err) {
+        validation = ValidatedOptions.error;
+      }
+    }
+    setMatchExpressionValid(validation);
+    matchedTargets.next(matches);
+    console.log(matchExpression);
```
maxcao13, line 299: remove console log?

src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx
```diff
@@ -0,0 +1,227 @@
...
+        showClose={!inProgress}
+        className="add-credential-modal"
+        onClose={onDismiss}
+        title="Store Credentials"
+        description="Create stored credentials for target JVMs. Cryostat will use these credentials to connect to Cryostat agents or target JVMs over JMX (if required)."
+      >
+        <Grid hasGutter style={{ height: '100%' }}>
+          <GridItem xl={4}>
+            <Card isFullHeight isFlat>
+              <CardBody style={{ overflow: 'auto' }}>
```
maxcao13, line 95: seems like there are a lot of CardBody's with this style. Maybe better to add a class like "cryo-m-overflow-auto" and then a css rule like

.pf-c-card.cryo-m-overflow-auto {
    overflow: auto
}

src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx
andrewazores: The fact that it's JavaScript is sort of an internal implementation detail that might change in the (distant) future. This should really be a much more limited scripting language that can't do all of what the JS engine is capable of doing, which is why I had to lock down the JS parser so much on the backend when it validates expressions before evaluating them. I would rather have done it so that this uses an actual subset of Java syntax directly rather than running it through a JS engine, but it was a pragmatic choice to get something running in reasonable fashion, and not something I want to say we're committing to.

tthvo:
> If it's not hard to do, keep the positions of the nodes when switching between graph view and list view?

Should be okay to do I think. Just that this might cause another huge chunk of data to be stored if # of targets is large. Especially, topology view also saves positions too.

> Should the message be different when the expression is valid?

https://www.patternfly.org/v4/components/form/design-guidelines

No guideline for valid message... I think the green color + check icon should be enough to indicate that.

src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx
tthvo: Makes sense! I will keep the Java for now then :D

tthvo:
Updated to to save graph positions and clean up css, console.logs :))

src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx
andrewazores: Another possible avenue, rather than implementing a subset of Java syntax as an embedded scripting language, is to use JShell or something similar as a replacement for the current nashorn JS engine. If there is a way to similarly lock down the JShell parser so that it rejects expressions with dangerous constructs then this would mean the expression syntax really is Java, instead of Java-like.

https://github.com/cryostatio/cryostat/issues/990

maxcao13 (approved):
Looks great!

Issue #593, opened 2022-10-28
[Task] Add "test connection" button to JMX Credentials creation form


andrewazores:

tthvo:
I think maybe its also nice to have, for example, an expandable section that allows the user to enter credentials for this specific target on the target creation form. This way, they don't have to define a custom target and then go to Security tab.

andrewazores:
That's a good idea.

PR #920, opened 2023-03-22
feat(cred): credential test table

tthvo:
# Welcome to Cryostat! 👋

Fixes: #592 
Fixes: #593 
Related to #914 
Depends on https://github.com/cryostatio/cryostat/issues/1431

## Description of the change:

- [x] Added a table for testing credentials against matching targets.
- [x] Hide/Fade targets when match expression is empty.
- [x] Listview should be scrollable (allow sticky radio button)
- [x] Remove explicit mentioning of JMX Credentials (mostly internals that are not visible to the users).

## Motivation for the change:

See #591 

## How to manually test:
2. Go to Security -> Add Store Credentials -> Test tab.

## Screenshots

**EDIT:** See https://github.com/cryostatio/cryostat-web/pull/920#issuecomment-1483511361

Screenshot from 2023-03-23 21-17-23


maxcao13:
The UI looks great! Just a few questions:

image

Using a "true" matchExpr and a smoketest smoketest credentials:

Should there be an indication that a target does not need credentials? Right now the test will say that it is "valid" it doesn't mean the credentials are valid for that target.

tthvo:
> Should there be an indication that a target does not need credentials? Right now the test will say that it is "valid" it doesn't mean the credentials are valid for that target.

Try hover over the valid label. It should say some warning :)) Any better than that?

maxcao13:
The only thing I can think of right now, is to send a graphql request to each of the targets with the form
```graphql
query {
    targetNodes {
        mbeanMetrics {
            runtime {
                inputArguments
            }
        }
    }
}
```
and then parse the resulting array for the input argument: 
```
"-Dcom.sun.management.jmxremote.authenticate=false"
```
and checking if it `[exists, false, or true]`.

Not sure if there are better ways. The query should fail if the credentials are wrong anyways since we are calling the remote mxbean on the target, so it would be similar to getting the recordings in terms of checking if credentials work.

tthvo:
Oh niceee! Sounds good I will redirect to this approach :))

tthvo:
Worked very nicee! Thanks a lot!

Screenshot from 2023-03-24 18-26-35

andrewazores:
> The only thing I can think of right now, is to send a graphql request to each of the targets with the form
> 
> ```graphql
> query {
>     targetNodes {
>         mbeanMetrics {
>             runtime {
>                 inputArguments
>             }
>         }
>     }
> }
> ```
> 
> and then parse the resulting array for the input argument:
> 
> ```
> "-Dcom.sun.management.jmxremote.authenticate=false"
> ```
> 
> and checking if it `[exists, false, or true]`.
> 
> Not sure if there are better ways. The query should fail if the credentials are wrong anyways since we are calling the remote mxbean on the target, so it would be similar to getting the recordings in terms of checking if credentials work.
> 
> EDIT: Apprently that authenticate property is default true, so I assume if it doesn't exist, the jvm should need auth.

This feels a little fragile...

https://docs.oracle.com/javadb/10.10.1.2/adminguide/radminjmxenablepwd.html#radminjmxenablepwd

I would need to do a lot more checking of different docs for different JDK versions, but I would hesitate to say that `-Dcom.sun.management.jmxremote.authenticate` alone is sufficient to tell if the JVM is configured to use JMX credentials or not. Neglecting that property but setting `-Dcom.sun.management.jmxremote.password.file` would probably implicitly enable credentials too, I think, and there are probably other such combinations, and the behaviour could be different for different JDK versions.

----

If we are able to make this query and determine the `inputArguments` then this already means we're able to establish a connection to the target, so either it did not require credentials or we were successfully able to use stored ones.

Why don't we just have a connection checking endpoint that specifically does not load any stored credentials? It either uses only the session credentials header for the test, or it ignores the session credentials header and requires the credentials to test to be passed as part of the request body. We can use this endpoint and pass no credentials at first, and just try to open a JMX connection to read some basic MBean data. If that succeeds then we know the target did not require auth. Then the user can supply credentials and we can repeat the test using those.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -1154,6 +1174,64 @@ export class ApiService {
...
+          mbeanMetrics {
+            runtime {
+              inputArguments
+            }
+          }
+        }
+      }
+      `,
+      {
+        filter: { name: target.connectUrl },
```
andrewazores, line 1203: Hmm. Seems like I probably made a mistake designing this request filter. The `labels` and `annotations` fields can use syntax like `myLabel in (a, b, c)`, which is equivalent to a syntax like `myLabel == "a" || myLabel == "b" || myLabel == "c"`. It looks like it would've been useful to be able to apply this kind of syntax to other things (node names) than just labels and annotations. I didn't originally envision GraphQL queries being used this way but it does make sense. If we could make requests like this then it would mean the whole test could be done with one request-response pair between client and server, instead of one pair for each target under test. Right now the overall test looks like:

```
client -> server -> targetA
client <- server <__/
...
client -> server -> targetB
client <- server <__/
...
client -> server -> targetC
client <- server <__/
```

But with this properly constructed batch request from the client, it could look like:

```
                            /--> targetA \___
client -> server ---> targetB -----  > server -> client
                           \__> targetC __/
```

In the end there may not be a ton of difference if the client is able to make concurrent requests to the server in the first case, since in the end either way ends up with multiple connections opened to targets at a time, so the overall wall time from first request start to last response end is probably not too much different. But, there is a different in overhead of network traffic and request parsing etc. that can be reduced.

andrewazores: https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/resources/types.graphqls#L19

https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java#L69

https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java#L108 / https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java#L115

https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java#L97 / https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java#L104

I think this functionality can be retrofitted into the existing system pretty cleanly actually:

```diff
diff --git a/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java b/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java
index ee783a7c..a845453d 100644
--- a/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java
@@ -67,6 +67,7 @@ class FilterInput {
     enum Key {
         ID("id"),
         NAME("name"),
+        NAMES("names"),
         LABELS("labels"),
         ANNOTATIONS("annotations"),
         SOURCE_TARGET("sourceTarget"),
diff --git a/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java b/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
index 2e6df720..a966d1bd 100644
--- a/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
@@ -112,6 +112,16 @@ class TargetNodesFetcher extends AbstractPermissionedDataFetcher<List<TargetNode
                             .filter(n -> Objects.equals(n.getName(), nodeName))
                             .collect(Collectors.toList());
         }
+        if (filter.contains(FilterInput.Key.NAMES)) {
+            List<String> names = filter.get(FilterInput.Key.NAMES);
+            for (String name : names) {
+                result =
+                        result.stream()
+                                .filter(n -> LabelSelectorMatcher.parse(name).test(Map.of("name",
+                                                n.getName())))
+                                .collect(Collectors.toList());
+            }
+        }
         if (filter.contains(FilterInput.Key.LABELS)) {
             List<String> labels = filter.get(FilterInput.Key.LABELS);
             for (String label : labels) {
diff --git a/src/main/resources/types.graphqls b/src/main/resources/types.graphqls
index 410efb66..5b15af3d 100644
--- a/src/main/resources/types.graphqls
+++ b/src/main/resources/types.graphqls
@@ -17,6 +17,7 @@ input EnvironmentNodeFilterInput {
 input TargetNodesFilterInput {
     id: Int
     name: String
+    names: [String]
     labels: [String]
     annotations: [String]
 }
```

haven't tested this, just an idea. This would just allow for a single query to match a given node if its name is not just equal to a specific value, but if its name is a member of a given set. Then clients can make much more flexible requests targeting sets of nodes and doing further queries on those all server-side. Thoughts? If this makes sense I will go ahead and work on a PR.

tthvo: Great idea and makes sense to me! This affects only the `Test All` functionality right? Individual tests will still send individual requests?

andrewazores: Right, this would mostly just be useful for `Test All`. The individual testing requests could still use the same query syntax if it's easier to reuse code that way on the frontend though. Basically the difference would be something like:

```graphql
query {
  // $someNode should be passed as a variable, not embedded in the query
  targetNodes(filter: { name: "$someNode" } ) {
    name
  }
}
```

for testing using the single-node specific query

vs

```graphql
query {
  // $someNode list contents should be passed as a variable, not embedded in the query
  targetNodes(filter: { names: ["name in ($someNode)"] } ) {
    name
  }
}
```

for testing a single node with the set matching query.

andrewazores:
I guess this can use either the new explicit credentials testing endpoint or the GraphQL name-from-set filtering, but it doesn't really make sense for this feature to try to use both, right?

In the near-term it's probably more sensible to just use the credentials testing endpoint since it will be able to provide all three state possibilities (none needed, good credentials, bad credentials) whereas the GraphQL query would only be able to report whether the query succeeded or not, so the detail about whether the credentials worked or were just not needed would be lost.

tthvo:
> I guess this can use either the new explicit credentials testing endpoint or the GraphQL name-from-set filtering, but it doesn't really make sense for this feature to try to use both, right?

Yehh i am leaning towards the test endpoint for this PR since the front-end states are set up to work with that. Maybe we can have the graphQL functionalities for 2.4.0 or later?

maxcao13:
image

The tooltip says "ignoring filters", does that mean the test all feature should ignore filters and test all anyways?

tthvo:
Yehh it just means that all matching targets will start testing even if you filter it out (e.g. for example, by status or by name).

maxcao13 (approved):
Looks great to me! Awesome job.

andrewazores (approved):
Awesome indeed.

PR #921, opened 2023-03-23
feat(topology): expanded row for resource details

tthvo:
# Welcome to Cryostat! 👋

Related to #891 
Depends on https://github.com/cryostatio/cryostat/issues/1428

## Description of the change:

Each resource row can have an expanded section that show detail view. For now, only `ActiveRecording` row is supported. Others will show an empty text.

## Motivation for the change:

See https://github.com/cryostatio/cryostat-web/pull/906#issuecomment-1472888342

## Screenshots

Screenshot from 2023-03-22 21-38-37

### DIFF
--- a/src/app/Topology/Actions/CreateTarget.tsx
+++ b/src/app/Topology/Actions/CreateTarget.tsx
@@ -47,6 +47,7 @@ import { Target } from '@app/Shared/Services/Target.service';
 import '@app/Topology/styles/base.css';
 import { getFromLocalStorage } from '@app/utils/LocalStorage';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
+import { portalRoot } from '@app/utils/utils';
 import {
   Accordion,
   AccordionContent,
@@ -495,6 +496,7 @@ export const SampleNodeDonut: React.FC<SampleNodeDonutProps> = ({
       <Flex className={css(className)} direction={{ default: 'column' }}>
         <FlexItem alignSelf={{ default: 'alignSelfCenter' }}>
           <Tooltip
+            appendTo={portalRoot}
             content={
               _actionEnabled
                 ? `Click to test${validation.option !== ValidatedOptions.default ? ' again' : ''}.`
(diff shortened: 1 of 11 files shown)


andrewazores:
Looks good so far. Can there also be a badge like this one:

image

that indicates whether there are recordings present? Maybe if this can't be determined (ie. jvmId is null which indicates we haven't been able to connect, as is now the case) show a warning icon. If we can check and there is at least one running recording then show a green OK state icon. Otherwise, either show no icon, or show some neutral grey state icon. WDYT?

tthvo:
Error  & No-running-recording states won't show any icon. Only targets with running recordings will show the green icon.

Just a notice that: we should order our css after PF css or else some rules might be overriden unexpectedly.

Screenshot from 2023-03-23 18-25-33

tthvo:
Looks like the `useResources` hook in the PR can be generalize to accept a target and reused in many other places :D

maxcao13:
image

Should this be reactive to backend emitted websocket notifications? It seems like it does not right now.

tthvo:
Oh seems to affect the `stopped` state. I will look into it.

andrewazores:
I don't know for sure off-hand but I would assume it's because that notification gets emitted with the recording object as the payload right before the recording actually gets stopped, or at least the payload object on the server is still holding the original state that was pulled from the target without it getting updated with the new state after the stoppage occurred. Seems like it should be a minor and easily fixable backend bug, I suspect in the `RecordingTargetHelper`

tthvo:
Right thanks! Seems to return the descriptor without mutating the state here:

https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/recordings/RecordingTargetHelper.java#L276

tthvo:
Latest changes just append the tooltips to `portalRoot` and remove checks for stop-rec notification since backend now should send correct rec state :))

andrewazores (approved):

Issue #918, opened 2023-03-21
[Task] Add dark mode


maxcao13:
## Describe the feature
Seems like patternfly silently added dark mode components but have not documented it. This is what it looks like with Cryostat:

Dashboard:
![image](https://user-images.githubusercontent.com/69827283/226744437-75152c17-f6e1-437c-ab69-eb66598498af.png)

Topology:
![image](https://user-images.githubusercontent.com/69827283/226743948-7e7bd4cf-4afd-4ac2-b9e2-7e87e6e50cc1.png)

Recordings:
![image](https://user-images.githubusercontent.com/69827283/226744565-0e757b9e-0bc3-417e-877c-81a773ba90b6.png)

Some components may look a bit weird with this color scheme but I think at least having the option in for now doesn't hurt for this release. It is really easy to implement anyways because it is just adding a class to the document element.

PR #925, opened 2023-03-27
feat(themes): add 'dark mode' theme

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #918

## Description of the change:
* Adds dark mode option to Cryostat. :smile: 
There may still be potential dark mode conflicting css styles in the code, I tried my best to fix them all.

* fixed popovers + tooltips to appendTo portalRoot instead of document body.
* renamed `Language & Region` Settings category to `General`.
* fixed dependency cycles with `Settings.tsx`

## How to manually test:
1. Run normally and go to settings and go to `General` to try out dark mode.
2. Try rendering different Cryostat pages and components and notice if there are any contrast/accesibility issues with the coloring scheme and add a review comment if so.


src/app/app.css
```diff
@@ -484,7 +509,7 @@ input[type=number].datetime-picker__number-input {
 }
 
 .expandable-form__form-group.pf-c-form__group .pf-c-form__label-text {
-  color: #000000;
+  color: var(--pf-global--Color--100);
```
tthvo, line 512: We gonna need to have:

- Light: `var(--pf-global--Color--100)`
- Dark: `var(--pf-global--palette--white)` to make it the same color as the form group above.

tthvo (commented):
Looks awesome! Finally a long-waiting dark mode is coming. Here are some that need changes:

### Meridiem Picker

Screenshot from 2023-03-27 16-35-28


### Label Filters

Screenshot from 2023-03-27 16-36-50

### Language preference

This should lead the general tab but i think tab name might be incorrect here. Also, we need to set the default tab to General when going to setting (now its defaulting to Connectivity).

Screenshot from 2023-03-27 16-38-12


### Entity Detail

The title text should be given a brighter color in dark mode.

Screenshot from 2023-03-27 16-42-50

### Custom Target Node

Icon and name label (below the target) needs adjusting. Seems to disappear for now.

Screenshot from 2023-03-27 16-43-50

src/app/Topology/styles/base.css
```diff
@@ -304,6 +303,10 @@ Below CSS rules only apply to Topology components
   background-color: var(--pf-global--BackgroundColor--200);
 }
 
+.pf-theme-dark .topology__quicksearch__tab-content {
+  background-color: var(--pf-global--BackgroundColor--300);
```
tthvo, line 307: Just saw this one: https://github.com/patternfly/patternfly/pull/2818.

tthvo: Probably use this: `var(--pf-global--BackgroundColor--dark-300)`

maxcao13:
I think iframes are unable to be changed because of the cross origin issues, as for the grafana, I use localstorage to set the theme to the user default directly accessing localstorage since the descriptor is not within a FC.

src/app/Dashboard/Charts/jfr/JFRMetricsChartCard.tsx
```diff
@@ -321,8 +323,8 @@ export const JFRMetricsChartCardDescriptor: DashboardCardDescriptor = {
     {
       name: 'CHART_CARD.PROP_CONTROLS.THEME.NAME',
       key: 'theme',
-      values: ['light', 'dark'],
-      defaultValue: 'light',
+      values: Object.values(ThemeType),
+      defaultValue: getFromLocalStorage('THEME', ThemeType.LIGHT),
```
tthvo, line 327: I think this is only intialized once so if the user changes preferences in between, it wont get updated. Maybe just remove this config and add a subscription to theme in JFR charts?

src/app/app.css
```diff
@@ -194,16 +194,25 @@ html, body, #root {
...
-  background-color: var(--pf-global--Color--light-200);
-  border-color: var(--pf-global--BorderColor--100);
+  background-color: var(--pf-global--BackgroundColor--light-300);
+  border-color: var(--pf-global--BorderColor--light-100);
   border-style: solid;
   border-width: 0.1em;
   border-radius: 0.5em;
 }
 
+.pf-theme-dark .pf-c-label-group.pf-m-vertical.pf-c-label-group {
```
tthvo, line 211: I think we can remove this? There is already some styling for dark-mode label-group that looks quite the same?

tthvo (commented):
Looks good! Just topology-related issues and JFR chart dark theme.

andrewazores:
> Report
> Dark mode this one too? If not possible, i guess its fine since its an iframe (not exactly part of our front-end).

Well there is this: https://github.com/cryostatio/cryostat-core/blob/main/src/main/java/io/cryostat/core/reports/ReportTransformer.java

but that's a lot of change pipeline just to support dark mode, since this means there needs to be a way for the client to request a dark mode report from the backend, which needs to either produce it directly (with the forked subprocess generator) or include that parameter along to the report generator sidecar, which means `cryostat-reports` also needs to implement a `ReportTransformer` for dark mode...

It'd be faster, better, and easier just to replace the embedded iframe here with the report component that Max wrote for the dashboard card.

> Embeded Grafana
> Would it make sense to override user choice to set theme as dark here? Or maybe we can just remove the option to set theme when adding card and get these graphs to listen to our theme settings?
> I think iframes are unable to be changed because of the cross origin issues, as for the grafana, I use localstorage to set the theme to the user default directly accessing localstorage since the descriptor is not within a FC.

Removing the explicit choice and just rendering the charts based on the current user setting sounds like the right thing to do here.

src/app/About/About.tsx
```diff
@@ -48,12 +51,22 @@ export interface AboutProps {}
 
 export const About: React.FC<AboutProps> = (_) => {
   const { t } = useTranslation('public');
+  const theme = useTheme();
+  const [logo, setLogo] = React.useState(cryostatLogo);
+
+  React.useEffect(() => {
```
tthvo, line 57: I think we can reduce this to:

```
const log = React.useMemo(() => theme === ThemeType.DARK? cryostatLogoDark): cryostatLogo, [theme]);
```

src/test/Settings/FeatureLevels.test.tsx
```diff
@@ -75,15 +75,19 @@ describe('<FeatureLevels/>', () => {
     expect(ul).toBeInTheDocument();
     expect(ul).toBeVisible();
 
-    await user.selectOptions(ul, testT(FeatureLevel[FeatureLevel.BETA], 'common'));
+    await user.selectOptions(ul, testT(FeatureLevel[FeatureLevel.BETA], { ns: 'common' }));
 
-    expect(ul).not.toBeInTheDocument(); // Should close menu
+    await expect(ul).not.toBeInTheDocument(); // Should close menu
```
tthvo, line 80: Maybe not await here since `expect` is not async.

src/test/Settings/FeatureLevels.test.tsx
```diff
@@ -75,15 +75,19 @@ describe('<FeatureLevels/>', () => {
...
 
-    const betaOption = screen.getByText(testT(FeatureLevel[FeatureLevel.BETA], 'common'));
+    const betaOption = await screen.findByText(
```
tthvo, line 82: Was there an issue with this? Ideally, this line:

```ts
 await user.selectOptions(ul, testT(FeatureLevel[FeatureLevel.BETA], { ns: 'common' }));
```

should flush out all tasks and correctly update component state

src/app/AppLayout/AppLayout.tsx
```diff
@@ -403,7 +413,7 @@ const AppLayout: React.FC<AppLayoutProps> = ({ children }) => {
                   <Button
                     onClick={handleSettingsButtonClick}
                     variant="link"
-                    icon={<CogIcon color="white " size="sm" />}
+                    icon={<CogIcon color="var(--pf-global--palette--white)" size="sm" />}
```
tthvo, line 416: The setting icon appears to bright in dark mode. Lets try this:

```diff
diff --git a/src/app/AppLayout/AppLayout.tsx b/src/app/AppLayout/AppLayout.tsx
index b4e78c36..892be75c 100644
--- a/src/app/AppLayout/AppLayout.tsx
+++ b/src/app/AppLayout/AppLayout.tsx
@@ -412,8 +412,8 @@ const AppLayout: React.FC<AppLayoutProps> = ({ children }) => {
                 <ToolbarItem>
                   <Button
                     onClick={handleSettingsButtonClick}
-                    variant="link"
-                    icon={<CogIcon color="var(--pf-global--palette--white)" size="sm" />}
+                    variant='plain'
+                    icon={<CogIcon size="sm" />}
                   />
                 </ToolbarItem>
                 <ToolbarItem>
```

tthvo: We can change the variant for the notification test button too.

src/app/app.css
```diff
@@ -264,6 +264,10 @@ html, body, #root {
   max-width: 100%;
 }
 
+:where(.pf-theme-dark) .automated-analysis-grid-item .pf-c-label {
+  --pf-c-label--BackgroundColor: var(--pf-global--BackgroundColor--dark-100);
```
tthvo, line 268: Seems like there is a higher specificity rule somewhere:

Screenshot from 2023-03-29 01-18-39


We can add `!important` here

src/test/Settings/Theme.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    const option = within(ul).getByText(testT('SETTINGS.THEME.DARK'));
+    expect(option).toBeInTheDocument();
+    expect(option).toBeVisible();
+
+    await act(async () => {
+      await user.click(option);
+    });
+
+    screen.logTestingPlaygroundURL();
```
tthvo, line 89: Interesting here... We might not need it here but seems like some we can explore for test framework.

src/app/Topology/styles/base.css
```diff
@@ -289,7 +304,7 @@ Below CSS rules only apply to Topology components
   box-sizing: border-box;
   border-radius: 8px;
   padding: 4px;
-  background-color: #fff;
+  background-color: var(--pf-global--palette--white);
 }
 
 .topology__quicksearch__tab-text {
```
tthvo, line 310: Lets try this:

```diff
diff --git a/src/app/Topology/Actions/QuickSearchPanel.tsx b/src/app/Topology/Actions/QuickSearchPanel.tsx
index 90a03828..e8c182b8 100644
--- a/src/app/Topology/Actions/QuickSearchPanel.tsx
+++ b/src/app/Topology/Actions/QuickSearchPanel.tsx
@@ -197,6 +197,7 @@ export const QuickSearchPanel: React.FC<QuickSearchPanelProps> = ({ ...props })
           <Sidebar {...props} tabIndex={0} style={{ height: 'max-content' }} hasGutter>
             <SidebarPanel variant="sticky">
               <Tabs
+                className={"topology__quicksearch__tabs"}
                 isVertical
                 unmountOnExit
                 expandable={{ default: 'nonExpandable', md: 'nonExpandable', lg: 'nonExpandable', sm: 'expandable' }}
diff --git a/src/app/Topology/styles/base.css b/src/app/Topology/styles/base.css
index 3f77a668..cb4972c6 100644
--- a/src/app/Topology/styles/base.css
+++ b/src/app/Topology/styles/base.css
@@ -315,10 +315,22 @@ Below CSS rules only apply to Topology components
   margin: 0 !important;
 }
 
+:where(.pf-theme-dark) .topology__quicksearch__tabs {
+  background-color: var(--pf-global--BackgroundColor--dark-300);
+}
+
+:where(.pf-theme-dark) .topology__quicksearch__tabs .pf-c-label {
+  background-color: var(--pf-global--BackgroundColor--dark-200);
+}
+
 .topology__quicksearch__tab.pf-m-current {
   background-color: var(--pf-global--BackgroundColor--200);
 }
 
+:where(.pf-theme-dark) .topology__quicksearch__tab.pf-m-current {
+  background-color: var(--pf-global--BackgroundColor--100);
+}
+
 :where(.pf-theme-dark) .topology__quicksearch__tab-content {
   background-color: var(--pf-global--BackgroundColor--dark-300);
 }
@@ -330,10 +342,14 @@ Below CSS rules only apply to Topology components
 }

```

Screenshot from 2023-03-29 02-25-13

src/app/Topology/styles/base.css
```diff
@@ -50,10 +50,14 @@ Below CSS rules only apply to Topology components
   aspect-ratio: 1;
   border-radius: 50%;
   border: 1px solid var(--pf-global--palette--black-300);
-  background-color: #fff;
+  background-color: var(--pf-global--palette--white);
   margin: 1em;
 }
```
tthvo, line 56: Lets try this:

```diff
diff --git a/src/app/Topology/styles/base.css b/src/app/Topology/styles/base.css
index 3f77a668..c17b356f 100644
--- a/src/app/Topology/styles/base.css
+++ b/src/app/Topology/styles/base.css
@@ -55,7 +55,7 @@ Below CSS rules only apply to Topology components
 }
 
 :where(.pf-theme-dark) .sample-node-donut__node-wrapper {
-  background-color: var(--pf-global--BackgroundColor--300);
+  background-color: var(--pf-global--BackgroundColor--dark-300);
 }
 
 .sample-node-donut__node-wrapper .sample-node-donut__node-icon {
@@ -96,7 +96,7 @@ Below CSS rules only apply to Topology components
 }
 
 :where(.pf-theme-dark) .sample-node-donut__node-wrapper .sample-node-donut__status-indicator {
-  background-color: var(--pf-global--BackgroundColor--300);
+  background-color: var(--pf-global--BackgroundColor--dark-300);
 }
 
 .sample-node-donut__node-wrapper .sample-node-donut__status-indicator svg {
@@ -113,6 +113,11 @@ Below CSS rules only apply to Topology components
   box-shadow: 0 0.25rem 0.5rem 0rem rgba(3, 3, 3, 0.16), 0 0 0.25rem 0 rgba(3, 3, 3, 0.06);
   font-weight: 700;
   word-break: break-all;
+  background-color: var(--pf-global--palette--white);
+}
+
+:where(.pf-theme-dark) .sample-node-donut__node-label {
+  background-color: var(--pf-global--BackgroundColor--dark-300);
 }
```

Screenshot from 2023-03-29 02-43-31

tthvo (commented):
Looks great! Just some more suggestions for topology-related styles :))

src/test/Settings/Theme.test.tsx
maxcao13: Oh, I just forgot to remove it. I use it to debug what the testing env looks like when I write tests :D

tthvo (dismissed):
Look awesome to me!

andrewazores (approved):
Looks great! I will wait to merge this until a few of the other open PRs are merged, just so any further dark-mode related changes can be part of this PR rather than the others.

andrewazores:
@maxcao13 rebase please :-)

maxcao13:
Oh, I need to fix the quick start settings, and some more coloring.

src/app/QuickStarts/quickstarts/automated-rules-quickstart.tsx
```diff
@@ -36,12 +36,15 @@
...
 import cryostatLogoIcon from '@app/assets/cryostat_icon_rgb_default.svg';
-import cryostatLogo from '@app/assets/cryostat_logo_vert_rgb_default.svg';
+import cryostatLogoIconDark from '@app/assets/cryostat_icon_rgb_reverse.svg';
 import build from '@app/build.json';
 import { FeatureLevel } from '@app/Shared/Services/Settings.service';
+import { withTheme } from '@app/utils/withTheme';
 import { QuickStart } from '@patternfly/quickstarts';
+import React from 'react';
+
+const Icon = withTheme(cryostatLogoIcon, cryostatLogoIconDark, 'Cryostat Icon');
```
tthvo, line 47: Missing metadata order here?

src/app/index.tsx
```diff
@@ -36,6 +36,7 @@
  * SOFTWARE.
  */
 
+import '@patternfly/patternfly/patternfly-charts.css';
```
tthvo, line 39: Import after base.css? Maybe alongside quickstarts.css

andrewazores:
Is it much work to add an "auto" setting here? From a cursory glance at the implementation it seems doable.

image

Basically so the existing light/dark settings force that appearance, and "auto" would use the current behaviour you have implemented to detect what to default to on first run, but it would do that on every start. This way if the user changes themes often (perhaps by time of day) then our UI will also dynamically update, at least each time they open/refresh it.

andrewazores:
image

image

Should the little triangular bit also be darkened?

src/app/Settings/SettingsUtils.ts
```diff
@@ -79,7 +79,17 @@ export const selectTab = (tabKey: SettingCategory) => {
...
 };
 
-export enum ThemeType {
-  LIGHT = 'light',
+export const getDefaultTheme = (): ThemeSetting => {
+  if (window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
+    return ThemeSetting.DARK;
+  }
+  return ThemeSetting.LIGHT;
+};
```
andrewazores, line 87: This could be made into a `() => Observable<ThemeSetting>` by using this snippet (taken from Stack Overflow):

```ts
window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', event => {
    const newColorScheme = event.matches ? "dark" : "light";
});
```

and emitting `newColorScheme` as the `.next()` value of some `Subject`. Seems like a relatively straightforward refactor I think, but it /should/ allow the UI to switch themes without refreshing/reloading.

andrewazores: Pretty close:

https://notiz.dev/blog/media-observable

src/app/QuickStarts/QuickStartsCatalogPage.tsx
```diff
@@ -53,6 +53,11 @@ const QuickStartsCatalogPage: React.FunctionComponent<QuickStartsCatalogPageProp
       hint={t('QuickStarts.CATALOG_PAGE.HINT')}
       showTitle
       showFilter
+      sortFnc={(a, b) => {
+        console.log(a.metadata.order, b.metadata.order);
+        console.log(a.spec.displayName, b.spec.displayName);
```
andrewazores, line 58: cleanup

src/app/utils/withTheme.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import { ThemeSetting } from '@app/Settings/SettingsUtils';
+import { useTheme } from '@app/utils/useTheme';
+import React from 'react';
+
+export const withTheme = (icon: string, darkIcon: string, alt: string): React.FC => {
```
andrewazores, line 42: Can this component be replaced with a plain HTML5 element?

```ts
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="./docs/images/cryostat_logo_hori_rgb_reverse.svg">
    <img src="./docs/images/cryostat_logo_hori_rgb_default.svg">
  </picture>
```

(https://github.com/cryostatio/cryostat/blob/a5733e78017f4cb4cd26246859e8a00991fccccd/README.md?plain=1#L6)

maxcao13: It would be nice if this worked like this, but the html elements would only respond to the media query events within the browser but not to the Cryostat theme setting. Is there another way to do this only using plain html?

src/app/Shared/Services/Settings.service.tsx
```diff
@@ -90,12 +90,32 @@ export class SettingsService {
...
     this._featureLevel$.subscribe((featureLevel: FeatureLevel) => saveToLocalStorage('FEATURE_LEVEL', featureLevel));
     this._visibleNotificationsCount$.subscribe((count: number) =>
       saveToLocalStorage('VISIBLE_NOTIFICATIONS_COUNT', count)
     );
     this._datetimeFormat$.subscribe((format: DatetimeFormat) => saveToLocalStorage('DATETIME_FORMAT', format));
+    this._theme$.subscribe((theme: ThemeSetting) => saveToLocalStorage('THEME', theme));
+  }
+
+  // taken from https://github.com/garygrossgarten/media-query
+  media(query: string): Observable<boolean> {
```
tthvo, line 105: I think this observable should emi the MediaQueryList instead. The consumer can just map to something else as needed so that this can be reusable.

src/app/utils/useTheme.ts
```diff
@@ -0,0 +1,77 @@
...
+  }, [services.settings, themeRef]);
+
+  React.useEffect(() => {
+    mediaRef.current = services.settings.media('(prefers-color-scheme: dark)').subscribe((dark) => {
+      if (setting === ThemeSetting.AUTO) {
+        setTheme(dark ? ThemeSetting.DARK : ThemeSetting.LIGHT);
+      }
+    });
+    return () => mediaRef.current && mediaRef.current.unsubscribe();
+  }, [services.settings, mediaRef, setting]);
```
tthvo, line 74: Relying on setting here means this effect is run for every setting change. Ideally, we want to subscribe once.

Work around is to call setState but return old state. This way u can have access to most-up-to-date state and effect only runs once.

```ts
mediaRef.current = services.settings.media('(prefers-color-scheme: dark)').subscribe((dark) => {
  setSetting((setting) => {
    if (setting === ThemeSetting.AUTO) {
      setTheme(dark ? ThemeSetting.DARK : ThemeSetting.LIGHT);
    }
    return setting;
  });
});
```

src/app/utils/withTheme.tsx
tthvo: yehh...thats unfortunate cuz this setup is outside React life cycle so there is hardly anyway around.

tthvo, line 42: Btw lets call it `WithThemeIcon` as we are building icon only. Just to avoid confusion :))

test-setup.js
```diff
@@ -1,6 +1,20 @@
 // Mock out the services shared across the app in order to help isolate 
 // components from the ServiceContext
 
+Object.defineProperty(window, 'matchMedia', {
+  writable: true,
+  value: jest.fn().mockImplementation(query => ({
```
tthvo, line 6: Try this? Just cleaner

```ts
value: jest.fn(query => ({
    matches: false,
    media: query,
    onchange: null,
    addListener: jest.fn(), // deprecated
    removeListener: jest.fn(), // deprecated
    addEventListener: jest.fn(),
    removeEventListener: jest.fn(),
    dispatchEvent: jest.fn(),
  })),
```

src/app/utils/useTheme.ts
```diff
@@ -0,0 +1,77 @@
...
+// setting is the option, but theme is the color scheme what we actually render
+export function useTheme(): [ThemeType, ThemeSetting] {
+  const [setting, setSetting] = React.useState<ThemeSetting>(ThemeSetting.LIGHT);
+  const [theme, setTheme] = React.useState<ThemeType>(ThemeSetting.LIGHT);
+  const themeRef = React.useRef<Subscription>();
+  const mediaRef = React.useRef<Subscription>();
+  const services = React.useContext(ServiceContext);
+
+  React.useEffect(() => {
+    themeRef.current = services.settings.theme().subscribe((theme) => {
```
tthvo, line 52: Lets call this `services.settings.theme()` -> `services.settings.themeSetting()` because its a setting, not the exact theme.

tthvo (commented):
Very niceee! Just some more comments :))

test-setup.js
```diff
@@ -3,7 +3,7 @@
 
 Object.defineProperty(window, 'matchMedia', {
   writable: true,
-  value: jest.fn().mockImplementation(query => ({
+  value: jest.fn(query => ({
```
tthvo, line 6: Why not returning `mockMediaQueryList` here?

andrewazores (approved):
LGTM, any other thoughts @tthvo ?

tthvo (approved):
Nope! Looks great to me!!!

Issue #955, opened 2023-04-14
[Bug] Entity Detail failed to update when recordings are created


tthvo:
## Current Behavior:

Crash with error view saying template is not defined.

## Expected Behavior:

Recording counts should be updated normally.

## Screenshots

![Screenshot from 2023-04-14 15-38-58](https://user-images.githubusercontent.com/68053619/232140056-ec9cf214-74c1-4394-ac81-c0ac1ceb827d.png)

tthvo:
I think I accidentally removed the common handler for Active/Archived recordings when cleaning up UploadedArchives and it switch case leaks through EventTemplate. Will add it back.

tthvo:
Very fortunate that we have an error boundary to see the error :))

PR #958, opened 2023-04-14
fix(topology): fix resource update in entity detail view

tthvo:
# Welcome to Cryostat! 👋

Fixes: #955 

## Description of the change:

- [x] Add back patchFn for active and archived recording (missing due to refactors).
- [x] Do not show expanded view when there is an error.
- [x] Fix observable being cancelled on error. This causes the resource table to not update.  

## Motivation for the change:

See #955

### DIFF
--- a/src/app/Topology/Shared/Entity/EntityDetails.tsx
+++ b/src/app/Topology/Shared/Entity/EntityDetails.tsx
@@ -84,6 +84,7 @@ import {
   getExpandedResourceDetails,
   getLinkPropsForTargetResource,
   mapSection,
+  Nothing,
   TargetOwnedResourceType,
   TargetOwnedResourceTypeAsArray,
   TargetRelatedResourceType,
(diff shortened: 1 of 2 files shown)


tthvo:
What a painful lession with Observable :)) Errors make the entire observable chain to cancel and nothing updates (need to chain a catchError).

Also, one question is: since we have Podman discovery + JDP, the same target is represented twice with different connectionURL. This causes the resource table not to update (if choose target 1.1) when u try to create a recording in target 1.2. Maybe is there another way to check? JvmID? The event sends only connectionURL tho.

andrewazores:
> Also, one question is: since we have Podman discovery + JDP, the same target is represented twice with different connectionURL. This causes the resource table not to update (if choose target 1.1) when u try to create a recording in target 1.2. Maybe is there another way to check? JvmID? The event sends only connectionURL tho.

Let's put a task for the next release that all events like this that currently only emit the connectionUrl should also emit the JvmId. That should help solve these cases properly.

andrewazores (approved):

Issue #997, opened 2023-04-26
[Bug] Error popover is not showing when uploading labels


tthvo:
## Current Behavior:

After https://github.com/cryostatio/cryostat-web/pull/624, there is a popover that will show if any metadata files are invalid. However, it no longer shows.

## Expected Behavior:

Error popover should show if any metadata files are invalid.

## Steps To Reproduce:
1. With latest cryostat image, launch Cryostat.
2. Go to Archives and click Upload.
3. Expand metadata options and try uploading any invalid metadata files (a valid ones are usually downloaded with recordings and invalid ones can be any files with invalid JSON).
4. Notice how the popover does not show.
5. However, use browser inspector to check that the DOM node is mounted but popper css applies a hidden rule to the DOM.

## Anything else

There seems to be issues with mounting the upload modal on `portalRoot`. Somehow, the popper sees the button as clipped/hidden and apply `data-popper-reference-hidden` attribute to the popper div.

```css
/* Hide the popper when the reference is hidden */
#popper[data-popper-reference-hidden] {
  visibility: hidden;
  pointer-events: none;
}
```

See: https://popper.js.org/docs/v2/modifiers/hide/

PR #998, opened 2023-04-26
fix(uploads): fix archive upload modal invalid labels popover 

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #997

## Description of the change:
The change reverts ArchiveUploadModal to append to document body for the popover to show again. I cannot figure out quite the reason why appending the modal to portalRoot makes popperjs think that the popper is being cutoff in some way.

According to an [article](https://dev.to/atomiks/everything-i-know-about-positioning-poppers-tooltips-popovers-dropdowns-in-uis-3nkl) I read: 
```
Popper attaches attributes in the following cases:

    data-popper-escaped: When the popper escapes the reference element's clipping container (it appears detached)
    data-popper-reference-hidden: When the reference element is hidden from view (it appears attached to nothing)
```
But it doesn't seem the reference is ever hidden from view, even when the expandedContent is expanded, and it doesn't make sense why having the modal append to portalRoot vs root/body would add these attributes.

For now, appending the modal to document body is completely ok.

The change also shows the popover if invalid jsons that dont have the `label` field are uploaded.

## How to manually test:
1. go to `Archives` -> `Upload` -> Test out bad metadata json files e.g. `{ASDfsadf}` and `{"labels":{"asdf"""":"asdf"}}`

### DIFF
--- a/locales/en/public.json
+++ b/locales/en/public.json
@@ -274,6 +274,10 @@
       "TITLE": "Quick Starts"
     }
   },
+  "RecordingLabelFields": {
+    "INVALID_UPLOADS_one": "The following file did not contain valid recording metadata:",
+    "INVALID_UPLOADS_other": "The following files did not contain valid recording metadata:"
+  },
   "SETTINGS": {
     "AUTO_REFRESH": {
       "CHECKBOX_LABEL": "Enabled",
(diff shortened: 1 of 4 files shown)


tthvo (approved):
Looks good! Very weird with `Popper.js` indeed.

PR #1004, opened 2023-05-01
fix(settings): set unmountOnExit behaviour for setting groups

tthvo:
# Welcome to Cryostat! 👋

Related to #793 

## Description of the change:

Unmount setting tab content that is not selected. This helps avoid situation where AA configurations cause the Auth modal to appear when first visiting `/settings` (users might wonder why this opens as they don't see the AA config). This also allows faster load as React only needs to build 1 tab content at a time.
 
Extend the min height for local target select to show more targets (currently it looks like, without scrolling, there is a single target).

## Motivation for the change:

AA configurations causes the setting page to show Auth modal even if the AA configuration is not visible (i.e. Dashboard tab not selected).

### DIFF
--- a/src/app/Settings/Settings.tsx
+++ b/src/app/Settings/Settings.tsx
@@ -193,12 +193,10 @@ export const Settings: React.FC<SettingsProps> = (_) => {
               </Tabs>
             </SidebarPanel>
             <SidebarContent>
-              {settingGroups.map((grp) => (
-                <div
-                  key={`${grp.groupKey}-setting`}
-                  className={css('settings__content', grp.groupKey === activeTab ? 'active' : '')}
-                >
-                  <Form>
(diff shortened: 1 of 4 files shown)


maxcao13 (approved):

andrewazores:
In both Firefox and Chrome this change seems to just break the Settings view for me:

image

Every tab selection is just empty. The DOM elements *are* there, but they have css `display: none`:

image

If I use the inspector to remove that css property then it's fine.

image

tthvo:
Ohh thanks that should not be there. Interesting that it was working for me with that css in place (maybe I was messing around with browser inspector at the time).

andrewazores (approved):

andrewazores (approved):

Issue #1017, opened 2023-05-05
[Task] Parallelize CI tests for eslint, format, and jest tests


maxcao13:
## Feature:
It would be nice if the 3 CI tests could be in parallel, it seems the yarn tests take the most time ~ 2min, while eslint may take 1~2 minutes and the format checker takes around ~12 seconds. This is already a feature in the cryostat main repo where the shellcheck, spotless, and junit tests are done in parallel.

PR #1029, opened 2023-05-10
chore(ci): run testing jobs in parallel

tthvo:
# Welcome to Cryostat! 👋

Fixes: #1017 

## Description of the change:

- Split test ci steps into jobs that can be run in parallel.
- Remove `yarn build:notests` as a step since we don't need to build the asset for any tests.
- Set node version to `16.18.x` (cryostat builds use node `16.18.1`).

## Motivation for the change:

See #1017 

## Screenshots

Screenshot from 2023-05-10 05-42-30

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,13 +15,11 @@ on:
       - cryostat-v[0-9]+.[0-9]+
 
 jobs:
-  build:
+  prettier-check:
     runs-on: ubuntu-latest
-
     strategy:
       matrix:
-        node-version: [16.x]
-
+        node-version: [16.x, 18.x]
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: CI Test (PR)
```
andrewazores, line 1: `pr-ci.yml` is the one that runs against PRs. This file is the one that runs on pushes to `main`, or when PRs are merged to `main`, so I think this rename is a bit inaccurate.

tthvo: oh yes! thanks for catching that!

.github/workflows/ci.yaml
```diff
@@ -30,6 +28,29 @@ jobs:
         node-version: ${{ matrix.node-version }}
     - uses: bahmutov/npm-install@v1
     - run: yarn format:check
+  eslint-check:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
andrewazores, line 34: I think you can lift the `strategy` up to a higher level in the YAML to avoid repeating this matrix definition for each job:

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

.github/workflows/ci.yaml
```diff
@@ -14,22 +14,37 @@ on:
...
+  NODE_VERSION: 16.18.x
+
 jobs:
-  build:
+  prettier-check:
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        node-version: [16.x]
```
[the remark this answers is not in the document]
tthvo: Right that makes sense. I added back matrix with 16x and 18x.

> What version did 16.x end up resolving to?

I think it resolves to `16.20.x`. M

.github/workflows/ci.yaml
andrewazores: Wait, or maybe not. I may have misread this...

tthvo: Oh sorry not very familiar with these. But which level can we push this strategy to?

tthvo: Oh okay. Saw ur comment just in time.

andrewazores: Nevermind, I think it can only be done easily here for steps within a job as you've done, not for all the jobs within a workflow. There is probably some way to do it using a matrix with reusable workflows or something but I'm not sure and it may not be worth the effort for such a relatively small and simple workflow file.

andrewazores (approved):

tthvo:
Hmm what is that old `build (16.x) Expected — Waiting for status to be reported`sticking around?

andrewazores:
Oh, I think it's just because the name of the job changed, so I need to update the repo config since that specific name is set up as a required check for PR reviews.

tthvo:
Ohh thats new! Thanks!

andrewazores:
There we go, fixed.

Issue #1026, opened 2023-05-09
[Story] Mock API server for rapid frontend development


andrewazores:
See prior discussion:

https://github.com/cryostatio/cryostatio.github.io/pull/109#issuecomment-1540830435

https://github.com/cryostatio/cryostatio.github.io/pull/109#discussion_r1189072837

PR #1039, opened 2023-05-31
feat(dev): mock API server with miragejs

tthvo:
# Welcome to Cryostat! 👋

Fixes: #1026 

## Description of the change:

Use Mirage JS to mock API server and web socket for cases where a quick preview is needed. This also includes 2 yarn script to start hot-reload server and build asset with these mocks injected.

## Motivation for the change:

See #1026 

## How to manually test:

To start development server with API server mocked:

```
yarn start:dev:preview
```

To build assets with mocks injected

```
yarn build:preview:notests

# To quickly launch a server and serve those assets, use
cd dist && npx browser-sync start --server
```


tthvo:
Ready for review now!

 I got some issues with production css order when including Mirage JS that imports internal types (i.e. Target, Rule, ActiveRecording). So, for now, I just type as `any` to bypass static type checking.

We probably want to move all common types into a separate `src/app/typings.ts`. Maybe another follow-up PR then?

src/index.tsx
```diff
@@ -53,4 +55,12 @@ if (process.env.NODE_ENV !== 'production') {
   axe(React, ReactDOM, 1000, config);
 }
 
+if (process.env.PREVIEW === 'true') {
```
andrewazores, line 58: I think `process.env.NODE_ENV` is commonly used for the same sort of thing. Usual values for it are `development` and `production`, but nothing stops us from using a `preview` value. Maybe that's worth exploring? It would just be a little more conventional, I don't expect it to have any real material impact.

src/index.tsx
```diff
@@ -53,4 +55,12 @@ if (process.env.NODE_ENV !== 'production') {
...
+  // Start Mirage
+  startMirage();
+}
+
+// Initialize services
+initilizeServices();
```
andrewazores, line 64: If I'm understanding the reasoning behind this `init()` refactor, I think it could be avoided if we ensured that Mirage was always loaded before React was, and therefore before any of the Services could be instantiated. That's similar to what my patch for the live demo on the upstream website did.

I think it could be possible to achieve the same thing here by using some Webpack plugin to check for `process.env.NODE_ENV=='preview'` (or `process.env.PREVIEW=true`) and patch the `src/index.html` to add a `<script>` that loads Mirage. What do you think?

andrewazores: Or perhaps instead of directly patching the `index.html`, using the Webpack entrypoint a little differently might do the job:

https://webpack.js.org/concepts/entry-points/

andrewazores: Or this:
https://webpack.js.org/plugins/html-webpack-plugin/

Either way, I think the goal would be to achieve Mirage loading in fully before React starts, so that the React application code can remain as unchanged as possible. Ideally the application code does not have to make any accommodations for the possibility that Mirage is shimmed in.

src/index.tsx
tthvo: oh yes my thought originally was that we would want to be able to have preview (i.e mirage js) in hot-reload dev server and "production" build (i.e. for PR preview or main preview or upstream cryostatio).

So i use a different environment var to control (i.e. set in webpack). What do u think?

src/index.tsx
tthvo: Ohh yess I actually tried to split the mirage JS chunk out but was worried about about execution order. But from that link above, seems like we can use `dependOn`.

src/index.tsx
tthvo: Tho, from reading this: https://webpack.js.org/guides/production/#specify-the-mode

I think mode: `production` or `NODE_ENV: production` is needed for some tools to work. So, its not really recommended to set it to `preview`?

tthvo:
Updated to split mirage into a separate chunk. Its not obvious that the chunk is executed before `app*.js` but seems to work to minimize changes. Tho, we probably can't set NODE_ENV to preview?

src/index.tsx
tthvo: Latest updates only uses `process.env.PREVIEW` within webpack process to apply mirage bundle. So, no checking in the actual bundle.

andrewazores:
Sweet, this is looking really good.

Are any of the changes in the application bundle still needed for the Mirage shim to work? Or is that just remaining cleanup that was done along the way in the first approach?

andrewazores:
I just tested it out with the two manual test modes you gave and both look great.

If the other refactoring/cleanup changes I mentioned in the last comment aren't required for this PR, please split them out into a separate one. I'd just like to keep the history separated, just in case it turns out that this causes some problems later, it's easier to maintain and back this out or carry it as a separate patch set etc.

tthvo:
> Are any of the changes in the application bundle still needed for the Mirage shim to work? Or is that just remaining cleanup that was done along the way in the first approach?


Oh no. These changes are just eslint fixes and refactoring. I will open a new PR.

tthvo:
Updated to remove eslint fixes and refactoring now :))

Changes required for integrating MirageJS is included along side with some webpack production build optimization:
- Remove comments from bundle and css.
- Remove source maps (only enabled for development mode).

andrewazores (approved):
Very nice, thank you.

PR #1059, opened 2023-07-12
fix(login): submit form to log out from OpenShift OAuth

ebaron:
# Welcome to Cryostat! 👋

Fixes: https://github.com/cryostatio/cryostat/issues/1489

## Description of the change:
On OpenShift, `setLoggedOut` now submits a form to the logout endpoint of OpenShift's OAuth server. This avoids making a CORS request and passes the `HttpOnly` session cookie to the endpoint. I added several new test cases that test logging out with both Basic and Bearer methods.

## Motivation for the change:
This fixes a bug where users on OpenShift could not log out without manually deleting cookies.

## How to manually test:
1. Have a running OpenShift cluster.
2. Build a Cryostat image including this fix (or use mine: `quay.io/ebaron/cryostat:logout-test-web-10`).
3. Deploy the operator using this image: `make deploy CORE_IMG=quay.io/ebaron/cryostat:logout-test-web-10`, substituting the image from step 2 if needed.
4. Log into Cryostat with kubeadmin.
5. Log out.
6. You should be redirected to the login page and any attempts to open Cryostat without logging back in should do the same.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,12 +5,6 @@ module.exports = {
   // Automatically clear mock calls and instances between every test
   clearMocks: true,
 
-  // Automatically reset mock state before every test
-  resetMocks: true,
-
-  // Automatically restore mock state and implementation before every test
-  restoreMocks: true,
-
   // The directory where Jest should output its coverage files
   coverageDirectory: 'coverage',
 
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,6 +5,12 @@ module.exports = {
   // Automatically clear mock calls and instances between every test
   clearMocks: true,
 
+  // Automatically reset mock state before every test
+  resetMocks: true,
+
+  // Automatically restore mock state and implementation before every test
+  restoreMocks: true,
+
   // The directory where Jest should output its coverage files
   coverageDirectory: 'coverage',
 
--- a/src/app/Shared/Services/Login.service.tsx
+++ b/src/app/Shared/Services/Login.service.tsx
@@ -39,7 +39,7 @@ import { Base64 } from 'js-base64';
 import { combineLatest, Observable, ObservableInput, of, ReplaySubject } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
 import { catchError, concatMap, debounceTime, distinctUntilChanged, first, map, tap } from 'rxjs/operators';
-import { ApiV2Response, HttpError } from './Api.service';
+import { ApiV2Response } from './Api.service';
 import { Credential, AuthCredentials } from './AuthCredentials.service';
 import { isQuotaExceededError } from './Report.service';
 import { SettingsService } from './Settings.service';
@@ -133,7 +133,7 @@ export class LoginService {
         return jsonResp.meta.status === 'OK';
       }),
       catchError((e: Error): ObservableInput<boolean> => {
-        window.console.error(JSON.stringify(e));
+        window.console.error(JSON.stringify(e, Object.getOwnPropertyNames(e)));
         this.authMethod.complete();
         return of(false);
       })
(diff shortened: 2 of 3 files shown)


ebaron:
The tests for resetting the auth method currently fail, because it seems the auth method cannot be reset due to being completed. @tthvo @maxcao13 @andrewazores any suggestions for how we could fix this?

maxcao13:
Trying to understand the code, at which point does the `setLoggedOut` function in the Login service reset the AuthMethod back to None? 

I see that the `openshiftLogout` function is called if the method is `BEARER` which eventually should reset the auth method if we get a response, but what about the basic auth?

src/app/Shared/Services/Login.service.tsx
```diff
@@ -197,45 +197,100 @@ export class LoginService {
...
           mode: 'cors',
           method: 'POST',
           body: null,
           headers: this.getAuthHeaders(token, method),
         });
+        return combineLatest([of(token), of(method), resp]);
       }),
-      concatMap((response) => {
-        if (response.status === 302) {
+      concatMap((parts) => {
```
tthvo, line 210: How about using array destructure?

```diff
-      concatMap((parts) => {
-        const token = parts[0];
-        const method = parts[1];
-        const response = parts[2];
-
+      concatMap(([_, method, response]) => {
```

src/app/Shared/Services/Login.service.tsx
```diff
@@ -197,45 +197,100 @@ export class LoginService {
...
+    const resp = fromFetch(`${this.authority}/api/v2.1/auth`, {
+      credentials: 'include',
+      mode: 'cors',
+      method: 'POST',
+      body: null,
+    });
+
+    return resp.pipe(
+      first(),
+      concatMap((response) => {
```
tthvo, line 251: I think we can use `map` and returns `loginUrl` since we don't have to wait to get that value. Same for below (line 269). 

Just that we don't need to create a new inner observable object and subscribe to it.

src/app/Shared/Services/Login.service.tsx
```diff
@@ -247,9 +302,13 @@ export class LoginService {
...
   }
 
-  private navigateToLoginPage(): void {
+  private resetAuthMethod(): void {
     this.authMethod.next(AuthMethod.UNKNOWN);
     this.removeCacheItem(this.AUTH_METHOD_KEY);
+  }
+
+  private navigateToLoginPage(): void {
+    this.resetAuthMethod();
```
tthvo, line 311: How about we call `resetAuthMethod` outside `navigateToLoginPage`, just to be clear:

```ts
     tap(() => {
        this.resetSessionState();
        this.resetAuthMethod();
        this.resetAuthMethod();
      }),
```

tthvo, line 311: How about calling this `resetAuthMethod` outside `navigateToLoginPage`?

```ts
tap(() => {
  this.resetSessionState();
  this.resetAuthMethod();
  this.navigateToLoginPage();
})
```

tthvo:
> The tests for resetting the auth method currently fail, because it seems the auth method cannot be reset due to being completed. @tthvo @maxcao13 @andrewazores any suggestions for how we could fix this?

Hey, we are thinking commenting out this line here will fix the issue:

https://github.com/cryostatio/cryostat-web/blob/13db02830ebfdb6f884d8b71406c648eb3b4202a/src/app/Shared/Services/Login.service.tsx#L293


Perhaps before we wanted to set auth method to a fixed issue. However, later on, the log out workflow was added and this got outdated.

maxcao13:
About the test failure, on line 112 or so, there is a call to `completeAuthMethod` within the `checkAuth` method within Login.service` which completes the `authMethod` subject so its cant emit values anymore. Seems like it gets called everytime since we are checking auth before every test. That's the reason why the tests fail is because the observable completes before we can emit new values.

src/test/Shared/Services/Login.service.test.tsx
```diff
@@ -0,0 +1,442 @@
...
+import { TargetService } from '@app/Shared/Services/Target.service';
+import { AuthCredentials } from '@app/Shared/Services/AuthCredentials.service';
+import { ApiV2Response } from '@app/Shared/Services/Api.service';
+import { SettingsService } from '@app/Shared/Services/Settings.service';
+import { firstValueFrom, of, timeout } from 'rxjs';
+import { fromFetch } from 'rxjs/fetch';
+
+jest.mock('rxjs/fetch', () => {
+  return {
+    fromFetch: jest.fn((url: any, opts: any): any => {}),
```
maxcao13, line 49: To resolve this eslint error, just change to 
```    
fromFetch: jest.fn((url: any, opts: any): any => undefined),
```

Running `yarn eslint:apply` should fix some other warnings to do with dependency import order, but thats just a bit of cleanup.

jest.config.js
```diff
@@ -5,6 +5,12 @@ module.exports = {
   // Automatically clear mock calls and instances between every test
   clearMocks: true,
 
+  // Automatically reset mock state before every test
```
maxcao13, line 8: I don't think we should reset and restore all mocks in the global config because most of other tests in our suite are using mocks. It does break a lot of tests if we run `yarn test`.

Instead I think we can just call `jest.resetAllMocks()` within the first `beforeEach()` to copy the behaviour just to this file. I don't think we need the `restoreAllMocks` because the documentation says it only effects `jest.spyOn` created mocks.

src/test/Shared/Services/Login.service.test.tsx
```diff
@@ -0,0 +1,442 @@
...
+          },
+        });
+        const logoutResp = createResponse(200, true);
+        mockFromFetch
+          .mockReturnValueOnce(of(initAuthResp))
+          .mockReturnValueOnce(of(authResp))
+          .mockReturnValueOnce(of(logoutResp));
+        const token = 'user:d74ff0ee8da3b9806b18c877dbf29bbde50b5bd8e4dad7a3a725000feb82e8f1';
+        window.location.href = 'https://example.com/';
+        location.href = window.location.href;
```
tthvo, line 126: Wondering if we need to set `location.ref`? Is it for completeness when setting `window.location.href`?

jest.config.js
tthvo: Just adding to Max. 

If I understand correctly, it seems the only place that needs resetting is `mockFromFetch`. So, we can call `mockFromFetch.mockReset();`  instead to limit the impact.

tthvo (commented):
Looks good to me! The issue is fixed when running on OCP 4.12.

ebaron:
> > The tests for resetting the auth method currently fail, because it seems the auth method cannot be reset due to being completed. @tthvo @maxcao13 @andrewazores any suggestions for how we could fix this?
> 
> Hey, we are thinking remove this line here will fix the issue:
> 
> https://github.com/cryostatio/cryostat-web/blob/13db02830ebfdb6f884d8b71406c648eb3b4202a/src/app/Shared/Services/Login.service.tsx#L293
> 
> Perhaps previously we wanted to set auth method to a fixed issue when logging in. However, later on, the log out workflow was added and this got outdated since auth method should be reset after logout.
> 
> Then, this means we should rename the function to something like `updateAuthMethod`. And remove the check here since the subject won't be stopped anytime (also it's deprecated to call this check):
> 
> https://github.com/cryostatio/cryostat-web/blob/13db02830ebfdb6f884d8b71406c648eb3b4202a/src/app/Shared/Services/Login.service.tsx#L111
> 
> What do u think?

That would be an easy fix. I was unsure if calling `complete` on the AuthMethod was still used by the code somewhere. If not, I can just remove it.

jest.config.js
ebaron: I'll remove these. I think this was from trying to ensure that LoginService didn't get mocked in my tests. I was able to just use `jest.unmock`, but couldn't figure out why it was being mocked in the first place.

src/test/Shared/Services/Login.service.test.tsx
ebaron: `getTokenFromUrlFragment` and `hasBearerTokenUrlHash` use `location.hash` instead of `window.location.hash`. I wasn't sure if there was a meaningful difference or if they could be changed to `window.location`. Thoughts?

src/test/Shared/Services/Login.service.test.tsx
ebaron: Thanks, I wasn't aware of this!

src/test/Shared/Services/Login.service.test.tsx
tthvo: I think `location` is just another shorthand reference to `window.location.hash`. I tend to use `window.*` for clarity :))

tthvo:
> That would be an easy fix. I was unsure if calling complete on the AuthMethod was still used by the code somewhere. If not, I can just remove it.

I did some digging into the sources:

- No places set the auth method other than the `checkAuth` (Login.service.tsx). 
- Logouts reset the auth method so it requires the stream to still be open.

tthvo:
> Perhaps previously we wanted to set auth method to a fixed issue when logging in. However, later on, the log out workflow was added and this got outdated since auth method should be reset after logout.

The `complete` call was added in the early stage, long before the log out flow here:  https://github.com/cryostatio/cryostat-web/pull/43. Might have been missed.

src/app/Shared/Services/Login.service.tsx
ebaron: Thanks for the suggestion, I'm still getting used to RxJS

ebaron:
> > That would be an easy fix. I was unsure if calling complete on the AuthMethod was still used by the code somewhere. If not, I can just remove it.
> 
> I think `complete` is called in order to prevent subscribers to AuthMethod to be called repeatedly anytime the users log in. This is undesired since AuthMethod won't change for a specific deployment.
> 
> Tho, such repeat is mostly twice (initialization + log in) as logout reloads the whole page with fresh states.
> 
> We can prevent this repeat by using `distinctUntilChanged` in handlers. Also, the stream needs to be open for resetting when logging out (tho, it won't matter as reloads will bring fresh states). So, `complete` should be removed.

I've done this now with `getAuthMethod` in the same manner as `getSessionState`. How does it look?

tthvo (approved):
Looks good to me! Thanks!

jest.config.js
ebaron: Thanks for pointing that out!

maxcao13 (approved):

PR #1069, opened 2023-07-27
feat(match-expression): match expression should be evaluated on backend

tthvo:
# Welcome to Cryostat! 👋

Fixes: #1067 
Depends on https://github.com/cryostatio/cryostat/issues/1601

## Description of the change:

Update front-end handler for async checks when evaluating match expression (i.e. done on server via endpoint `/api/beta/matchExpression`)

## Motivation for the change:

See #1067

### DIFF
--- a/src/app/Rules/CreateRule.tsx
+++ b/src/app/Rules/CreateRule.tsx
@@ -288,20 +288,23 @@ const CreateRuleForm: React.FC<CreateRuleFormProps> = ({ ...props }) => {
     const matchedTargets = matchedTargetsRef.current;
     addSubscription(
       combineLatest([
-        matchExprService.searchExpression(),
+        matchExprService.searchExpression({
+          immediateFn: () => {
+            setEvaluating(true);
+            setMatchExpressionValid(ValidatedOptions.default);
+          },
+        }),
         context.targets.targets().pipe(tap((ts) => setSampleTarget(ts[0]))),
       ])
         .pipe(
-          switchMap(([input, targets]) => {
-            setEvaluating(true);
-            setMatchExpressionValid(ValidatedOptions.default);
-            return input
+          switchMap(([input, targets]) =>
+            input
               ? context.api.matchTargetsWithExpr(input, targets).pipe(
                   map((ts) => [ts, undefined]),
                   catchError((err) => of([[], err]))
                 )
-              : of([undefined, undefined]);
-          })
+              : of([undefined, undefined])
+          )
         )
         .subscribe(([ts, err]) => {
           setEvaluating(false);
--- a/src/app/Rules/CreateRule.tsx
+++ b/src/app/Rules/CreateRule.tsx
@@ -281,7 +281,7 @@ const CreateRuleForm: React.FC<CreateRuleFormProps> = ({ ...props }) => {
         .subscribe((templates) => {
           setTemplates(templates);
           setTemplate((old) => {
-            const matched = templates.find((t) => t.name === old.name && t.type === t.type);
+            const matched = templates.find((t) => t.name === old.name && t.type === old.type);
             return matched ? { name: matched.name, type: matched.type } : {};
           });
         })
@@ -297,12 +297,23 @@ const CreateRuleForm: React.FC<CreateRuleFormProps> = ({ ...props }) => {
       ])
         .pipe(
           switchMap(([input, targets]) =>
-            input ? context.api.matchTargetsWithExpr(input, targets).pipe(catchError((_) => of([]))) : of(undefined)
+            input
+              ? context.api.matchTargetsWithExpr(input, targets).pipe(
+                  map((ts) => [ts, undefined]),
+                  catchError((err) => of([[], err]))
+                )
+              : of([undefined, undefined])
           )
         )
-        .subscribe((ts) => {
+        .subscribe(([ts, err]) => {
           setMatchExpressionValid(
-            !ts ? ValidatedOptions.default : ts.length ? ValidatedOptions.success : ValidatedOptions.warning
+            err
+              ? ValidatedOptions.error
+              : !ts
+              ? ValidatedOptions.default
+              : ts.length
+              ? ValidatedOptions.success
+              : ValidatedOptions.warning
           );
           matchedTargets.next(ts || []);
         })
--- a/src/app/Rules/CreateRule.tsx
+++ b/src/app/Rules/CreateRule.tsx
@@ -45,7 +45,7 @@ import { SelectTemplateSelectorForm } from '@app/Shared/SelectTemplateSelectorFo
 import { TemplateType } from '@app/Shared/Services/Api.service';
 import { ServiceContext } from '@app/Shared/Services/Services';
 import { Target } from '@app/Shared/Services/Target.service';
-import { SearchExprService, SearchExprServiceContext } from '@app/Topology/Shared/utils';
+import { SearchExprService, SearchExprServiceContext, useExprSvc } from '@app/Topology/Shared/utils';
 import { useSubscriptions } from '@app/utils/useSubscriptions';
 import { portalRoot } from '@app/utils/utils';
 import {
@@ -75,7 +75,7 @@ import _ from 'lodash';
 import * as React from 'react';
 import { useHistory, withRouter } from 'react-router-dom';
 import { forkJoin, iif, of, Subject } from 'rxjs';
-import { catchError, debounceTime, map, switchMap } from 'rxjs/operators';
+import { catchError, concatMap, debounceTime, map, switchMap } from 'rxjs/operators';
 import { Rule } from './Rules';
 
 // FIXME check if this is correct/matches backend name validation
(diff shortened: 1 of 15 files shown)


src/app/Rules/CreateRule.tsx
```diff
@@ -286,26 +286,20 @@ const CreateRuleForm: React.FC<CreateRuleFormProps> = ({ ...props }) => {
...
+          error: (_) => {
+            setMatchExpressionValid(ValidatedOptions.error);
+          },
+        })
+      );
     }
-    setMatchExpressionValid(validation);
-    matchedTargets.next(matches);
-  }, [matchExpression, targets, matchedTargets, setMatchExpressionValid]);
+  }, [matchExpression, targets, matchedTargets, context.api, setMatchExpressionValid, addSubscription]);
```
andrewazores, line 302: This same comment applies to multiple similar changes following the same pattern -

This looks like it will fire an API request on every keystroke updating the matchExpression. If that's the case, I think there should be a `debounceTime` or similar RxJS operator applied so that the request is only fired after the user has stopped typing for some time.

tthvo: Oh yesss! There is actually a match expression hook with `debountime`. I will update all components to use this instead.

tthvo: Updated now. Tho, some will use the expression service directly so avoid calling unnecessary effect hook. Tho, we will need to use `switchMap` to ignore previous requests too.

Need a little more refactoring :))

tthvo:
Remaining tasks:

- [ ] Use `switchMap` to cancel/ignore previous API request. Seems like using the expression hook might not be needed anymore. Just use the expression service directly.
- [ ] Fix unit tests. Add more tests.
- [ ] Update API request (/beta/matchExpression`) body to include JSON instead of multi-part. Depends on https://github.com/cryostatio/cryostat/issues/1601

tthvo:
Ready for review now! 

There are a lot more changes than expected but I tried my best to minimize the number of API requests to `/matchExpressions` endpoint.

One fix that might raise questions is in the stored credential table. When a target appears or disappears, I opted to just refresh the entire table (1 API request) instead of each match expression checking themselves (a lot API requests and React reducer is not fit for async handling).

src/app/Topology/Shared/utils.tsx
```diff
@@ -230,8 +232,8 @@ export const isTargetNodeFiltered = ({ target }: TargetNode, filters?: TopologyF
 export class SearchExprService {
   private readonly _state$ = new BehaviorSubject<string>('');
 
-  searchExpression(): Observable<string> {
-    return this._state$.asObservable();
+  searchExpression(debounceMs = 0): Observable<string> {
```
andrewazores, line 235: I think it'd be better to use the `DEFAULT_MATCH_EXPR_DEBOUNCE_TIME` as the, well, default parameter value here. Seems like it would be more expected as a developer/maintainer coming along down the line and calling this function.

src/app/Topology/Shared/utils.tsx
```diff
@@ -241,33 +243,33 @@ export class SearchExprService {
 
 export const SearchExprServiceContext = React.createContext(new SearchExprService());
 
-export const useSearchExpression = (debounceMs = 0): [string, (expr: string) => void] => {
-  const [expr, setExpr] = React.useState('');
-  const exprSvc = React.useContext(SearchExprServiceContext);
-  const _subRef = React.useRef<Subscription>();
+export const useExprSvc = (): SearchExprService => React.useContext(SearchExprServiceContext);
+
+export const DEFAULT_MATCH_EXPR_DEBOUNCE_TIME = 250; // ms
```
andrewazores, line 248: Maybe I just type a little too slowly but in manually testing this with my browser's devtools open, I find that quite often I haven't actually _stopped_ typing but am just pausing or quickly proofreading what I've written, and the API request fires right as I am about to resume writing. There's a balancing act here between input latency and number of requests for sure, but I think this default could be longer. I don't think it's really worth making this into a configurable setting though.

tthvo: sure! I am actually reading and picking idea from here: https://stackoverflow.com/questions/42361485/how-long-should-you-debounce-text-input.

tthvo: Any value you have in mind? I am thinking 275-300ms?

andrewazores: Sure, 300ms sounds fine. I was thinking maybe 350ms before reading through that. The current test cases we're looking at are going to have extremely low network latency compared to realistic live deployments so it's going to feel a little slower in the real world anyway. I just am also mindful that real deployments may possibly have many more targets to test against than our test scenarios, so that's where the balancing act comes in.

src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx
```diff
@@ -140,53 +145,60 @@ interface AuthFormProps extends Omit<CreateCredentialModalProps, 'visible'> {
...
-        validation = atLeastOne ? ValidatedOptions.success : ValidatedOptions.warning;
-      } catch (err) {
-        validation = ValidatedOptions.error;
-      }
-    }
-    setMatchExpressionValid(validation);
-  }, [matchExpression, targets, setMatchExpressionValid]);
+    addSubscription(
+      combineLatest([
+        matchExprService.searchExpression(DEFAULT_MATCH_EXPR_DEBOUNCE_TIME),
```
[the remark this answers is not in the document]
tthvo: I tried something like below where a "Checking..." text is shown. However, its a little too quick to see.

```diff
diff --git a/src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx b/src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx
index fc5be089..843ff9cc 100644
--- a/src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx
+++ b/src/app/SecurityPanel/Credentials/CreateCredentialModal.tsx
@@ -152,6 +152,7 @@ export const AuthForm: React.FC<AuthFormProps> = ({ onDismiss, onPropsSave, prog
   const testPool = React.useContext(TestPoolContext);
   const [saving, setSaving] = React.useState(false);
   const [isDisabled, setIsDisabled] = React.useState(false);
+  const [checkExpr, setCheckExpr] = React.useState(false);
 
   const [sampleTarget, setSampleTarget] = React.useState<Target>();
 
@@ -177,16 +178,18 @@ export const AuthForm: React.FC<AuthFormProps> = ({ onDismiss, onPropsSave, prog
         context.targets.targets().pipe(tap((ts) => setSampleTarget(ts[0]))),
       ])
         .pipe(
-          switchMap(([input, targets]) =>
-            input
+          switchMap(([input, targets]) => {
+            setCheckExpr(true);
+            return input
               ? context.api.matchTargetsWithExpr(input, targets).pipe(
                   map((ts) => [ts, undefined]),
                   catchError((err) => of([[], err]))
                 )
-              : of([undefined, undefined])
-          )
+              : of([undefined, undefined]);
+          })
         )
         .subscribe(([ts, err]) => {
+          setCheckExpr(false);
           setMatchExpressionValid(
             err
               ? ValidatedOptions.error
@@ -198,7 +201,15 @@ export const AuthForm: React.FC<AuthFormProps> = ({ onDismiss, onPropsSave, prog
           );
         })
     );
-  }, [matchExprService, context.api, context.targets, setSampleTarget, setMatchExpressionValid, addSubscription]);
+  }, [
+    matchExprService,
+    context.api,
+    context.targets,
+    setSampleTarget,
+    setMatchExpressionValid,
+    addSubscription,
+    setCheckExpr,
+  ]);
 
   React.useEffect(() => {
     progressChange && progressChange(saving);
@@ -254,7 +265,9 @@ export const AuthForm: React.FC<AuthFormProps> = ({ onDismiss, onPropsSave, prog
         isRequired
         fieldId="match-expression"
         helperText={
-          matchExpressionValid === ValidatedOptions.warning
+          checkExpr
+            ? 'Evaluating Match Expression...'
+            : matchExpressionValid === ValidatedOptions.warning
             ? `Warning: Match expression matches no targets.`
             : `
         Enter a match expression. This is a Java-like code snippet that is evaluated against each target

```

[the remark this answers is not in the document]
tthvo: Screenshot from 2023-08-01 12-11-26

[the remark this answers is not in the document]
tthvo: Is this good or do u think we can wait for PF5?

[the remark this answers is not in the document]
tthvo: Maybe a separate "good_first_issue" for investigating?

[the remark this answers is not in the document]
tthvo: sounds good! Updated just now to include the loading helper text.

src/app/Rules/CreateRule.tsx
```diff
@@ -274,38 +277,55 @@ const CreateRuleForm: React.FC<CreateRuleFormProps> = ({ ...props }) => {
...
-  }, [matchExpression, targets, matchedTargets, setMatchExpressionValid]);
+    const matchedTargets = matchedTargetsRef.current;
+    addSubscription(
+      combineLatest([
+        matchExprService.searchExpression(),
+        context.targets.targets().pipe(tap((ts) => setSampleTarget(ts[0]))),
+      ])
+        .pipe(
+          switchMap(([input, targets]) => {
+            setEvaluating(true);
```
andrewazores, line 296: Can this state get set as soon as the text input element sees a keystroke, rather than getting set when the API request is about to be fired? ie. it shouldn't be subject to the debounce time

tthvo: Oh yeh it should be possible.

Just that `matchExprService.searchExpression()` need refactoring to remove internal default debouceTime and each consumer needs to specify there own.

tthvo: ah mvnind found a better way :))

tthvo: Updated :))

andrewazores (approved):

PR #1113, opened 2023-09-18
build(ci): submodule should sign commits

aali309:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Fixes: #916 

## Description of the change:
This change will sign all commits on `git submodule update`

## Motivation for the change:
All commits should be verified.

## How to manually test:
1. *Run CRYOSTAT_IMAGE=quay.io... sh smoketest.sh...*
2. *...*

PR #1124, opened 2023-09-29
ci(submodule): remove bot user git configurations

tthvo:
# Welcome to Cryostat! 👋

Related to #1113 

## Description of the change:

Remove bot user git configurations to allow the signed commits to be verified.

## Motivation for the change:

https://github.com/cryostatio/cryostat-web/pull/1113#issuecomment-1738896699

### DIFF
--- a/.github/workflows/submodule.yaml
+++ b/.github/workflows/submodule.yaml
@@ -41,10 +41,7 @@ jobs:
         git submodule update --init
         git submodule update --remote
     - name: Commit and push submodule
-      # User info comes from: https://github.community/t/github-actions-bot-email-address/17204/5
       run: |
-        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
-        git config user.name "github-actions[bot]"
         git add --all
         git_hash="$(git rev-parse --short :web-client)"
         git commit -S -m "build(web-client): update submodule to $git_hash" || echo "No changes to commit"


andrewazores (approved):

andrewazores:
https://github.com/cryostatio/cryostat-web/actions/runs/6380683359

andrewazores:
https://github.com/cryostatio/cryostat/commits/main

image

Not verified, but the new bot name is nice at least.

tthvo:
Hmm, I wonder if its because of the this action:

https://github.com/crazy-max/ghaction-import-gpg#inputs

Seems like we automatically add the author info from there:

https://github.com/cryostatio/cryostat-web/blob/a019a0cd4be00633d646c20a8646e0199c3cd2df/.github/workflows/submodule.yaml#L37

tthvo:
Or maybe since this is a cross-repo action, is the gpg key added to organization level as a signing key?

tthvo:
@andrewazores Might not be the issue if github see the same thing. Tho, I have 2 questions:

- Since this is a cross-repo action, does cryostat repo need to know about this key? We should add this key to cryostatio organization?
- Perhaps, the email could be one used for cryostatio? This way, github can recognize it. Otherwise, `rhjavamonitoring@gmail.com` is unknown.

andrewazores:
I'm not aware of a way to add the key or an email address to the org itself like that, but when I have some more free time to look into this I'll see if I can find anything. Normally those are things that are added to individual users' profiles, not orgs...

tthvo:
Ah ur right! How about this?

https://github.com/Nautilus-Cyberneering/pygithub/blob/main/docs/how_to_sign_automatic_commits_in_github_actions.md#solution-02-using-your-own-pgp-key-as-a-secret


Basically,we need to create a bot account (new GitHub account) with name Cryostat CI and the rhjavamonitoring@gmail.com. Then, add the public gpg key to that account.

tthvo:
Otherwise, following above pattern, I think you can add the gpg public key to your github account and manually set git configurations in this submodule workflow to point to https://github.com/andrewazores.

The commit will show up as you, I believe.

andrewazores:
I was thinking we might need a separate GitHub bot account for this, but I've been trying to avoid having to deal with that. It might just have to be though, if we do want those commits to show up as Verified.

If we're going to bother with signature verification I would rather it's a generic bot account rather than showing up as any particular one of our own accounts.

Issue #1110, opened 2023-09-18
[Epic] Upgrade to Patternfly v5


andrewazores:
### Describe the feature

Patternfly v5 is out: https://www.patternfly.org/

Cryostat Web is currently using v4: http://v4-archive.patternfly.org/v4/

Upgrade doc: https://www.patternfly.org/get-started/upgrade/

PR #1143, opened 2023-10-21
build(deps): upgrade react-router to v6

tthvo:
# Welcome to Cryostat! 👋

Related to #482 
Related to #1110 
Continue 726 

## Description of the change:

WIP

## Motivation for the change:

WIP


src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -187,8 +185,8 @@ export const ActiveRecordingsTable: React.FC<ActiveRecordingsTableProps> = (prop
   );
 
   const handleCreateRecording = React.useCallback(() => {
-    routerHistory.push(`${url}/create`);
-  }, [routerHistory, url]);
+    navigate('/recordings/create');
```
andrewazores, line 188: I would rather keep something more similar to the previous logic where the navigation to the subview is basically just a relative path resolution, and the component doesn't need to know its own absolute path. This looks promising:

https://stackoverflow.com/questions/72683709/react-router-v6-useroutematch-equivalent

src/app/Shared/Components/FileUploads.tsx
```diff
@@ -258,7 +257,11 @@ export const MultiFileUpload: React.FC<MultiFileUploadProps> = ({
 
   return (
     <>
-      <Prompt when={uploading} message="Are you sure you wish to cancel the file upload?" />
+      {/* 
+        TODO: Add back when supported
```
andrewazores, line 261: Can we work around this in the meantime?

https://stackoverflow.com/questions/70178993/alternative-for-prompt-in-react-router-v6

src/app/Recordings/ActiveRecordingsTable.tsx
andrewazores: Or even just https://reactrouter.com/en/main/hooks/use-navigate#optionsrelative

src/app/Rules/CreateRule.tsx
```diff
@@ -205,7 +205,7 @@ export const CreateRuleForm: React.FC<CreateRuleFormProps> = (_props) => {
     [setFormData],
   );
 
-  const exitForm = React.useCallback(() => history.push('/rules'), [history]);
+  const exitForm = React.useCallback(() => navigate('/rules'), [navigate]);
```
andrewazores, line 208: This could probably also be a `navigate('..', { relative: 'path' })`, and there are a couple of other places where relative navigation looks like it would make sense - most places that use `navigate()` would probably be worth re-evaluating.

src/app/Recordings/ActiveRecordingsTable.tsx
tthvo: Ahh that slipped my mind again. Updated to use relative path now.

src/app/Shared/Components/FileUploads.tsx
tthvo: Hmm unfortunately, that requires the `Prompt` to be under a `Router`. 

```
useBlocker must be used within a data router.  See https://reactrouter.com/routers/picking-a-router.
    at invariant (webpack-internal:///12599:280:11)
    at useDataRouterContext (webpack-internal:///89250:672:85)
    at useBlocker (webpack-internal:///89250:814:7)
    at Prompt (webpack-internal:///5438:29:74)
    at renderWithHooks (webpack-internal:///73116:14985:18)
    at mountIndeterminateComponent (webpack-internal:///73116:17811:13)
    at beginWork (webpack-internal:///73116:19049:16)
    at HTMLUnknownElement.callCallback (webpack-internal:///73116:3945:14)
    at Object.invokeGuardedCallbackDev (webpack-internal:///73116:3994:16)
    at invokeGuardedCallback (webpack-internal:///73116:4056:31)
```

Right now, this `FileUploads` are used on modals, which are mounted on a node `rootPortal` outside root node.

https://github.com/cryostatio/cryostat-web/blob/316e9c7cb8c27b125a23a80b95e3e0fd25faa9c1/src/index.html#L30

I believe it was done that way to ensure modals are always on top.

tthvo: We can exploring movign `portal-root` down to below `Router` but I am not sure how that gonna break modals. Would u prefer to have it as a separate issue for more research?

src/app/Rules/CreateRule.tsx
tthvo: Updated all "exitForm" to go to `..` path.

src/app/Shared/Components/FileUploads.tsx
tthvo: Ahh it needs the `createBrowserRouter` and `RouterProvider` way to be able to use.

https://reactrouter.com/en/main/routers/picking-a-router#using-v64-data-apis

andrewazores: Okay, let's create a separate issue to track it. I don't think this is that big of a loss to spend a ton of time on or to block the version upgrade until this is fixed.

andrewazores: #1147

andrewazores:
This is interesting:

image

It seems like now, if I kill the `smoketest.sh` server (presumably any Cryostat server instance) and the websocket connection to the client drops, then most views become 404s rather than showing the basic auth login. About and Settings still work, and Dashboard brings me to login, but everything else in the navigation drawer becomes a 404 when the client cannot connect to the server.

The same thing happens if I log out in the top-right user menu - most views become 404s, but I expected this instead to just display the login view everywhere instead.

I haven't tried this in an OpenShift setup to see how that auth works in a similar case.

tthvo:
Ahh I think that I added 

```
<Route key={'not-found'} path={'*'} element={<PageNotFound />} />
```
 to fall back to not-found. That would mean logging out will match only this path. Ahh, I will have a look again.

andrewazores:
I guess one possibility would be to redirect the user to `/` (Dashboard) before performing the actual logout call. The user would lose their state of whatever view they were on before logging out, but I don't think that would be too surprising.

tthvo:
Sure! How about:

```
<Route key={'not-found'} path={'*'} element={loggedIn? <PageNotFound />: <Login />} />
```

I would use Login when logging out. Haven't tested it yet but seems reasonable.

andrewazores:
Seems to work in `smoketest.sh` at least - there's a weird white flash on logout before coming back to the login screen, but I'm not too worried about that. I will try this in OpenShift as well.

andrewazores:
Worked well enough in OpenShift. There is also a white screen flash but I don't think it's a big deal. Also, I use everything in dark mode, which really emphasizes it.

andrewazores (approved):

PR #1155, opened 2023-11-01
chore(strings): update 2.4 user-facing strings

mwangggg:
# Welcome to Cryostat! 👋

Related to: https://github.com/cryostatio/cryostat/issues/1695

### DIFF
--- a/locales/en/public.json
+++ b/locales/en/public.json
@@ -258,9 +258,9 @@
     "DESCRIPTION": "Select Dashboard Layout Template configuration file(s) to upload. File(s) must be in valid JSON format.",
     "ERROR": {
       "CONFIG_INVALID": "Card configurations are invalid.",
-      "DESCRIPTION_INVALID": "Template description is invalid.",
+      "DESCRIPTION_INVALID": "Template description is invalid. The description can only include letters, numbers, spaces, and some special characters. It must also be 100 characters or less.",
       "DUPLICATE_UPLOAD": "A previously uploaded file has the same template name as the current file: \"{{name}}\".",
-      "NAME_INVALID": "Template name is invalid.",
+      "NAME_INVALID": "Template name is invalid. The name must be alphanumeric and be 20 characters or less.",
       "NAME_TAKEN": "Template with name \"{{name}}\" already exists.",
       "PARSE": "JSON parse error.",
       "TEMPLATE_INVALID": "Template has an invalid configuration.",
--- a/locales/en/public.json
+++ b/locales/en/public.json
@@ -258,7 +258,7 @@
     "DESCRIPTION": "Select Dashboard Layout Template configuration file(s) to upload. File(s) must be in valid JSON format.",
     "ERROR": {
       "CONFIG_INVALID": "Card configurations are invalid.",
-      "DESCRIPTION_INVALID": "Template description is invalid. The description can only include letters, numbers, spaces, and some special characters. It must also be 100 characters or less.",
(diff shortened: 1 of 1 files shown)


locales/en/public.json
```diff
@@ -258,9 +258,9 @@
     "DESCRIPTION": "Select Dashboard Layout Template configuration file(s) to upload. File(s) must be in valid JSON format.",
     "ERROR": {
       "CONFIG_INVALID": "Card configurations are invalid.",
-      "DESCRIPTION_INVALID": "Template description is invalid.",
+      "DESCRIPTION_INVALID": "Template description is invalid. The description can only include letters, numbers, spaces, and some special characters. It must also be 100 characters or less.",
```
andrewazores, line 261: Maybe give at least a few examples of which special characters are allowed?

mwangggg: period (.), comma (,), hyphen (-), apostrophe ('), semicolon (;), double quote ("), exclamation mark (!), question mark (?), at symbol (@), number sign (#), dollar sign ($), percent sign (%), caret (^), ampersand (&), asterisk (*), left parenthesis ((), right parenthesis ()), left square bracket ([), right square bracket (]), underscore (_), plus sign (+), colon (:), left curly brace ({), right curly brace (}).

mwangggg: ^ that's the list of allowed special characters. I'm not really sure how to shorten it? I guess some are more common like periods and commas. Which do you think I should list?

andrewazores: "letters, numbers, spaces, quotations, punctuation, and some other special characters such as underscores and brackets"

At least `. , ! ? ; - ' "` would be covered by "quotations and punctuation". `[ ] ( ) { }` could all be considered "brackets".

If the list of allowed characters is that extensive, I wonder why we even do it. Notably absent from this list of allowed characters are things like `/ \`, things that could be taken as filesystem path separators. Maybe there's a deeper issue here where this should be updated to reject (or better: sanitize) these particular characters on a blocklist, instead of using this wide-ish allowed list?

Because other notably excluded characters in this set include anything that is a letter in any alphabet other than English.

mwangggg: `/^[a-zA-Z0-9\s.,\-'";?!@#$%^&*()[\]_+=:{}]*$/`. Here's the regex of characters allowed in the description. Maybe there's a way to check against a list of characters that shouldn't be in the description? Although this list could be much longer if we include anything that isn't English?

andrewazores: I don't remember much about reviewing this feature originally, but I'm curious why there is any restriction on the character set allowed for this description text anyway. Especially because this is for the dashboard layout templates, which only get stored in the browser's localstorage - the server never even sees them, so there's absolutely no risk that somehow these get interpreted and special characters get used to do nefarious things.

The deepest question here is "what makes a template description invalid, and why?" It seems to me that even a length restriction is arbitrary and unnecessary unless there are technical constraints on how large browsers allow localstorage entries to be. Restricting the character set seems completely unnecessary. At worst, the client code should encode the user's entered text before storing it, then decode it again before displaying it, if there is some concern about the text being interpreted incorrectly by the client application/browser - for example, if the user enters text that can be interpreted as HTML.

Anyway - that's getting off on a tangent and out of scope of what you set out to do here. Your updated text is at least an improvement on the current text, so that's fine. I think it would be worth filing a bug about the restrictions on this text field however, and when those restrictions are lifted, these texts would need to be updated again.

andrewazores (approved):

andrewazores:
Do you have any other strings to update, and if so do you want to include them in this PR or separately?

mwangggg:
These are the last two strings that haven't been updated from the 2.3 spreadsheet- I'll update others in a separate PR after the documentation team looks at them

andrewazores (approved):

PR #1164, opened 2023-11-16
chore(strings): update user string

mwangggg:
# Welcome to Cryostat! 👋

Related: https://github.com/cryostatio/cryostat/issues/1695

### DIFF
--- a/locales/en/public.json
+++ b/locales/en/public.json
@@ -79,7 +79,7 @@
     "MAXIMUM_SIZE": "Maximum size ({{unit}})",
     "SAVE_CHANGES": "Save changes",
     "TEMPLATE_HELPER_TEXT": "The Event Template to be applied to Automated Analysis recordings.",
-    "TEMPLATE_INVALID_WARNING": "WARNING: Setting a Target Template as a default template type configuration may not apply to all Target JVMs if the JVMs do not support them."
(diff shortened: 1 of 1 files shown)


locales/en/public.json
```diff
@@ -79,7 +79,7 @@
     "MAXIMUM_SIZE": "Maximum size ({{unit}})",
     "SAVE_CHANGES": "Save changes",
     "TEMPLATE_HELPER_TEXT": "The Event Template to be applied to Automated Analysis recordings.",
-    "TEMPLATE_INVALID_WARNING": "WARNING: Setting a Target Template as a default template type configuration may not apply to all Target JVMs if the JVMs do not support them."
+    "TEMPLATE_INVALID_WARNING": "WARNING: Setting the Target Template as a default template type configuration may not apply to all Target JVMs if the JVMs do not support them."
```
andrewazores, line 82: I'm not sure about this change. It's describing event templates of the kind "Target", as opposed to "Custom":

image

There are usually at least two such Target Templates for any given target application.

This warning is to say that if you pick a Target Template, then this only makes sense if every other Target has that same kind of Template (matching by name).

To give a concrete example of such an invalid configuration:

image

I could select the Cryostat event template here, but any configuration using this template (Automated Analysis card, Automated Rule) would not be able to apply this event template to another target which doesn't have that same template:

image

mwangggg: Ok I'll keep it how it is :+1:

PR #1171, opened 2023-12-05
chore(ci): request necessary permissions

aali309:
# Welcome to Cryostat! 👋

Fixes: #1172 
Related to: [#1783](https://github.com/cryostatio/cryostat/issues/1783)

## Description of the change:
*This change adds allows a match expression example to be copied to the clipboard...*

## Motivation for the change:

## How to manually test:
2. *...*

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,8 +20,6 @@ jobs:
     strategy:
       matrix:
         node-version: [16.x, 18.x]
-    permissions:
-      pull-requests: read
     steps:
     - uses: actions/checkout@v3
     - name: Use Node.js ${{ matrix.node-version }}
@@ -36,8 +34,6 @@ jobs:
     strategy:
       matrix:
         node-version: [16.x, 18.x]
-    permissions:
-      pull-requests: read
     steps:
     - uses: actions/checkout@v3
     - name: Use Node.js ${{ matrix.node-version }}
@@ -66,10 +62,6 @@ jobs:
     strategy:
       matrix:
         node-version: [16.x, 18.x]
-    permissions:
-      statuses: write
-      pull-requests: write
-      packages: read
     steps:
     - uses: actions/checkout@v3
     - name: Use Node.js ${{ matrix.node-version }}
@@ -84,10 +76,6 @@ jobs:
     strategy:
       matrix:
         node-version: [16.x, 18.x]
-    permissions:
-      statuses: write
-      pull-requests: write
-      packages: read
     steps:
     - uses: actions/checkout@v3
     - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yaml
```diff
@@ -76,6 +84,10 @@ jobs:
     strategy:
       matrix:
         node-version: [16.x, 18.x]
+    permissions:
+      statuses: write
+      pull-requests: write
+      packages: read
```
mwangggg, line 90: these should all be unnecessary? I don't think we're adding anything to the status checks or commenting on the PR in this job? Also the base permissions of the org should have `packages: read` by default

.github/workflows/ci.yaml
```diff
@@ -62,6 +66,10 @@ jobs:
...
```
mwangggg, line 72: same here as above

.github/workflows/ci.yaml
```diff
@@ -34,6 +36,8 @@ jobs:
     strategy:
       matrix:
         node-version: [16.x, 18.x]
+    permissions:
+      pull-requests: read
```
mwangggg, line 40: is this necessary? I don't see any step that requires this permission

.github/workflows/ci.yaml
```diff
@@ -20,6 +20,8 @@ jobs:
...
```
mwangggg, line 24: same here as above.. unless I'm missing something?

.github/workflows/dependent-issues.yml
```diff
@@ -20,6 +20,10 @@ on:
 
 jobs:
   check:
+    permissions:
+      issues: write
+      pull-requests: write
+      statuses: write
```
mwangggg, line 26: I believe these are all unnecessary too? should be fine with just the default permissions we have

aali309: [https://github.com/cryostatio/cryostat-web/actions/runs/6952595800/job/18916412442?pr=1153](https://github.com/cryostatio/cryostat-web/actions/runs/6952595800/job/18916412442?pr=1153)
I thought this was a permission issue?

mwangggg:
yes yes that was me being silly

.github/workflows/semantic-pr.yml
```diff
@@ -11,6 +11,9 @@ on:
 jobs:
   main:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      statuses: write
```
mwangggg, line 16: no need for statuses

.github/workflows/dependent-issues.yml
mwangggg: I don't think we need statuses here...

mwangggg: nvm I'm silly

mwangggg (approved):
LGTM :)

andrewazores (approved):

Issue #1120, opened 2023-09-28
[Bug] Raw HTTP error response displayed when uploading actions fail


andrewazores:
### Current Behavior

When ex. uploading a JMX SSL cert, if the upload fails for any reason, the notification popup displayed shows the raw HTTP response JSON payload.

### Expected Behavior

The error message should be nicely formatted, and more concise/informative.

### Steps To Reproduce

1. run smoketest.sh
2. go to Security view
3. try to upload an SSL cert more than once, ex. https://github.com/andrewazores/vertx-fib-demo/blob/master/src/main/extras/app/resources/vertx-fib-demo.cer
4. go to Automated Rules
5. create a rule
6. download the rule
7. delete the rule
8. try to upload the same rule twice

### Anything else?

![image](https://github.com/cryostatio/cryostat-web/assets/3787464/e5fe741b-b299-4e3b-b1a1-47615e6ebdfd)

![image](https://github.com/cryostatio/cryostat-web/assets/3787464/ca7e90c5-eca5-4de6-8574-4a3d21274fef)

andrewazores:
Upload requests go through here:

https://github.com/cryostatio/cryostat-web/blob/bdcdfb5f122da27a81ba107c77e685c0518c6f16/src/app/Shared/Services/Api.service.tsx#L1408

Error handling happens here:

https://github.com/cryostatio/cryostat-web/blob/bdcdfb5f122da27a81ba107c77e685c0518c6f16/src/app/Shared/Services/Api.service.tsx#L1504

Since this is not a WebSocket notification like most other graphical notifications that appear, we would need callers to pass in more details for notifications like the title that should appear on the notification. We could try to parse the `detail` body as JSON and use the `detail.data.reason` as the message to display, and if that fails simply show `detail` directly.

PR #1192, opened 2023-12-24
fix(notifications): parse detail for notification body

mwangggg:
# Welcome to Cryostat! 👋

Fixes: #1120
Depends on: https://github.com/cryostatio/cryostat3/pull/260

### DIFF
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -226,7 +226,7 @@ export class ApiService {
 
     const headers = {};
     headers['Content-Type'] = 'application/json';
-    return this.sendLegacyRequest('v2', 'rules', {
+    return this.sendLegacyRequest('v2', 'rules', 'Rule Upload Failed', {
       method: 'POST',
       body: JSON.stringify(rule),
       headers: headers,
@@ -572,7 +572,7 @@ export class ApiService {
 
     const body = new window.FormData();
     body.append('template', file);
-    return this.sendLegacyRequest('v1', 'templates', {
+    return this.sendLegacyRequest('v1', 'templates', 'Template Upload Failed', {
       body: body,
       method: 'POST',
       headers: {},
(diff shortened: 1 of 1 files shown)


src/app/Shared/Services/Api.service.tsx
```diff
@@ -1486,7 +1488,12 @@ export class ApiService {
       } else {
         Promise.resolve(error.xmlHttpResponse.body as string).then((detail) => {
           if (!suppressNotifications) {
-            this.notifications.danger(`Request failed (${error.xmlHttpResponse.status} ${error.message})`, detail);
+            try {
+              const body = JSON.parse(detail).data.reason;
+              this.notifications.danger(`${title} Upload Failed`, body);
```
andrewazores, line 1493: I suppose these failures won't strictly be because of "Upload Failed". It would probably be better to just do ex. `this.notifications.danger(title, body)` here, and append "Upload Failed" to each case where that is relevant. Right now that might happen to be every case, but it doesn't seem to necessarily apply in this very generic error handling function.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -1450,7 +1452,12 @@ export class ApiService {
       } else {
         Promise.resolve(error.xmlHttpResponse.body as string).then((detail) => {
           if (!suppressNotifications) {
-            this.notifications.danger(`Request failed (${error.xmlHttpResponse.status} ${error.message})`, detail);
+            try {
+              const body = JSON.parse(detail).data.reason;
+              this.notifications.danger(`${title}`, body);
```
andrewazores, line 1457: I think ```${title}``` can just be `title` - pass the string variable directly instead of interpolating it inside of another string.

andrewazores (approved):

andrewazores:
CI failure due to -core not being buildable because it's still trying to pull and build an old version that has the broken `jfrog` repository configuration.

I'm trying to get -core 2.28.0 onto Maven Central now. If/when that works and we update all our projects to that dependency version then we should be finally out of the woods with this GitHub Packages Maven mess and not need authentication to download the dependency, or need to rebuild it like this in other project dependency chains, etc.

andrewazores:
Should be OK after https://github.com/cryostatio/cryostat3/pull/279 : https://repo.maven.apache.org/maven2/io/cryostat/cryostat-core/

andrewazores:
https://github.com/cryostatio/cryostat3/pull/279

andrewazores (approved):

PR #1211, opened 2024-02-04
fix(rules): remove id key from uploaded rule

mwangggg:
# Welcome to Cryostat! 👋

Fixes: https://github.com/cryostatio/cryostat3/issues/261

### DIFF
--- a/src/app/Rules/RulesUploadModal.tsx
+++ b/src/app/Rules/RulesUploadModal.tsx
@@ -35,6 +35,9 @@ export const parseRule = (file: File): Observable<Rule> => {
   return from(
     file.text().then((content) => {
       const obj = JSON.parse(content);
+      if (obj.id) {
+        obj.remove('id')
+      }
       if (isRule(obj)) {
         return obj;
       } else {
--- a/src/app/Rules/RulesUploadModal.tsx
+++ b/src/app/Rules/RulesUploadModal.tsx
@@ -36,7 +36,7 @@ export const parseRule = (file: File): Observable<Rule> => {
     file.text().then((content) => {
       const obj = JSON.parse(content);
       if (obj.id) {
-        obj.remove('id')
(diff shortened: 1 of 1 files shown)


andrewazores:
I think this should really be done on the server side, and this is actually a broader change that affects more than just rule definitions.

https://github.com/cryostatio/cryostat3/blob/bd6f61d05436f26fcbf8d3edd27e767be21dc8c9/src/main/java/io/cryostat/discovery/CustomDiscovery.java#L101

Here we have an example of a POST endpoint that expects a form submission (URL-encoded or multipart) for creating a custom target. There is no `@RestForm` for the entity's `id`, so there is no way for a client to hit this endpoint and supply an ID that the server will try to (re)use. This is good - the ID should only ever be assigned by the server, or really by the database. In any case the client should not be allowed to specify an ID for a created entity.

https://github.com/cryostatio/cryostat3/blob/bd6f61d05436f26fcbf8d3edd27e767be21dc8c9/src/main/java/io/cryostat/discovery/CustomDiscovery.java#L90

Here we have a problem. This is an equivalent endpoint that expects a JSON payload body. This means the client is free to submit a JSON that includes `{ "id": 1234 }` and the API definition accepts it. This will be deserialized into a `Target` instance that really has that field with that value set. The server will then try to persist that into the database. This should not be allowed. Either the server should silently scrub and ignore the `id` field and allow it to be assigned by the database, or the server should refuse the request outright and respond with a 400/409 status before it even gets to the database.

andrewazores (approved):
Looks good. Please double check if there are any other UI workflows that perform a JSON POST like this and might also need to have `id` fields stripped off.

mwangggg:
> Looks good. Please double check if there are any other UI workflows that perform a JSON POST like this and might also need to have `id` fields stripped off.

This should be the only user upload where stripping the id applies.

PR #1241, opened 2024-04-16
fix(jmxauth): remove references to client-side localstorage JMX credential passthrough

### LINKED ISSUE #1239: [Bug] Remove "locally-stored credentials" from Security panel and Settings
### Current Behavior

The "locally storage credentials" note in the security panel and application settings should be removed. Cryostat 3.0 does not reimplement this bugged mechanism, which was also less secure since it relied on keeping applications' credentials in plaintext within the browser's localstorage. Instead, Cryostat 3.0 requires clients to register credentials into its encrypted database table, also using the more power matchExpression engine to determine which credentials can be applied to outgoing application connections. This server-side storage is already implemented on the web-client so all that must be done is to remove the other localstorage option.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

andrewazores:
# Welcome to Cryostat! 👋

Fixes #1239

## Description of the change:
The `X-JMX-Authentication` and `X-JMX-Authorization` headers were historically used by older Cryostat versions so that clients could supply these headers with API requests, and Cryostat would use them to set the JMX connection credentials when connection to target applications. Cryostat 3.0 does not implement this behaviour, so here references to it and the mechanism for selecting it are removed from the UI. It is no longer supported because it has been replaced by the server-side encrypted database/encrypted table for JMX credentials, which are more powerful due to the matchExpression mechanism, as well as more secure since they are encrypted at rest in the database rather than held in browser localStorage in plaintext.

## How to manually test:
2. *...*


src/app/SecurityPanel/Credentials/StoreCredentials.tsx
```diff
@@ -179,7 +178,7 @@ const tableColumns: TableColumn[] = [
 
 const tableTitle = 'Stored Credentials';
 
-export const StoreCredentials = () => {
+export const StoredCredentials = () => {
```
mwangggg, line 181: would it make more sense to rename this file to `StoredCredentials` as well?

mwangggg (approved):
lgtm

PR #1275, opened 2024-06-11
fix(matchexpr): tooltip hint checks for annotation existence before asserting value

andrewazores:
# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/cryostat3/issues/511

## Description of the change:
Adds a check that the annotation exists before reading it to compare its value.

## Motivation for the change:
In the old JavaScript-based expression evaluation, indexing the map with a nonexistent key would simply result in an `undefined` value which would compare as `false` against the test value. In CEL it is a runtime error to index with a nonexistent key, so there should be an existence check before the access and comparison to ensure that the result comes back as a successful evaluation with a `false` result, rather than as an error.

## How to manually test:
1. Check out cryostat3 with this PR and build
2. `./smoketest.bash -Ot`
3. Go to Automated Rules > Create, click a target in the miniature topology view
4. Click the match expression tooltip `?` icon. Copy the suggested expression and paste it into the match expression box.
5. If the suggested annotation exists only on some targets and not others, then the evaluation would fail prior to this PR.

### DIFF
--- a/src/app/Shared/Components/MatchExpression/MatchExpressionHint.tsx
+++ b/src/app/Shared/Components/MatchExpression/MatchExpressionHint.tsx
@@ -31,7 +31,9 @@ export const MatchExpressionHint: React.FC<MatchExpressionHintProps> = ({ target
     if (!target || !target.alias || !target.connectUrl) {
       body = 'true';
     } else {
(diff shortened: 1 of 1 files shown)


ebaron (approved):
Works for me

Issue #1272, opened 2024-06-06
[Bug] Target list fails to populate if WebSocket connection is not established


andrewazores:
### Current Behavior

See #1271

This is an old behaviour because of the web-client's history of originally using WebSockets exclusively, back when it was the "command channel", and the UI would send requests and receive responses entirely over WS. Since this has been phased out in favour of an HTTP REST-ish API, and WebSockets are only used to push notifications from the server, it no longer makes sense for the web-client to wait for the WebSocket connection to be established before querying for the target list or any other data from the server.

PR #1276, opened 2024-06-11
fix(Targets): Target list fails to populate if WebSocket connection is not established

aali309:
# Welcome to Cryostat! 👋

Fixes: #1272

## Description of the change:
see [https://github.com/cryostatio/cryostat-web/issues/1272](https://github.com/cryostatio/cryostat-web/issues/1272)

## Motivation for the change:
[https://github.com/cryostatio/cryostat-web/issues/1272#issue-2338382998](https://github.com/cryostatio/cryostat-web/issues/1272#issue-2338382998)

## How to manually test:
2. *...*


andrewazores:
I think the fix for this problem should include changing this behaviour:

https://github.com/cryostatio/cryostat-web/blob/1074c00536d4623085082681975acebb4874114b/src/app/Shared/Services/Targets.service.tsx#L38

ie there are places where `getSessionState()` is called to prevent things from updating until the websocket is connected. The line linked above prevents the target list from loading until there is a websocket connection. Instead of changing the session state calculation, which I think already accurately reflects the actual websocket connection state, I think it might be better to search for places that call `getSessionState()` and see if it's really necessary to call it and check for a session.

For example here as well:

https://github.com/cryostatio/cryostat-web/blob/1074c00536d4623085082681975acebb4874114b/src/app/Shared/Services/Api.service.tsx#L85

src/app/Shared/Services/Targets.service.tsx
```diff
@@ -33,12 +33,9 @@ export class TargetsService {
     login: LoginService,
     notificationChannel: NotificationChannel,
   ) {
-    login
-      .getSessionState()
-      .pipe(concatMap((sessionState) => (sessionState === SessionState.USER_SESSION ? this.queryForTargets() : EMPTY)))
```
andrewazores, line 38: Lots of unused imports now that can be removed.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -81,19 +81,15 @@ export class ApiService {
     private readonly login: LoginService,
   ) {
     // show recording archives when recordings available
```
andrewazores, line 83: This comment can probably be removed too. It used to be, long ago, that Cryostat could be deployed/run in a configuration without any archived recording storage. Nowadays we expect that there is an S3 storage provider configured, and if there is not that is just an invalid deployment configuration and not something the UI needs to be able to handle.

`this.archiveEnabled` can probably just be removed since it should always simply emit `true` now. Maybe for now to keep the patch small, the Subject can be replaced with a `BehaviorSubject` that starts with the `true` value and the call to `.next(true)` can be removed. Later we can clean up to remove it fully.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -66,11 +66,10 @@ import {
 import { isHttpError, includesTarget, isHttpOk, isXMLHttpError } from './api.utils';
 import { LoginService } from './Login.service';
 import { NotificationService } from './Notifications.service';
-import { SessionState } from './service.types';
 import { TargetService } from './Target.service';
 
 export class ApiService {
-  private readonly archiveEnabled = new ReplaySubject<boolean>(1);
+  private readonly archiveEnabled = new ReplaySubject<boolean>(true);
```
andrewazores, line 72: I think this should be `BehaviorSubject`.

Screenshot_2024-06-11_14-49-10

andrewazores (approved):

PR #1295, opened 2024-07-09
feat(graphql): improve error handling

Josh-Matsuoka:
Fixes https://github.com/cryostatio/cryostat/issues/394

Currently when graphQL queries fail due to missing SSL configuration or authentication failure the Web UI can't handle them correctly because they aren't throw in the traditional way. Rather than generating a 427 or 502 and accompanying HttpException they're instead masked by GraphQL and return a 200 but with a special response.

This response takes the form of an error object with various fields determined by the error handling configuration on the server. The accompanying cryostat3 PR changes this configuration to give us a form we can work with to determine the cause of the error and deal with it accordingly. With the server correctly configured this gives us the underlying exception and message, and the graphQL error classification (i.e. DataFetchingError).

This PR updates the graphQL queries and surrounding error handling to account for errors correctly and set the auth/ssl failure flags on the target accordingly.

### DIFF
--- a/src/app/Shared/Services/Api.service.tsx
+++ b/src/app/Shared/Services/Api.service.tsx
@@ -18,7 +18,18 @@ import { LayoutTemplate, SerialLayoutTemplate } from '@app/Dashboard/types';
 import { authFailMessage, missingSSLMessage } from '@app/ErrorView/types';
 import { createBlobURL } from '@app/utils/utils';
 import { ValidatedOptions } from '@patternfly/react-core';
-import { EMPTY, forkJoin, from, Observable, ObservableInput, of, ReplaySubject, shareReplay, throwError } from 'rxjs';
+import {
+  BehaviorSubject,
+  EMPTY,
+  forkJoin,
+  from,
+  Observable,
+  ObservableInput,
+  of,
+  ReplaySubject,
+  shareReplay,
+  throwError,
+} from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
 import { catchError, concatMap, filter, first, map, mergeMap, tap } from 'rxjs/operators';
 import {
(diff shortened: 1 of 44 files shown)


src/app/Shared/Services/Api.service.tsx
```diff
@@ -1278,7 +1279,7 @@ export class ApiService {
   }
 
   getTargetMBeanMetrics(target: TargetStub, queries: string[]): Observable<MBeanMetrics> {
-    return this.graphql<MBeanMetricsResponse>(
+    return this.graphql<any>(
```
andrewazores, line 1282: Would `MBeanMetricsResponse | Error` work as a type declaration here?

andrewazores (commented):
Quite a few eslint warnings emitted:

https://github.com/cryostatio/cryostat-web/pull/1295/files

Josh-Matsuoka:
There are a few test failures I'm working on resolving and I need to fix the snapshot tests as well. eslint and formatting issues are fixed

src/app/Shared/Services/Api.service.tsx
```diff
@@ -1295,6 +1302,15 @@ export class ApiService {
         }`,
       { id: target.id! },
     ).pipe(
+      tap((resp) => {
+        if (resp.data == undefined) {
+          if (isGraphQLAuthError(resp)) {
```
andrewazores, line 1307: Would it make sense to extract this behaviour out from `getTargetMBeanMetrics` into the more general `this.graphql` function? This way the error handling applies to all GraphQL responses, not only MBean metrics responses.

Also, there is a `handleError` function in this class which would be nice to reuse, or to modify it to use these `isGraphQL(Auth/SSL)Error` functions if possible.

src/app/Shared/Services/Api.service.tsx
```diff
@@ -1295,6 +1302,15 @@ export class ApiService {
         }`,
       { id: target.id! },
     ).pipe(
+      tap((resp) => {
+        if (resp.data == undefined) {
```
andrewazores, line 1306: Current example of a failing MBean metrics response:

```json
{
  "errors": [
    {
      "message": "Client Error (427)",
      "locations": [
        {
          "line": 1,
          "column": 105
        }
      ],
      "path": [
        "targetNodes",
        0,
        "target",
        "mbeanMetrics"
      ],
      "extensions": {
        "classification": "DataFetchingException",
        "exception": "io.vertx.ext.web.handler.HttpException"
      }
    }
  ],
  "data": {
    "targetNodes": [
      {
        "target": {
          "mbeanMetrics": null
        }
      }
    ]
  }
}
```

I think this should check for a condition like `resp.errors && Array.isArray(resp.errors) && resp.errors.length > 0`, rather than `resp.data == undefined`.

src/app/Shared/Services/Api.service.tsx
Josh-Matsuoka: Makes sense, I've reorganized it to throw an HttpError in this.graphql if a request comes back with errors and filter on the graphql exception in handleError.

andrewazores (approved):

PR #1334, opened 2024-08-23
chore(pf5): revamp general table styles and upgrade archive/events views to PF5

tthvo:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Related to #1303 

## Description of the change:

- Implemented ideas for improving overall UI related to Tables.
  - Views affected are `Rules`, `Recordings`, `Events`, `Archives`.
  - Security view is also affected but will be handled in #1333.
- Upgrades to PF5 are included for the following view along the way:
  - `Archive` view to Patternfly 5 (Dark theme should also be compatible here).
  - `Event` view to Patternfly 5 (Dark theme should also compatible here).

**Notes**: Security view can use some of these changes here. Maybe those can be applied in #1333 

## Minor changes

| Changes | Description | Screenshot |
|--------------|-------------------|------------------|
| Update placeholder texts to reflect "finding" functionality & Use `<SearchInput />` instead of `<TextInput />` component | Finding is not the same as filtering (i.e. persistent) |  image |
| Extend Recording table card to fill all available spaces & Toolbar is not sticky anymore | This is done to support scrolling (i.e. needed a concrete height) & The table is already sticky so toolbar does not need to. |  image |
| Tables in Archive view are now extended to fill and scrollable | Consistent with Recording view | image |
| Archive counts are presented with clickable file icons. If clicked, expanded view is toggle | It gives a feeling of  "open"/"close" to see/unsee files |  image |
| A minimal/clean & copy-able code block look for Match Expression | Nicer look for "code-like" data | image |

## Major changes

| Changes | Description | Screenshot |
|--------------|-------------------|------------------|
[...]

PR #1303, opened 2024-07-15
build(deps): upgrade to Patternfly 5

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Fixes: #1110

~~**TODO** update from PF 5.1.0 to 5.3.0~~

Notably broken things:
- dashboard card creation/catalog cannot progress: 
image
- menu item for Quick Starts must be clicked on the actual text - clicking the rest of the menu item does nothing: 
image
- about modal has strange image rendering: 
image
- topology view has weird black/white items: 
image 
image
- various dropdown menus are missing drop shadows and/or include odd borders 
image
- the target context selector dropdown has a very limited height. We are using PF 5.1.0 here, but it looks like at least 5.3.0 (re)introduced a property to control this `maxMenuHeight` - https://www.patternfly.org/components/menus/dropdown#dropdown : 
image
- various features in the Settings menu do not work: 1. Advanced > Feature Levels > selecting a different level causes the panel to go blank. 2. Notifications & Messages toggles do not work and emit a console warning about cyclic value. 3. Connectivity > Auto Refresh checkbox can be checked but not unchecked, but after a page refresh it goes back to unchecked. 4. General > Theme does not work.

PR #1335, opened 2024-08-24
feat(recordings): UI improvements for recording filters

tthvo:
# Welcome to Cryostat! 👋

Related to #1303 
Depends on #1334 

## Description of the change:

- Implemented some ideas to improve filter toolbar in Recording table.
- Fixed the typeahead toggle's focus. In PF5, first list item is focused by default, which causes the text input to lose focus, thus typing does not register any characters.

## Minor changes

| Changes | Description | Screenshot |
|--------------|-------------------|------------------|
| Use "read" style for count badge (i.e. grey background) | Following [PF5 examples](https://www.patternfly.org/components/menus/select#checkbox-select). Mostly, to avoid potential misunderstanding for an unread action |image |

## Major changes

| Changes | Description | Screenshot |
|--------------|-------------------|------------------|
| Duration Filter should allow range selection and have a search button for confirmation. "Enter" key is still supported | Range selection allows more flexible filter options. Also, it was previously hard to figure out the "Enter" key is for confirmation. The continuous checkout is moved down to make space & it seems similar to that in the Recording Create Form | image |
| Combined view for date and time picker. 24-hr mode switch is now replaced with a toggle group (to represent modes, for example, YAML/JSON view). Some border is fixed to match PF styles | A single view reduces the steps required to select a datetime (i.e. avoid switching tabs back and forth) | image |


src/app/Recordings/Filters/DurationFilter.tsx
```diff
@@ -38,37 +43,50 @@ export const DurationFilter: React.FC<DurationFilterProps> = ({
...
-          isChecked={isContinuous}
-          onChange={handleContinuousCheckBoxChange}
-        />
+    <Flex spaceItems={{ default: 'spaceItemsSm' }}>
+      <Flex direction={{ default: 'column' }}>
+        <FlexItem>
+          <TextInput
+            type="number"
+            value={duration}
+            id="duration-input"
```
andrewazores, line 64: This seems like a good time to enhance this duration picker input. Having to match an exact duration is pretty limiting, especially since the UI only allows for granularity of seconds. "Duration greater than" and "Duration less than" with a number spinner and unit dropdown would be nice, so the user can quickly do something like "filter for any recording at least 5 minutes long"

tthvo: Oh nice, great idea! How about this design below? The `Duration` filter is displayed as a range selection. We can do the same for `Datetime` filter (i.e. `StartedBeforeDate` and `StartedAfterDate`).

image

The latest commit here should have this implemention.

andrewazores: Looks pretty good to me. I think adding an "unlimited" or "infinite" option to either end of the range would be good, so that it's quick to express the same "start time before X" as "start time from 'infinite past' to X", without forcing the user to pick the Unix Epoch manually as their infinite past or something like that. Maybe that just exists as a checkbox? Not sure. Or maybe the existing "start time before/after" filters should just stay as they are for that purpose?

tthvo: I guess we can follow the same pattern here with the Duration by considering "unset"/"undefined" input as `Infinity`. Something like below could mean `From 2024-09-05T14:50:23.940Z onwards`?

image

tthvo: This latest commit should have the datetime range selection now for `StarTime` field of a recording :D

image

src/app/Recordings/Filters/DatetimeFilter.tsx
```diff
@@ -26,150 +27,239 @@ import {
...
+
+const _emptyDatetimeInput: _DatetimeState = {
   text: '',
   date: undefined,
   validation: ValidatedOptions.default,
 };
 
+export interface DateTimeRange {
+  from?: string; // ISO format
+  to?: string; // ISO format
```
andrewazores, line 55: This typecheck comment kind of scares me :-)

It looks like it would be worth defining a type like `DateTime`, which just looks roughly like:

```
class DateTime {
  private constructor(private readonly str: string) { }

  public toString(): string {
    return this.str;
  }

  public static of(str: string): DateTime {
    if (!validate(str)) {
      throw new Error();
    }
    return new DateTime(str);
  }
}
```

Basically, just a format-validated wrapper around a `string`, so that we can use the compiler for some build-time type safety to ensure runtime safety (assuming the `validate()` function is correct).

andrewazores: Or, maybe there is an existing type like `Date` that could be suitable here instead, if there is some function that can take a `Date` and print it in ISO-8601 `string` format?

andrewazores:
Awesome work on the filter redesign here, these look and feel great to use!

One behaviour I think might be worth changing:

image

image

To me, it "feels" like a Continuous recording should be included in a "duration longer than X" filter, because Continuous ~= "Infinity".

src/app/Recordings/Filters/DatetimeFilter.tsx
tthvo: > Basically, just a format-validated wrapper around a string, so that we can use the compiler for some build-time type safety to ensure runtime safety (assuming the validate() function is correct). This way, if we get some bad input, we have checked it early and failed at a well-known point where we have a nice validation function to tell us exactly what went wrong, rather than potentially failing somewhere deep in the call stack with strange error messages or just really unexpected behaviours.

Oh right, I think we could use `Date` type here. Though, within the Redux store, the filters will likely need to be stored as ISO strings as we persist the store data into Local Storage.

andrewazores: That's fine, in that case the Redux<->LocalStorage interface is where the (de)serialization and ISO-8601 handling is performed, so it can be at that boundary where the Date<->string conversion is done.

andrewazores:
> I can get the result I want here by putting in the "duration longer than X" filter plus ticking the Continuous checkbox, but somehow this doesn't feel quite intuitive to me.

Maybe hitting Enter/clicking Search with only a duration lower bound should auto-tick the Continuous checkbox?

tthvo:
> Maybe hitting Enter/clicking Search with only a duration lower bound should auto-tick the Continuous checkbox?

I think this idea make sense. Just that we have a filter value of "Continuous", which that checkbox handles. If this above idea is implemented, when uncheck the box will now also remove the filters without upper bound? Maybe that would not be okay?

src/app/Recordings/Filters/DatetimeFilter.tsx
tthvo: Sounds good! Fixed now. I added a `sanitize` func to convert ISO string to date when Redux store is loaded.

tthvo:
> > Maybe hitting Enter/clicking Search with only a duration lower bound should auto-tick the Continuous checkbox?
> 
> I think this idea make sense. Just that we have a filter value of "Continuous", which that checkbox handles. If this above idea is implemented, when uncheck the box will now also remove the filters without upper bound? Maybe that would not be okay?

The latest commit should have this behaviour. The checkbox is stucked when there is a filter without an upper bound...

andrewazores:
I think something about the LocalStorage/Redux changes may have gone wrong. This happens after clearing site data or using a private browsing window (Firefox).

Screenshot_2024-09-06_13-07-34

tthvo:
> I think something about the LocalStorage/Redux changes may have gone wrong. This happens after clearing site data or using a private browsing window (Firefox).

Oh hmmm, I don't see that on my side....Is there anything on the console that might give some insights?

tthvo:
> > > Maybe hitting Enter/clicking Search with only a duration lower bound should auto-tick the Continuous checkbox?
> > 
> > 
> > I think this idea make sense. Just that we have a filter value of "Continuous", which that checkbox handles. If this above idea is implemented, when uncheck the box will now also remove the filters without upper bound? Maybe that would not be okay?
> 
> The latest commit should have this behaviour. The checkbox is stucked when there is a filter without an upper bound...

For this, the filtering behaviour is as intended. But maybe leave out the visual part where the Continuous checkbox is auto-tick?

andrewazores:
Whoops, my bad. Ignore me. I forgot to update dependencies again after switching back to testing this branch! It's all good once I make sure I'm actually building with pf5 haha.

andrewazores:
>  >    The latest commit should have this behaviour. The checkbox is stucked when there is a filter without an upper bound...

> For this, the filtering behaviour is as intended. But maybe leave out the visual part where the Continuous checkbox is auto-tick?

Hmm. Yes, it's working the way my intuition likes it now, but it does feel like a bug that the box can't be unticked. Maybe disabling it and adding a hover tooltip to explain... ?

tthvo:
> Hmm. Yes, it's working the way my intuition likes it now, but it does feel like a bug that the box can't be unticked. Maybe disabling it and adding a hover tooltip to explain... ?

Actually, that sounds good! How about now?

locales/en/public.json
```diff
@@ -313,15 +313,29 @@
...
     },
-    "SELECTED_DATETIME": "Selected Date and Time",
-    "SELECTED_TIMEZONE": "Selected Timezone"
+    "DATETIME": "Date and Time"
+  },
+  "DurationFilter": {
+    "ARIA_LABELS": {
+      "FROM_DURATION": "cryostat_tmp",
+      "SEARCH_BUTTON": "Search for duration",
+      "TO_DURATION": "cryostat_tmp"
```
andrewazores, line 331: tmp strings here?

src/app/Recordings/Filters/DatetimeFilter.tsx
```diff
@@ -26,153 +27,239 @@ import {
...
+
+const _emptyDatetimeInput: _DatetimeState = {
   text: '',
   date: undefined,
   validation: ValidatedOptions.default,
 };
 
+export interface DateTimeRange {
+  from?: Date | string;
+  to?: Date | string;
```
andrewazores, line 55: Can these be always `Date`, and wherever they are currently assigned `string`s be forced to do the conversion at that point?

src/app/Recordings/Filters/DurationFilter.tsx
```diff
@@ -14,61 +14,190 @@
...
+} from '@patternfly/react-core';
+import { SearchIcon } from '@patternfly/react-icons';
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+import { DURATION_INPUT_MAXLENGTH } from './const';
+
+export const CONTINUOUS_INDICATOR = 'continuous';
+
+export const filterRecordingByDuration = (recordings: ActiveRecording[], filters?: DurationRange[]) => {
+  if (!recordings || !recordings.length || !filters || !filters.length) {
```
andrewazores, line 42: I think `!x || !x.length` can be collapsed down to `!x?.length`. It's a little more concise, but maybe not so readable... feel free to take or leave this style

src/app/Shared/Components/DurationUnitSelect.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export enum DurationUnit {
+  HOUR,
+  MINUTE,
+  SECOND,
+}
+
+const SupportedDurationUnits = Object.values(DurationUnit).filter((entry) => typeof entry === 'string');
+
+export const convertDurationToSeconds = (duration: number, unit: DurationUnit): number => {
```
andrewazores, line 29: This looks like it would be a good candidate to make into a method of the Duration type, or else to at least accept `Duration` as a parameter.

```typescript
export class Duration {
  constructor(readonly value: number, readonly unit: DurationUnit) { }

  convertTo(unit: DurationUnit): Duration {
    // convert between this.unit and unit parameter, return this.duration * scalar factor
  }
}

// elsewhere...

const dur = new Duration(10, DurationUnit.MINUTE);
const seconds = dur.convertTo(DurationUnit.SECOND).duration;
```

locales/en/public.json
tthvo: Opps, forgot to do that! Thanks!

src/app/Shared/Components/DurationUnitSelect.tsx
tthvo: > This looks like it would be a good candidate to make into a method of the Duration type, or else to at least accept Duration as a parameter.

Oh okayy right, I was writing with Golang styles of pointer receiver haha. Updated now. I still want to keep the interface type though (just for easier serialization of Redux state, i.e. avoid additional sanitizing).

andrewazores (approved):

PR #1426, opened 2024-09-24
feat(diagnostics): Add Dashboard card for invoking diagnostic operations

Josh-Matsuoka:
Hi,

Related to https://github.com/cryostatio/cryostat/issues/133

See https://github.com/cryostatio/cryostat-core/pull/459 and https://github.com/cryostatio/cryostat/pull/644/ for core and base framework

Adds a dashboard card for performing diagnostic commands on a target. Currently only adds support for running GC, it's extensible in future to add thread and heap dumps as well.

Opening as draft while I adjust it and the agent side to the endpoints created by the above PRs


src/app/Dashboard/Charts/diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,186 @@
...
+  return DiagnosticsCardKind[kind];
+}
+
+export const DiagnosticsCard: DashboardCardFC<DiagnosticsCardProps> = (props) => {
+  const { t } = useTranslation();
+  const serviceContext = React.useContext(ServiceContext);
+  const addSubscription = useSubscriptions();
+
+  const handleError = React.useCallback((error) => {
+    return <ErrorView title={'Error executing diagnostic command'} message={`${error.message}`} />;
```
andrewazores, line 65: This looks kind of odd. Should this fire a notification instead, or something? This callback returns a renderable component, but I don't see that this component is actually going to get mounted to the DOM and rendered when this callback is called.

src/app/Dashboard/Charts/diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,186 @@
...
+  }, []);
+
+  const handleGC = React.useCallback(() => {
+    addSubscription(
+      serviceContext.api.runGC().subscribe({
+        error: (err) => handleError(err),
```
andrewazores, line 71: It might be nice to put a progress spinner somewhere on the UI. Display it when the request is fired, and hide it when the response is received.

src/app/Dashboard/Charts/diagnostics/DiagnosticsCard.tsx
andrewazores, line 1: I think something like `src/app/Dashboard/Diagnostics` would be a better home for this file, since this isn't really a "chart card". It's more akin to the `JvmDetails` or `AutomatedAnalysis` cards, which have their own Dashboard subdirectories.

locales/en/public.json
```diff
@@ -157,6 +157,10 @@
     "MBEAN_METRICS_CARD_DESCRIPTION": "Display common performance metrics from current MBean data.",
     "MBEAN_METRICS_CARD_DESCRIPTION_FULL": "Display a single performance metric from a list of supported MBeans.",
     "MBEAN_METRICS_CARD_TITLE": "MBean Metrics Chart",
+    "DIAGNOSTICS_CARD_DESCRIPTION": "Perform diagnostic operations on the target.",
+    "DIAGNOSTICS_CARD_DESCRIPTION_FULL": "Perform diagonstic operations from a list of supported operations on the target.",
+    "DIAGNOSTICS_CARD_TITLE": "Diagnostics",
+    "DIAGNOSTICS_GC_BUTTON": "Start Garbage Collection",
```
andrewazores, line 163: These localization strings should also be separated out rather than nested under `CHART_CARD`.

src/app/Dashboard/Charts/diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,200 @@
...
+    return (
+      <CardHeader
+        actions={{
+          actions: <>{actions}</>,
+          hasNoOffset: false,
+          className: undefined,
+        }}
+      >
+        <CardTitle>
+          {t('CHART_CARD.TITLE', { chartKind: props.chartKind, duration: props.duration, period: props.period })}
```
andrewazores, line 120: Reusing this string is why the card has an odd title like this:

image

so after splitting this card out into its own subdirectory and its own localization "block" (not sure of the proper terminology there), it should also have its own independent set of localization strings for all these card titles and contents.

src/app/Dashboard/Diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+import {
+  DashboardCardTypeProps,
+  DashboardCardFC,
+  DashboardCardSizes,
+  DashboardCardDescriptor,
+} from '@app/Dashboard/types';
+import { ErrorView } from '@app/ErrorView/ErrorView';
+import { FeatureLevel } from '@app/Shared/Services/service.types';
+import { LoadingProps } from '@app/Shared/Components/types';
```
andrewazores, line 25: One small eslint suggestion here to fix up

src/app/Dashboard/Diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,204 @@
...
+export function kindToId(kind: string): number {
+  return DiagnosticsCardKind[kind];
+}
+
+export const DiagnosticsCard: DashboardCardFC<DiagnosticsCardProps> = (props) => {
+  const { t } = useTranslation();
+  const serviceContext = React.useContext(ServiceContext);
+  const addSubscription = useSubscriptions();
+  const [errorMessage, setErrorMessage] = React.useState('');
+  const [running, setRunning] = React.useState(false);
```
andrewazores, line 64: eslint reveals `setRunning` is never used - I think it should be called `setRunning(true)` immediately at the top of `handleGC`, and `setRunning(false)` within the service subscription there (and within the error handler).

src/app/Dashboard/Diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,204 @@
...
+        error: (err) => handleError(err),
+      }),
+    );
+  }, [addSubscription, serviceContext.api, handleError]);
+
+  const GCButton = React.useMemo(() => {
+    return (
+      <Button
+        key={0}
+        aria-label={t('DaignosticsCard.DIAGNOSTICS_GC_BUTTON', { chartKind: props.chartKind })}
```
andrewazores, line 86: Typo here: `Daignostics` -> `Diagnostics`

src/app/Dashboard/Diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,204 @@
...
+        onClick={handleGC}
```
andrewazores, line 87: If I'm following correctly,

image

This `<Button>` with the `<SyncAltIcon />` is the "refresh" button appearing in the card header, and when clicked it performs the same action as the big clue "Start Garbage Collection" button?

I think for now we should just pick one button or the other, rather than both. I don't know for sure what's best to go with since it probably depends on what other diagnostic actions or data retrieval we might implement in the future, and whether or not those additions in the future are part of this same card or are their own separate cards. Anyway, we're flagging this as a Beta component, so I don't mind if the UI is not totally settled and ends up changing significantly in the future, so let's just pick one button or the other and stick with it.

If we stick with this "refresh" header button then I think it will at least need a hover tooltip or something to inform the user about what the button does.

src/app/Dashboard/Diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,204 @@
...
+  const actions = React.useMemo(() => {
+    const a = props.actions || [];
+    return [GCButton, ...a];
+  }, [props.actions, GCButton]);
+
+  const gcButtonLoadingProps = React.useMemo(
+    () =>
+      ({
+        spinnerAriaValueText: 'Invoke GC',
+        spinnerAriaLabel: 'saving-credentials',
```
andrewazores, line 104: Seems like the wrong text for this label?

src/app/Dashboard/Diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+export const DiagnosticsCardDescriptor: DashboardCardDescriptor = {
+  featureLevel: FeatureLevel.BETA,
+  title: 'DiagnosticsCard.DIAGNOSTICS_CARD_TITLE',
+  cardSizes: DiagnosticsCardSizes,
+  description: 'DiagnosticsCard.DIAGNOSTICS_CARD_DESCRIPTION',
+  descriptionFull: 'DiagnosticsCard.DIAGNOSTICS_CARD_DESCRIPTION_FULL',
+  component: DiagnosticsCard,
+  propControls: [],
+  icon: <TachometerAltIcon />,
```
andrewazores, line 193: I think we already use this Tachometer icon for the JFR metrics cards:

image

it would probably be better to choose a unique icon.

Maybe the Wrench icon is a good visual for diagnostic actions?

image

https://www.patternfly.org/design-foundations/icons

src/app/Dashboard/Diagnostics/DiagnosticsCard.tsx
```diff
@@ -0,0 +1,204 @@
...
+      isDraggable={props.isDraggable}
+      isResizable={props.isResizable}
+      isFullHeight={props.isFullHeight}
+    >
+      <CardBody>
+        <Bullseye>
+          <EmptyState variant={EmptyStateVariant.lg}>
+            <EmptyStateHeader
+              titleText={<>{t('DiagnosticsCard.DIAGNOSTICS_CARD_TITLE')}</>}
+              icon={<EmptyStateIcon icon={DataSourceIcon} />}
```
andrewazores, line 149: Same here about the DataSourceIcon - I think this should match the icon that represents this card in the card catalog, ex. the WrenchIcon perhaps?

src/app/Dashboard/Diagnostics/DiagnosticsCard.tsx
Josh-Matsuoka: I think sticking with the GC button makes more sense, I'll remove the refresh button

andrewazores (approved):

PR #1491, opened 2024-11-21
feat(async): long-running async job notification handling

Josh-Matsuoka:
# Welcome to Cryostat! 👋

Related to: https://github.com/cryostatio/cryostat/issues/286

Companion to https://github.com/cryostatio/cryostat/pull/698

Adds notification handling for the new notifications emitted for long running API operations, refreshing the recordings table and automated analysis when the client received the relevant notifcations

### DIFF
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -930,7 +930,8 @@ export const ActiveRecordingRow: React.FC<ActiveRecordingRowProps> = ({
 
   React.useEffect(() => {
     addSubscription(
-      context.notificationChannel.messages(NotificationCategory.ReportSuccess).subscribe(() => {
+      context.notificationChannel.messages(NotificationCategory.ReportSuccess).subscribe((notification) => {
+        if (context.reports.jobIds.includes(notification.message.jobId))
         handleLoadAnalysis();
       }),
     );
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -874,6 +874,8 @@ export const ActiveRecordingRow: React.FC<ActiveRecordingRowProps> = ({
   handleRowCheck,
   updateFilters,
 }) => {
+
+  const addSubscription = useSubscriptions();
   const { t } = useTranslation();
   const [dayjs, datetimeContext] = useDayjs();
   const context = React.useContext(ServiceContext);
@@ -926,6 +928,15 @@ export const ActiveRecordingRow: React.FC<ActiveRecordingRowProps> = ({
     }
     handleLoadAnalysis();
   }, [isExpanded, handleLoadAnalysis]);
+
+  React.useEffect(() => {
+    addSubscription(
+      context.notificationChannel.messages(NotificationCategory.ReportSuccess)
+      .subscribe((event) => {
+        handleLoadAnalysis();
+      })
+    )
+  }, [addSubscription, context.notificationChannel, handleLoadAnalysis])
 
   const recordingOptions = (recording: ActiveRecording): KeyValue[] => {
     const options: KeyValue[] = [];
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -931,10 +931,10 @@ export const ActiveRecordingRow: React.FC<ActiveRecordingRowProps> = ({
   React.useEffect(() => {
     addSubscription(
       context.notificationChannel.messages(NotificationCategory.ReportSuccess).subscribe((notification) => {
-        if (context.reports.jobIds.includes(notification.message.jobId)) handleLoadAnalysis();
+        if (context.reports.getJobIds().includes(notification.message.jobId)) handleLoadAnalysis();
       }),
     );
-  }, [addSubscription, context.notificationChannel, context.reports.jobIds, handleLoadAnalysis]);
+  }, [addSubscription, context.notificationChannel, context.reports, handleLoadAnalysis]);
 
   const recordingOptions = (recording: ActiveRecording): KeyValue[] => {
     const options: KeyValue[] = [];
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -931,11 +931,10 @@ export const ActiveRecordingRow: React.FC<ActiveRecordingRowProps> = ({
   React.useEffect(() => {
     addSubscription(
       context.notificationChannel.messages(NotificationCategory.ReportSuccess).subscribe((notification) => {
-        if (context.reports.jobIds.includes(notification.message.jobId))
-        handleLoadAnalysis();
+        if (context.reports.jobIds.includes(notification.message.jobId)) handleLoadAnalysis();
       }),
     );
-  }, [addSubscription, context.notificationChannel, handleLoadAnalysis]);
+  }, [addSubscription, context.notificationChannel, context.reports.jobIds, handleLoadAnalysis]);
 
   const recordingOptions = (recording: ActiveRecording): KeyValue[] => {
     const options: KeyValue[] = [];
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -930,11 +930,11 @@ export const ActiveRecordingRow: React.FC<ActiveRecordingRowProps> = ({
 
   React.useEffect(() => {
     addSubscription(
-      context.reports.getJobIds().subscribe(() => {
+      context.reports.onJobCompletion().subscribe(() => {
         handleLoadAnalysis();
       }),
     );
-  }, [addSubscription, context.reports, context.reports.getJobIds, handleLoadAnalysis]);
+  }, [addSubscription, context.reports, context.reports.onJobCompletion, handleLoadAnalysis]);
 
   const recordingOptions = (recording: ActiveRecording): KeyValue[] => {
     const options: KeyValue[] = [];
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -385,7 +385,7 @@ export const ActiveRecordingsTable: React.FC<ActiveRecordingsTableProps> = (prop
 
   const handleArchiveRecordings = React.useCallback(() => {
     setActionLoadings((old) => ({ ...old, ARCHIVE: true }));
-    const tasks: Observable<boolean>[] = [];
+    const tasks: Observable<string>[] = [];
     filteredRecordings.forEach((r: ActiveRecording) => {
       if (checkedIndices.includes(r.id)) {
         handleRowCheck(false, r.id);
@@ -394,13 +394,24 @@ export const ActiveRecordingsTable: React.FC<ActiveRecordingsTableProps> = (prop
     });
     addSubscription(
       forkJoin(tasks).subscribe({
-        next: () => handlePostActions('ARCHIVE'),
+        next: (jobIds) => {
+          addSubscription(
+            context.notificationChannel
+              .messages(NotificationCategory.ArchiveRecordingSuccess)
+              .subscribe((notification) => {
+                if (jobIds.includes(notification.message.jobId)) {
+                  handlePostActions('ARCHIVE');
+                }
+              }),
+          );
+        },
         error: () => handlePostActions('ARCHIVE'),
       }),
     );
   }, [
     filteredRecordings,
     checkedIndices,
+    context.notificationChannel,
     handleRowCheck,
     context.api,
     addSubscription,
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -874,7 +874,6 @@ export const ActiveRecordingRow: React.FC<ActiveRecordingRowProps> = ({
   handleRowCheck,
   updateFilters,
 }) => {
-
   const addSubscription = useSubscriptions();
   const { t } = useTranslation();
   const [dayjs, datetimeContext] = useDayjs();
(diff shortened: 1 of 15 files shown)


src/app/Shared/Services/Report.service.tsx
```diff
@@ -43,7 +43,12 @@ export class ReportService {
       headers,
     }).pipe(
       concatMap((resp) => {
+        // 200 indicates that the backend has the report cached and it will return
+        // the json in the response.
         if (resp.ok) {
+          if (resp.status == 202) {
+            this.notifications.info('Report generation in progress', 'Report is being generated');
+          }
```
andrewazores, line 51: I think in the case that we get a 200 response, we want to check the job UUID sent back by the server and hang on to that somehow. Maybe something as simple as a list of "report job UUIDs" that we append to.

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -927,6 +928,14 @@ export const ActiveRecordingRow: React.FC<ActiveRecordingRowProps> = ({
     handleLoadAnalysis();
   }, [isExpanded, handleLoadAnalysis]);
 
+  React.useEffect(() => {
+    addSubscription(
+      context.notificationChannel.messages(NotificationCategory.ReportSuccess).subscribe(() => {
+        handleLoadAnalysis();
```
andrewazores, line 934: If we have that "report job UUIDs" list somewhere, like as an accessible member of the Report service, then we can filter these notifications to only ones that include an expected UUID and only react to those ones. That way we don't trigger the `handleLoadAnalysis()` function if we observe a `ReportSuccess` notification that has a wrong UUID, ie. which was emitted for some other reason (like another client requested a report on another recording).

src/app/Recordings/ActiveRecordingsTable.tsx
```diff
@@ -930,11 +930,11 @@ export const ActiveRecordingRow: React.FC<ActiveRecordingRowProps> = ({
 
   React.useEffect(() => {
     addSubscription(
-      context.notificationChannel.messages(NotificationCategory.ReportSuccess).subscribe(() => {
-        handleLoadAnalysis();
+      context.notificationChannel.messages(NotificationCategory.ReportSuccess).subscribe((notification) => {
+        if (context.reports.getJobIds().includes(notification.message.jobId)) handleLoadAnalysis();
```
andrewazores, line 934: It looks like there's a nice opportunity to do some refactoring here. If we give the Reports Service a handle on the Notification Channel also, then the Reports Service can expose an Observable that emits a value whenever an expected long-running Report generation response notification is observed. Then this table, and the automated analysis card, and whatever other UI components need to react can simply subscribe to that Observable instead (rather than each subscribing to the overall notification stream and then checking the Report Service's internal state on their own).

src/app/Shared/Services/Report.service.tsx
```diff
@@ -28,6 +28,8 @@ export class ReportService {
     private notifications: NotificationService,
   ) {}
 
+  jobIds = new Array<String>();
```
andrewazores, line 31: I think this should be `private readonly`

src/app/Recordings/ActiveRecordingsTable.tsx
Josh-Matsuoka: I've done some refactoring here bringing it in line with how Targets.service does it (grabbing the NotificationChannel in the constructor and subscribing there) but it's running into an issue at runtime now,

It looks like the notification channel it gets from Services is undefined, trying to access this and subscribe is throwing a TypeError. 

Is there some other initialization step I've missed or something I'm not understanding with Typescript? Why does TargetsService get a valid NotificationChannel but ReportsService doesn't?

Josh-Matsuoka: nevermind, found the isssue, FakeData needed to be updated as well.

andrewazores:
@Josh-Matsuoka take a look at the last commit that I just pushed and see if that makes sense to you. I tried to refactor it to clean up the notification handling so that the individual components just receive a notification if an expected report was generated. This still isn't perfect, because the user could have requested multiple report generations in rapid succession, and this implementation will result in the component receiving a notification and attempting to re-render for each one - even if the notification is about generation completing for a different request.

src/app/Shared/Services/Report.service.tsx
```diff
@@ -43,7 +52,13 @@ export class ReportService {
       headers,
     }).pipe(
       concatMap((resp) => {
+        // 200 indicates that the backend has the report cached and it will return
+        // the json in the response.
         if (resp.ok) {
+          if (resp.status == 202) {
+            this.notifications.info('Report generation in progress', 'Report is being generated');
+            resp.text().then((value) => this.jobIds.add(value));
+          }
```
andrewazores, line 61: In this case, we emit a notification to inform the user about the async generation, and we start tracking the job ID - but we still end up returning an `Observable<AnalysisResult[]>` to the calling component, which we try to resolve with a value immediately based on the current response status body.

I think what we need here is that the returned Observable has a resolved value immediately if the response from the Cryostat server was an immediate cached report. If the server responded with the status code indicating async generation, then the Observable that we return to the component should not resolve yet - it should resolve once the result is actually available. We'll know it's available when we get the notification from the server. Then I guess this service can re-fire a request to the server to retrieve the actual report response body, and then finally we can resolve the Observable that we returned to the component.

andrewazores:
Here's a patch revising that latest attempt:

```diff
diff --git a/src/app/Shared/Services/Report.service.tsx b/src/app/Shared/Services/Report.service.tsx
index 9e4fd995..4bf2149f 100644
--- a/src/app/Shared/Services/Report.service.tsx
+++ b/src/app/Shared/Services/Report.service.tsx
@@ -16,7 +16,7 @@
 import { Base64 } from 'js-base64';
 import { Observable, Subject, from, throwError } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
-import { concatMap, first, tap } from 'rxjs/operators';
+import { concatMap, filter, first, tap } from 'rxjs/operators';
 import { Recording, CachedReportValue, GenerationError, AnalysisResult, NotificationCategory } from './api.types';
 import { isActiveRecording, isQuotaExceededError, isGenerationError } from './api.utils';
 import { NotificationChannel } from './NotificationChannel.service';
@@ -28,14 +28,14 @@ export class ReportService {
     channel: NotificationChannel,
   ) {
     channel.messages(NotificationCategory.ReportSuccess).subscribe((v) => {
-      if (this.jobIds.delete(v.message.jobId)) {
-        this._jobCompletion.next();
+      if (this.jobIds.has(v.message.jobId)) {
+        this._jobCompletion.next(v.message.jobId);
       }
     });
   }
 
-  private readonly jobIds: Set<string> = new Set();
-  private readonly _jobCompletion: Subject<void> = new Subject();
+  private readonly jobIds: Map<string, string> = new Map();
+  private readonly _jobCompletion: Subject<string> = new Subject();
 
   reportJson(recording: Recording, connectUrl: string): Observable<AnalysisResult[]> {
     if (!recording.reportUrl) {
@@ -55,15 +55,45 @@ export class ReportService {
         // 200 indicates that the backend has the report cached and it will return
         // the json in the response.
         if (resp.ok) {
+          // 202 indicates that the backend does not have the report cached and will return an
+          // async job ID in the response immediately, then emit a notification with the same
+          // job ID later to inform us that the report is now ready and can be retrieved by
+          // sending a follow-up GET to the Location header value
           if (resp.status == 202) {
-            resp.text().then((value) => this.jobIds.add(value));
-            const ge: GenerationError = {
-              name: `Report Generation in progress`,
-              message: `Report is being generated`,
-              messageDetail: from(resp.text()),
-              status: resp.status,
-            };
-            throw ge;
+            let subj = new Subject<AnalysisResult[]>();
+            resp.text().then((jobId) => {
+              this.jobIds.set(jobId, resp.headers.get('Location') || recording.reportUrl);
+              this._jobCompletion
+                .asObservable()
+                .pipe(filter((id) => id === jobId))
+                .subscribe((id) => {
+                  const jobUrl = this.jobIds.get(id);
+                  if (!jobUrl) throw new Error(`Unknown job ID: ${id}`);
+                  this.jobIds.delete(id);
+                  fromFetch(jobUrl, {
+                    method: 'GET',
+                    mode: 'cors',
+                    credentials: 'include',
+                    headers,
+                  }).subscribe((resp) => {
+                    if (resp.ok && resp.status === 200) {
+                      resp
+                        .text()
+                        .then(JSON.parse)
+                        .then((obj) => subj.next(Object.values(obj) as AnalysisResult[]));
+                    } else {
+                      const ge: GenerationError = {
+                        name: `Report Failure (${recording.name})`,
+                        message: resp.statusText,
+                        messageDetail: from(resp.text()),
+                        status: resp.status,
+                      };
+                      subj.error(ge);
+                    }
+                  });
+                });
+            });
+            return subj.asObservable();
           }
           return from(
             resp
@@ -130,10 +160,6 @@ export class ReportService {
     };
   }
 
-  onJobCompletion(): Observable<void> {
-    return this._jobCompletion.asObservable();
-  }
-
   delete(recording: Recording): void {
     sessionStorage.removeItem(this.key(recording));
   }
```

I haven't actually tested this, and it will need some more adjustments to account for the removal of `onJobCompletion()`. The concept is simple: the existing `reportJson(): Observable<AnalysisResult[]>` function is already defined with an async interface, given that it returns an `Observable`. So whether the server's response comes "immediately" or with a significant delay doesn't really matter - either way, the UI component that will render this data has to handle it in an async fashion. This means that the Report service is also free to do things like handle the 200/202 response status codes and the long-running API flow internally, without the UI components having to do anything. If the server responds with a 200 then the Report service acts just like it always did before. If it responds with a 202, then the service can hang on to the job ID that was passed, wait for the corresponding completion notification, then re-fire a request to get the actual report data, and finally use that to resolve the UI component's Observable. This should also ensure that the UI component only receives an event (data) when the specific report it asked for is ready, not when just any report generation completes.

This way, the UI components (like the recordings tables, or the automated analysis card) should not actually need to be adjusted at all. They should just call `reportJson()` as before. Only "under the covers", in the service implementation code, does the long-running job API become apparent.

andrewazores:
Along with removing the analysis result from the websocket notification (https://github.com/cryostatio/cryostat/pull/698#discussion_r1872084035), I think we want to do some handling on the client side to suppress these notifications from actually being rendered graphically. The UI is already going to react by rendering the content (assuming the user still has it visible), so there shouldn't really be a need to display a graphical notification at the same time. These two spots should be useful for tackling client-side notification handling:

https://github.com/cryostatio/cryostat-web/blob/2e5202b4dbee4e216210eae8073ad6553ea4b2b8/src/app/Shared/Services/api.types.ts#L505

https://github.com/cryostatio/cryostat-web/blob/2e5202b4dbee4e216210eae8073ad6553ea4b2b8/src/app/Shared/Services/api.utils.ts#L201

As it is, the two PRs seem to be working together as expected, but the notifications that appear are quite noisy and unhelpful:

image

andrewazores:
https://github.com/cryostatio/cryostat-web/blob/2e5202b4dbee4e216210eae8073ad6553ea4b2b8/src/app/Recordings/RecordingActions.tsx#L62

This event handler will need to be updated - right now it opens the Grafana tab as soon as the upload function returns a response, so it should now instead listen for the long-running job notification before opening the tab.

src/app/Recordings/RecordingActions.tsx
```diff
@@ -36,7 +36,7 @@ export interface RecordingActionsProps {
   index: number;
   recording: Recording;
   sourceTarget?: Observable<Target>;
-  uploadFn: () => Observable<boolean>;
+  uploadFn: () => Observable<String>;
```
andrewazores, line 39: This should be `string`, not `String`:

https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

src/app/Recordings/RecordingActions.tsx
```diff
@@ -58,18 +62,26 @@ export const RecordingActions: React.FC<RecordingActionsProps> = ({ recording, u
...
-      context.notificationChannel.messages(NotificationCategory.GrafanaUploadSuccess).subscribe(() => {
-        notifications.success('Upload Success', `Recording "${recording.name}" uploaded`);
-        context.api
-          .grafanaDashboardUrl()
-          .pipe(first())
-          .subscribe((url) => window.open(url, '_blank'));
-      }),
-    );
-  }, [addSubscription, notifications, context.api, context.notificationChannel, recording, uploadFn]);
+    uploadFn().subscribe((jobId) => {
```
andrewazores, line 65: This `.subscribe()` creates a `Subscription`, which should be tracked with `addSubscription`.

Maybe this subscription can be mapped so that the later inner notification subscription and this one are combined into one? I think that might also remove the need to track the `jobIds` in a separate array data structure.

Josh-Matsuoka: > Maybe this subscription can be mapped so that the later inner notification subscription and this one are combined into one? I think that might also remove the need to track the jobIds in a separate array data structure.

What would this look like? AutomatedAnalysisCard and utils.tsx both seem to have nested subscriptions like this.

andrewazores: I'm imagining something like this:

```
const sub = uploadFn.pipe(
  tap(() => notifications.info('Upload Started', `Recording "${recording.name}" uploading...`)),
  concatMap(jobId=>
    context.notificationChannel.messages(NotificationCategory.GrafanaUploadSuccess.pipe(filter(n => n.message.jobId === jobId))
  ),
  first(),
  tap(() => notifications.success('Upload Success', `Recording ${recording.name} uploaded`)),
  concatMap(() => context.api.grafanaDashboardUrl().pipe(first()))
).subscribe((url) => window.open(url, '_blank'));
addSubscription(sub);
```

I might not have got all the `concatMaps` and things right, but something like that.

src/app/Shared/Services/api.utils.ts
```diff
@@ -424,4 +424,58 @@ export const messageKeys = new Map([
...
+      variant: AlertVariant.warning,
+      title: 'Grafana Upload Failed',
+      body: (evt) => `Grafana upload failed for job: ${evt.message.jobId}`,
+      hidden: false,
+    } as NotificationMessageMapper,
+  ],
+  [
+    NotificationCategory.ArchiveRecordingSuccess,
+    {
+      variant: AlertVariant.warning,
```
andrewazores, line 466: image

This seems like it should be a non-hidden success message, not a hidden warning

andrewazores:
https://github.com/cryostatio/cryostat-web/blob/2e5202b4dbee4e216210eae8073ad6553ea4b2b8/src/app/Recordings/ActiveRecordingsTable.tsx#L688

Some handling needs to be updated on the Archive button in or around this component.

With the old implementation, the Archive button becomes disabled and has a loading spinner from when the UI fires the request, and the button becomes re-enabled and the spinner removed when the client receives the server's response (when the archiving action has completed).

Now, the response being received does not actually indicate that the action has completed, only that the server has accepted the request. So the button becomes re-enabled and the spinner is removed prematurely. The UI should wait for the job notification with the correct ID to be received, and only then re-enable the button and remove the spinner.

andrewazores (approved):

Issue #1526, opened 2025-01-09
[Task] Allow setting i18n namespaces via an environment variable


aptmac:
At the moment, cryostat-web currently makes use of two i18n namespaces: `common` and `public`.

There may be instances where users want to configure their own namespace(s) to be used instead, e.g., `plugin__cryostat-plugin` or something similar in the case of a Console plugin.

PR #1527, opened 2025-01-09
feat(i18n): allow setting i18n namespaces via an environment variable

aptmac:
# Welcome to Cryostat! 👋

Fixes: [#1526 ](https://github.com/cryostatio/cryostat-web/issues/1526)

## Description of the change:
This change allows the i18n namespace(s) to be set as an environment variable.

To do so, I've added `useCryostatTranslation` to `@i18n/i18nextUtil` which is just `useTranslation` with the required namespaces being set.

I've also moved `i18nResources` and `i18nLanguages` from `@i18n/config` to `@i18n/i18nextUtil`, because I was encountering issues where the Settings page was breaking on the Console plugin side and using a different i18n object than what the Console was expecting.

I have a branch for the console plugin that includes some work to use these changes: https://github.com/aptmac/cryostat-openshift-console-plugin/commit/c8d2d547c34cc1ff1f3960bfd79148086594c3eb

At the moment that branch is out of sync with upstream, I'll be fixing that up soon. That commit includes the ability to combine the cryostat-web `common` and `public` with the console plugin side `common`, and renames it to `plugin__cryostat-plugin` so it can be picked up by the Console.

## Motivation for the change:
This change is helpful because it will allow the console plugin to pass it's i18n namespace (something like `plugin__cryostat-plugin`) and allow the cryostat-web pages to access the localized strings.

## Before
(nothing changes for cryostat-web, but here's the console plugin):

Screenshot from 2025-01-09 11-52-35

## After
Screenshot from 2025-01-09 11-52-45

### DIFF
--- a/LOCALIZATION.md
+++ b/LOCALIZATION.md
@@ -4,9 +4,9 @@ Cryostat-web uses [i18next](https://www.i18next.com/) as an internationalization
 
 ### Adding a new translation
 
-The current list of language locales supported in Cryostat can be found in `src/i18n/config.ts`. The translations themselves can be found in `locales/{LOCALE_CODE}` 
+The current list of language locales supported in Cryostat can be found in `src/i18n/i18nextUtil.ts`. The translations themselves can be found in `locales/{LOCALE_CODE}` 
 
-To add a new language, add a new entry to the `i18nResources` object in `src/i18n.ts`. The key should be the language locale, and the value should be the translation object containing the corresponding namespace json files in `locales`.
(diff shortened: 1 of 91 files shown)


LOCALIZATION.md
andrewazores, line 1: A few lines further down there is an example that still instructs the reader to use the `useTranslation` hook. I guess this should be updated to document this new `useCryostatTranslation` hook.

aptmac: Thanks for catching that, just updated it

andrewazores (approved):

Issue #1286, opened 2024-06-14
[Bug] Active Recordings table fails to update when restarting a recording with `replace: ALWAYS`


aali309:
### Current Behavior

![Screenshot from 2024-06-14 13-06-42](https://github.com/cryostatio/cryostat-web/assets/56743004/cf2bf1ea-2bf5-4d6b-b03c-85e25d4ff814)

### Expected Behavior

Active Recordings should update to reflect the create recording notification.

### Steps To Reproduce

1. checkout on cryostatio/cryostat main branch
2.  build and run smoketest
3. visit `localhost:8080` 
4. navigate to `Recordings`  table
5. run the below query on terminal and see the behavior on the web UI

``` 
`http --follow -v --auth='user:pass' POST :8080/api/v3/graphql \
      Content-Type:application/json \
      query='
  query {
    targetNodes(filter: { annotations: ["PORT = 9091"] }) {
      name
      target {
        doStartRecording(recording: {
          name: "test",
          template: "Profiling",
          templateType: "TARGET",
          replace: "ALWAYS"
        }) {
          name
          state
        }
      }
    }
  }'

```

### Environment

```Markdown
- OS: 
- Environment: 
- Version:
```

andrewazores:
This is just a notification handling issue, right? If I understand correctly, the notification appears but the table does not update. If you navigate away and back, or refresh, the table contains the correct new content?

andrewazores:
Please also check for any other similar cases with this view/table not updating. It sounds like the particular query you used to perform the update would not be the only way to trigger this behaviour. Like, is the `replace: ALWAYS` the only time it breaks, or do `STOPPED` and `NEVER` do it too?

tthvo:
I did a quick test locally and the behaviour is correct with the notification updating the table. Might just be another issue with flanky notification handling.

![image](https://github.com/cryostatio/cryostat-web/assets/68053619/ae06a3e6-2c54-477b-bc82-59b5a7daec74)

aali309:
> This is just a notification handling issue, right? If I understand correctly, the notification appears but the table does not update. If you navigate away and back, or refresh, the table contains the correct new content?

Yes, just like what was happening with the Archived tables before and labels.

aali309:
[...]

PR #1535, opened 2025-01-17
fix(recordings): Fix active recordings table refresh behavior for replace:"ALWAYS"

Josh-Matsuoka:
Fixes: [#1286](https://github.com/cryostatio/cryostat-web/issues/1286)

This changes the activeRecordingsTable notification handling logic to call refreshRecordingsList when receiving creation/deletion notifications rather than updating with only the fields in the notification. This ensures that the state of the table remains accurate.

When the query in the linked issue is run a second time it results in the notifications being fired in the following order:

Recording Created
Recording Deleted

Pre-patch the table would see the recording deleted notification last and remove it from the table despite the recording still being present in the target. Calling refreshRecordingsList and getting the state of the target's recordings from the API shows the same.

## How to manually test:
1. Run Cryostat via smoketest or ./mvnw quarkus:dev
2. Run the following query twice: (Replace 8080 with 8181 if in quarkus:dev mode)

http --follow -v --auth='user:pass' POST :8080/api/v4/graphql \
                                                                  Content-Type:application/json \
                                                                  query='
                                                          query {
                                                            targetNodes(filter: { annotations: ["PORT = 9091"] }) {
                                                              name
                                                              target {
                                                                doStartRecording(recording: {
                                                                  name: "test9", 
                                                                  template: "Profiling",
                                                                  templateType: "TARGET",
                                                                  replace: "ALWAYS" 
                                                                }) {
                                                                  name
                                                                  state
                                                                }
                                                              }
                                                            }
                                                          }'

### DIFF
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -260,10 +260,10 @@ export const ActiveRecordingsTable: React.FC<ActiveRecordingsTableProps> = (prop
         if (currentTarget?.jvmId != event.message.jvmId) {
           return;
         }
-        setRecordings((old) => old.concat([event.message.recording]));
+        refreshRecordingList();
       }),
     );
-  }, [addSubscription, context, context.notificationChannel, setRecordings]);
+  }, [addSubscription, context, context.notificationChannel, refreshRecordingList]);
 
   React.useEffect(() => {
     addSubscription(
@@ -277,12 +277,11 @@ export const ActiveRecordingsTable: React.FC<ActiveRecordingsTableProps> = (prop
         if (currentTarget?.jvmId != event.message.jvmId) {
           return;
         }
-
-        setRecordings((old) => old.filter((r) => r.name !== event.message.recording.name));
         setCheckedIndices((old) => old.filter((idx) => idx !== event.message.recording.id));
+        refreshRecordingList();
       }),
     );
-  }, [addSubscription, context, context.notificationChannel, setRecordings, setCheckedIndices]);
+  }, [addSubscription, context, context.notificationChannel, setCheckedIndices, refreshRecordingList]);
 
   React.useEffect(() => {
     addSubscription(
--- a/src/app/Recordings/ActiveRecordingsTable.tsx
+++ b/src/app/Recordings/ActiveRecordingsTable.tsx
@@ -260,10 +260,10 @@ export const ActiveRecordingsTable: React.FC<ActiveRecordingsTableProps> = (prop
         if (currentTarget?.jvmId != event.message.jvmId) {
           return;
         }
-        refreshRecordingList();
+        setRecordings((old) => old.concat([event.message.recording]));
       }),
     );
-  }, [addSubscription, context, context.notificationChannel, refreshRecordingList]);
+  }, [addSubscription, context, context.notificationChannel, setRecordings]);
 
   React.useEffect(() => {
     addSubscription(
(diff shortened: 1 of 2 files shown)


andrewazores:
Is there anything we can do on the server side to ensure that the recording is deleted first, then immediately recreated (replaced)? This PR hacks around the problem by partially ignoring the state update notifications, but the underlying problem is still there and would affect any other hypothetical clients listening to the same notifications.






---- On Fri, 17 Jan 2025 16:35:12 -0500 ***@***.*** wrote ----



Fixes: #1286

This changes the activeRecordingsTable notification handling logic to call refreshRecordingsList when receiving creation/deletion notifications rather than updating with only the fields in the notification. This ensures that the state of the table remains accurate.

When the query in the linked issue is run a second time it results in the notifications being fired in the following order:

Recording Created
Recording Deleted

Pre-patch the table would see the recording deleted notification last and remove it from the table despite the recording still being present in the target. Calling refreshRecordingsList and getting the state of the target's recordings from the API shows the same.

How to manually test:
Run Cryostat via smoketest or ./mvnw quarkus:dev
Run the following query twice: (Replace 8080 with 8181 if in quarkus:dev mode)

http --follow -v --auth='user:pass' POST :8080/api/v4/graphql
Content-Type:application/json
query='
query {
targetNodes(filter: { annotations: ["PORT = 9091"] }) {
name
target {
doStartRecording(recording: {
name: "test9",
template: "Profiling",
templateType: "TARGET",
replace: "ALWAYS"
}) {
name
state
}
}
}
}'

You can view, comment on, or merge this pull request online at:

  https://github.com/cryostatio/cryostat-web/pull/1535

Commit Summary
d9df210 Fix ActiveRecordingsTable refresh behavior
File Changes 

(1 file)

M src/app/Recordings/ActiveRecordingsTable.tsx (9)
Patch Links:
https://github.com/cryostatio/cryostat-web/pull/1535.patch
https://github.com/cryostatio/cryostat-web/pull/1535.diff

Josh-Matsuoka:
Looking at the server side we wait for deleteRecording to finish already before continuing to recreating the new one:

getActiveRecording(target, r -> r.name.equals(recordingName))
                .ifPresent(r -> this.deleteRecording(r).await().atMost(connectionFailedTimeout));
                
but it looks like this hooks into the existing transaction via QuarkusTransaction.joiningExisting() so the PostRemove hook will be called after PostPersist if I'm understanding correctly? I'm not entirely clear on how Jakarta and quarkus are interacting here.

andrewazores:
Makes sense. I think the best approach to fix this bug is to ensure that the operations and their hooks occur in the expected order. It might require some restructuring of that implementation.


https://quarkus.io/guides/transaction may be useful.






---- On Fri, 17 Jan 2025 17:31:57 -0500 ***@***.*** wrote ----



Looking at the server side we wait for deleteRecording to finish already before continuing to recreating the new one:

getActiveRecording(target, r -> r.name.equals(recordingName))
.ifPresent(r -> this.deleteRecording(r).await().atMost(connectionFailedTimeout));

but it looks like this hooks into the existing transaction via QuarkusTransaction.joiningExisting() so the PostRemove hook will be called after PostPersist if I'm understanding correctly? I'm not entirely clear on how Jakarta and quarkus are interacting here.

andrewazores:
I think there is also another potential avenue for a fix that doesn't involve reordering transactions, so it should be simpler and less invasive to implement. Since 3.0, recordings have not only a name but also a unique database `id` field, as well as a `remoteId` field which mirrors the recording ID assigned by the target JVM instance. I think the WebSocket notifications already include the `id` field, but the web client handling for the notifications was all written long before that when recordings had no IDs and names were used as unique keys instead. I think updating the client's internal modelling of recordings to include the `id`, and then listening for and processing the notifications by using the ID to perform state updates instead of the name, should also fix the problem.

andrewazores (approved):

PR #1557, opened 2025-02-03
fix(grafana): Grafana dashboard uses contextual path and parameters and capability to override URL

andrewazores:
# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/cryostat-openshift-console-plugin/issues/110

## Description of the change:
*This change adds allows a match expression example to be copied to the clipboard...*

## Motivation for the change:

## How to manually test:
2. *...*

### DIFF
--- a/src/app/Recordings/RecordingActions.tsx
+++ b/src/app/Recordings/RecordingActions.tsx
@@ -51,13 +51,18 @@ export const RecordingActions: React.FC<RecordingActionsProps> = ({ recording, u
   const addSubscription = useSubscriptions();
 
   React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


aptmac (approved):

andrewazores:
Hang on, I just double-checked this in smoketest and it isn't working as intended. It works well for the proxy case however.

andrewazores:
I think it's all good now for both smoketest (standalone) and plugin. The capability override should be the only way to override. If there is no override, then the contextual URL and query parameters should not be used since they aren't going to work in the plugin case anyway.

PR #1647, opened 2025-05-30
feat(diagnostics): Add frontend for thread dumps

Josh-Matsuoka:
Related to https://github.com/cryostatio/cryostat/issues/135

Adds an initial frontend for triggering and displaying thread dumps in cryostat. This is done via a new button on the Diagnostics card which changes to a redirect to the diagnostics page once complete, as well as a new diagnostics page with a table for displaying and downloading captured thread dumps.

To test, 

- check out and build the agent PR (https://github.com/cryostatio/cryostat-agent/pull/639)
- Build the quarkus-agent test application with the above version of the cryostat agent
-- TAGS=latest ARCHS=amd64 CRYOSTAT_AGENT_VERSION=0.6.0-SNAPSHOT ./build.bash
- check out and build the main cryostat PR (https://github.com/cryostatio/cryostat/pull/894)
- deploy everything via smoketest
- Open up the diagnostics card for a target with the agent enabled (cryostat-quarkus-agent)
-- Click the thread dump button ("Trigger thread dump"), verify that the thread dumps api is called and a notification is received indicating the thread dump completed
-- Verify that the button changes to redirect you to the diagnostics page
- Go to the diagnostics page, verify that the previously triggered thread dump is available in the table
-- Verify that the last modifed time is correct and that the thread dump can be downloaded/deleted
-- Click the upload button at the top of the table, verify that it triggers a thread dump and it appears in the table
-- Verify that triggering new thread dumps or deleting them from the table refreshes it.


andrewazores:
Review notes from recent call:
1. needs rebase for new nav groups
2. `src/app/Shared/Services/api.utils.ts` needs an update for the WebSocket notification category handling

andrewazores:
Also, some tests would be nice - at least some snapshot tests to make sure the table renders as expected when the mock `ApiService` returns either an empty or non-empty result.

src/app/Shared/Services/api.types.ts
```diff
@@ -537,6 +544,8 @@ export enum NotificationCategory {
   LayoutTemplateCreated = 'LayoutTemplateCreated', // generated client-side
   TargetCredentialsStored = 'TargetCredentialsStored',
   TargetCredentialsDeleted = 'TargetCredentialsDeleted',
+  ThreadDumpSuccess = 'ThreadDumpSuccess',
+  ThreadDumpFailure = 'ThreadFumpFailure',
```
andrewazores, line 548: typo: `Fump` :smile:

andrewazores:
Still failing `prettier` checks ^

src/app/Shared/Services/Api.service.tsx
```diff
@@ -943,6 +1005,12 @@ export class ApiService {
     });
   }
 
+  downloadThreadDump(threadDump: ThreadDump): void {
+    this.ctx.url(threadDump.downloadUrl).subscribe((resourceUrl) => {
+      this.downloadFile(resourceUrl, threadDump.uuid);
```
andrewazores, line 1010: The second argument to `downloadFile` is supposed to help set a different filename for the downloaded file, which I think could be useful here.

```
this.downloadFile(resourceUrl, `${target.alias}_${threadDump.uuid}.thread_dump`);
```

or something might be a better filename for the user than the raw UUID.

andrewazores (commented):
I think we also need something equivalent to the "All Targets Archives" view. As it stands, this UI design only allows the user to view/download/delete thread dumps which belong to a currently discovered target. If that target disappears then there is no more way via the UI to ever see those thread dumps to download or delete them.

Josh-Matsuoka:
CI doesn't seem to like toBeCalledWith, it's passing locally for me and the usage is consistent with other uses of it elsewhere so I'm not entirely sure what's wrong

src/app/routes.tsx
```diff
@@ -156,6 +158,18 @@ const flightRecorderRoutes: IAppRoute[] = [
...
 
+const diagnosticsRoutes: IAppRoute[] = [
+  {
+    component: Diagnostics,
+    label: 'Thread Dumps',
+    path: toPath('/diagnostics'),
+    title: 'Thread Dumps',
+    description: 'Create and view thread dumps on single target JVMs.',
+    navGroup: DIAGNOSTICS,
+    navSubgroup: CAPTURE,
```
andrewazores, line 169: I think this should actually be under the `ANALYZE` subgroup, since the Thread Dump table located here is analogous to the (JFR) Archives tables. Diagnostics > Capture would be a view that allows the user to request a new thread dump to be collected, so it would probably just look a lot like a scaled-up version of the dashboard card you've implemented.

src/app/Diagnostics/Diagnostics.tsx
```diff
@@ -0,0 +1,68 @@
...
+    [navigate, pathname, search],
+  );
+
+  const cardBody = React.useMemo(
+    () => (
+      <Tabs id="threadDumps" activeKey={activeTab} onSelect={onTabSelect} unmountOnExit>
+        <Tab
+          id="threadDumps"
+          eventKey={DiagnosticsTab.THREAD_DUMPS}
+          title={<TabTitleText>Thread Dumps</TabTitleText>}
```
andrewazores, line 49: This string should be extracted out/localized

Josh-Matsuoka:
> Also, when downloading a thread dump file, it currently just gets saved with the (UU)ID as the filename. I see that you have some logic in Api.service's downloadThreadDump that tries to set the filename like {target.alias}_{thread_dump.id}.thread_dump, but it seems not to work for me. You might try setting [the Content-Disposition response header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition) on the backend to get the browser to behave here.

The content-disposition header is already set in the backend, it follows the same structure as the archived recordings download.

Looking into this further the filename is set correctly in the frontend, anchor.download gets set to it, however looking at the request that gets sent to the download endpoint and it's headers, the filename doesn't seem to be specified anywhere. 

The query parameter that would set the filename is thus null when it makes it to the backend:

cryostat-1                | 2025-08-01 17:12:59,901 WARN  [io.cry.dia.Diagnostics] (executor-thread-8) Handling download Request for key: (IX6TpNMZWQSweVAWEtKTrm_j0OtjGZhC_EoJv-d-o6Y=,5e42ed2f-7ca3-4f33-b3ee-60348ef45060)
cryostat-1                | 2025-08-01 17:12:59,901 WARN  [io.cry.dia.Diagnostics] (executor-thread-8) Handling download Request for query: null

So it defaults to using the thread dump id as the filename.

Josh-Matsuoka:
@andrewazores any idea what's causing the above? ^

andrewazores:
If I open my browser's devtools and watch the network traffic when clicking on the Dashboard card item to trigger a thread dump, I see 1 request to perform the dump and then 3 identical requests in rapid succession to query available dumps (`GET https://localhost:8443/api/beta/diagnostics/targets/2/threaddump`). The response to the first query does not contain the new dump, and the second two responses do. It seems like only the final query should actually be sent.

Also, if I click the "Invoke Garbage Collection" button on the card instead, then there are also two identical threaddump `GET` queries sent.

I suspect something is not quite right with some React hooks on the Diagnostics dashboard card, causing it to re-render and re-query state too often.

src/app/Dashboard/Diagnostics/DiagnosticsCard.tsx
```diff
@@ -58,6 +64,30 @@ export const DiagnosticsCard: DashboardCardFC<DiagnosticsCardProps> = (props) =>
...
+          filter((target) => !!target),
+          first(),
+          concatMap(() => serviceContext.api.getThreadDumps()),
+        )
+        .subscribe({
+          next: (dumps) => (dumps.length > 0 ? setThreadDumpReady(true) : setThreadDumpReady(false)),
+          error: () => setThreadDumpReady(false),
+        }),
+    );
+  });
```
andrewazores, line 81: This hook is missing a dependencies array - it's likely the cause of, or related to, the multiple queries bug I described above.

andrewazores (approved):

PR #1825, opened 2025-09-26
feat(diagnostics): Add labels support to diagnostics views

Josh-Matsuoka:
Fixes: https://github.com/cryostatio/cryostat/issues/1072

Depends on: https://github.com/cryostatio/cryostat-web/pull/1809

TODO: Test label filtering for heap dumps, extend to thread dumps, add/adjust tests, merge/refactor once 1809 is merged.

## Description of the change:
- Built on top of #1809 
- Adds labels support to the Thread and Heap Dumps tables along the same lines as Archived Recordings

## How to manually test:
- Pull and build this PR as well as the backend PR (https://github.com/cryostatio/cryostat/pull/1073)
- bash smoketest.bash -O -t quarkus-cryostat-agent -A
- Trigger a thread and heap dump from the diagnostics card or capture/diagnostics page
- Navigate to the thread/heap dumps table, add/remove labels.
- Test that label filtering and deletion works


Josh-Matsuoka:
Found an issue with adding labels through the all-targets view after rebasing, deletion still works but adding labels doesn't even trigger a graphql request, so it looks like something's wrong with the filtering that gets done to determine which heap dumps need to update their metadata.  I'll push it once I've fixed that.

src/app/Diagnostics/BulkEditHeapDumpLabels.tsx
```diff
@@ -59,28 +60,34 @@ export const BulkEditHeapDumpLabels: React.FC<BulkEditLabelsProps> = ({ checkedI
...
-        tasks.push(context.api.postHeapDumpMetadata(r.heapDumpId, updatedLabels).pipe(first()));
-      }
-    });
     addSubscription(
-      forkJoin(tasks).subscribe({
-        next: handlePostUpdate,
-        error: handlePostUpdate,
+      propsTarget.subscribe((t) => {
+        if (!t) {
+          return;
```
andrewazores, line 68: Minor nitpick, I think this would be more idiomatic:

```typescript
propsTarget
  .pipe(filter(t => !!t))
  .subscribe((t) => { // t is truthy here
```

andrewazores:
Functionality between this and #1073 looks great.

PR #2052, opened 2026-01-24
feat(triggers): Add frontend for smart triggers

Josh-Matsuoka:
# Welcome to Cryostat! 👋

Related to: https://github.com/cryostatio/cryostat-agent/issues/225

Depends on: https://github.com/cryostatio/cryostat-agent/pull/780 https://github.com/cryostatio/cryostat-core/pull/625 https://github.com/cryostatio/cryostat/pull/1250

## Description of the change:
Adds a frontend showing active smart triggers for a target

## Motivation for the change:

## How to manually test:
- Check out and build/run all dependent PRs
- Start a target with the newly built agent instance
- Test that existing smart triggers are picked up by the UI
- Test the functionality of the triggers UI (Display, creation, deletion)

Opening as draft until tests are added.


andrewazores (commented):
Image

Image

We have two different page layouts here - the one that the new Triggers view is using is similar to what we have in Archives-related views, with tabs within the main content card and the possibility of selecting an "All Archives" state which is distinct from "All Targets" or "Specific Target". Since Smart Triggers are transient (in the sense that they just live in the target JVM at runtime, there is no database/object storage for them), there is no such view state as All Targets or All Archives, so I think it should be using the same layout as Events or (Active) Recordings.

Also, after I create a Smart Trigger, if I click the Create button again to bring the modal back up to define a second trigger, the previous trigger's contents are still in the text field. If I navigate away first and then come back then the modal starts with a blank text field. This looks like the modal's state is just not being cleared when the Submit/Cancel buttons are clicked.

I think it would be useful to have some kind of information or visualization about what the available metrics are and the expression syntax, since they are currently not discoverable - as a user, how will I know what to type into this text field? Check out how the Automated Rules form provides hints to the user (and this is also not really as detailed as I would like it to be, still).

And,

Image

there needs to be a better Agent detection handling than this error state if a non-Agent target is selected. Check how the Instrumentation view handles it perhaps. The Trigger creation modal then wouldn't need to include text about only being available for Agent targets, since that would be implied by the locked-out state when a non-Agent target is selected:

Image

Josh-Matsuoka:
> Also, do we need to have a column for the Expression when we also have the parsed components (condition, duration constraint, template)? I think just displaying the components should be enough - and I would put them in the order of condition, then duration, then template.

I was thinking from the POV of a user with less experience with smart triggers, showing the combined expression would make it clear how everything fits together, I can see it being unnecessary though

andrewazores (approved):

PR #2155, opened 2026-03-17
feat(triggers): Smart triggers form improvements

### LINKED ISSUE #2092: [Task] SmartTriggers creation form
Couple more minor nitpicks:
1. in the Smart Trigger creation modal, a link to the CEL documentation (like what the Rules creation form has) would be nice, as well as a sample expression to reference. And for a later enhancement, some way to list the different MBean attributes that are available to use. Maybe also have a link to the Events view so the user can see which (Target) Templates are available. Later down the line we can rework this from a plain text field into a form like how Recordings are created, and hide the details of CEL from the user.
2. when visiting the Trigger view with a non-Agent target selected, the new empty state lock out is good, but the HTTP 400 Bad Request notifications still appear. These should be suppressed
3. if the user creates a trigger definition with no duration clause, maybe that cell in the table should explicitly say something like "None" or "N/A", or even "Immediate", rather than just being blank? Also, do we need to have a column for the Expression when we also have the parsed components (condition, duration constraint, template)? I think just displaying the components should be enough - and I would put them in the order of condition, then duration, then template.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-web/issues/2052#issuecomment-3900097615_
            

For 1, rather than just a link to the Events view, if this becomes a proper trigger creation form then this can reuse the same target template selector component which is used in the Recording and Rules creation forms, except in this case it needs to filter for only Target template types since those are the only ones that Smart Triggers can currently use.

Josh-Matsuoka:
# Welcome to Cryostat! 👋

Fixes: [#<issue number>](https://github.com/cryostatio/cryostat-web/issues/2092)

## Description of the change:

- Adds a template selection component to the smart triggers creation form, when selected this will abstract away that component of the smart trigger definition from the user.
- Adds a list of available MBeans (TODO: Agent Endpoint to query available mbeans to pick up custom mbeans, currently just lists the ones available via mbean metrics)
- Selecting an mbean from the human readable list shows the name used for smart trigger definitions.

Opening as draft while I resolve an issue with template selection and update tests


locales/en/public.json
```diff
@@ -863,7 +863,9 @@
...
     "DEFINITION_HINT": "The set of constraints and target duration must be separated by a semicolon (;) character. Each constraint must include: the name of an MBean counter; a relational operator such as > (greater than), = (equal to), <(less than), and so on; and a specified value. The type of relational operator and value that you can specify depends on the associated MBean counter type. The constraint and duration must be enclosed in square brackets, followed by a tilde then the Recording Template name.",
-    "DEFINITION_HINT_BODY": "For an example definition: [ProcessCpuLoad>0.2;TargetDuration>duration(\"30s\")]~Profiling"
+    "DEFINITION_HINT_BODY": "For an example definition: [ProcessCpuLoad>0.2;TargetDuration>duration(\"30s\")]",
+    "TEMPLATE_SELECT": "Select an Event Template for starting a recording when the Trigger Condition is met.",
+    "AVAILABLE_MBEANS": "Select an Mbean to check the value of."
```
andrewazores: I think the canonical, official spelling is "MBean"

andrewazores:
Screenshot_2026-04-16_17-04-01

What if after selecting an MBean metric name like "Process CPU Load" from the dropdown menu, the user doesn't need to type in `[ProcessCpuLoad]` anymore? There's already helper text telling them what the metric name is, so can't we hide more of the expression construction away? There could just be an MBean metric selector, then a "comparator" selector with options for `less than (<)| less than or equal (<=) | equal (=) | greater than or equal (>=) | greater than (>)` and then a small text box that allows the user to enter a numeric value like `0.5`. Also, this implies the only MBean metrics available for selection should be numeric ones, but I think that's already the only reasonable kind of trigger expression that a user would write. I can't image it being sensible for a Smart Trigger to be defined based on the `OS Name`, `OS Arch`, or `Library Path` "metrics", for example.

Then there could also be some UI component(s) for the `TargetDuration`, for example a checkbox "only if this is continuously observed for:" and a duration picker similar to how we set Flight Recording lengths.

This way the only thing the user is manually typing in is the value of the MBean metric they want to compare against, and everything internal about how we format the Smart Trigger as a string is hidden away.

andrewazores:
> Also, this implies the only MBean metrics available for selection should be numeric ones, but I think that's already the only reasonable kind of trigger expression that a user would write. I can't image it being sensible for a Smart Trigger to be defined based on the OS Name, OS Arch, or Library Path "metrics", for example.

https://github.com/cryostatio/cryostat-web/blob/96aefce2b2f021cc9cbda80d8e4fed75ffd11450/src/app/Dashboard/Charts/mbean/MBeanMetricsChartCard.tsx#L168

I guess the definitions for the MBean Metrics dashboard cards can be used here, or some refactoring can be done so that these two things share a common set of usable metrics.

andrewazores (dismissed):

andrewazores:
Works really nicely now, much more polished and usable feeling experience. Just rebase please and we'll get this merged.

andrewazores:
Actually I've just noticed a minor bug - if you enter a value like `1`, that gets left as the literal `1` which then gets passed to the Agent verbatim, which parses it as an int instead of a double, and the CEL `>` operator fails to compare the double metric to the int value. I think all of the LHS MBean metrics are doubles, so the frontend and the backend (Agent) should both be looking for the RHS appearing to be formatted as an int and either reformatting or casting it so that CEL can evaluate it.

Josh-Matsuoka:
From my other PR showing the return types:

Screenshot From 2026-04-16 17-15-12

It looks like just the Operating System metrics are doubles. We can just check and format the input accordingly for these 3, the rest are longs and ints

src/app/Triggers/SmartTriggers.tsx
```diff
@@ -653,18 +712,49 @@ export const CreateSmartTriggersModal: React.FC<CreateSmartTriggersModalProps> =
...
+    var durationExpr = '';
+    var formattedExpr = expressionInput;
+    if (formData.duration != 0) {
+      durationExpr = ';TargetDuration>duration("' + formData.duration + formData.durationUnit + '")';
+    }
+    // These 3 attributes require a double, make sure it gets formatted correctly
+    if (
+      mbeanSelectValue == 'ProcessCpuLoad' ||
+      mbeanSelectValue == 'SystemCpuLoad' ||
+      mbeanSelectValue == 'SystemLoadAverage'
```
andrewazores: `HeapMemoryUsagePercent` is also a double - actually I think a more accurate name would have been Ratio rather than Percent, since it's going to have values like `0.314`.

Rather than hardcoding this list of string of MBean metric names, which are duplicated hardcoded values as in the MBeanOptions list above, maybe this can be refactored like:

```typescript
const MBeanOptions = [
  { value: 'ProcessCpuLoad', label: 'Process CPU Load', type: 'double' }
];
...
const handleSubmit = React.useCallback(() => {
  const opt = getOptionByName(mbeanSelectValue);
  if (opt.type === 'double') {
    // check for '.' and append .0 if none
  }
});
```

I see that `DaemonThreadCount` already has a `type: 'int'` attribute, so it looks like the seed of this idea is already there - it would make sense to flesh it out.

src/app/Triggers/SmartTriggers.tsx
```diff
@@ -636,13 +649,59 @@ export interface CreateSmartTriggersModalProps {
...
+    { value: 'Uptime', label: 'VM Uptime', disabled: false },
+  ];
+
+  const comparatorsOptions = [
+    { value: '>', label: 'Greater Than (>)', disabled: false },
+    { value: '>=', label: 'Greater Than/Equal To (>=)', disabled: false },
+    { value: '==', label: 'Equal To', disabled: false },
+    { value: '<=', label: 'Less Than/Equal To (>)', disabled: false },
+    { value: '<', label: 'Less Than (>)', disabled: false },
+  ];
```
andrewazores, line 713: What's the plan for these `disabled: false` attributes on these comparators and the MBeanOptions? If they're all hardcoded to `false` then it seems like it can just be removed and the component that references this property can just not be set disabled. Is there a plan in the future to add metrics or comparators that will be displayed but cannot be selected and therefore the UI is disabled... ?

src/app/Triggers/SmartTriggers.tsx
```diff
@@ -686,12 +852,42 @@ export const CreateSmartTriggersModal: React.FC<CreateSmartTriggersModalProps> =
...
+        <FormSelect
+          value={mbeanSelectValue}
+          onChange={onMbeanChange}
+          aria-label="MBean Input"
+          ouiaId="BasicFormSelect"
+          placeholder="Select an MBean Attribute"
+        >
+          <FormSelectOption key={-1} value={''} label={'Select an MBean Attribute'} isDisabled />
+          {MBeanOptions.map((option, index) => (
+            <FormSelectOption isDisabled={false} key={index} value={option.value} label={option.label} />
```
andrewazores: Looking good. Nitpicking here, but I think there's no need to set `isDisabled={false}` - the component will be enabled by default, so this prop can just be left out.

src/app/Triggers/SmartTriggers.tsx
```diff
@@ -686,12 +852,42 @@ export const CreateSmartTriggersModal: React.FC<CreateSmartTriggersModalProps> =
...
+        <FormSelect
+          value={comparatorSelectValue}
+          onChange={onComparatorChange}
+          aria-label="Comparator Input"
+          ouiaId="BasicFormSelect"
+          placeholder="Select a Comparator"
+        >
+          <FormSelectOption key={-1} value={''} label={'Select a Comparator'} isDisabled />
+          {comparatorsOptions.map((option, index) => (
+            <FormSelectOption isDisabled={false} key={index} value={option.value} label={option.label} />
```
andrewazores: Same here.

andrewazores:
Looks good now, just the one last little nitpick above, and the CI linting failures to address.

andrewazores (approved):

PR #2172, opened 2026-03-26
feat(archives): use audit log to enhance sub-table display names

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Fixes: #2168
Based on #2137 
Depends on #2137

## Description of the change:
Uses the audit log backend (if available) to enhance the All Archives view. Without the audit log backend, the All Archives table is rather limited in what information about the source target it can display - the JVM hash ID is still known, but not much else. Prior work has implemented an info `(i)` button on the sub-table in the view to retrieve audit log information about this `jvmId` so that the user can click this button to pull up a JVM details modal to display the connection URL, labels, annotations, discovery lineage, etc. information. This PR extends that idea so that the table itself automatically populates the sub-table names using audit log information if available.

## Motivation for the change:
Allows the user to see more information about source targets in the All Archives view without needing to click the `(i)` button on each to get more information. If the user is looking for archives from a particular target (ie a particular k8s Pod), this change makes it much quicker and easier for the user to locate that data.

## How to manually test:
1. Check out this PR and either build it into a Cryostat image, or use `yarn start:dev`
2. Deploy an environment ex. `smoketest.bash -t quarkus-cryostat-agent`
3. Collect some archived recordings or wait for the sample app to push some
4. Go to the JFR All Archives view

PR #2156, opened 2026-03-18
feat(audit): implement query UI for audit logs

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Fixes: #2067
Depends on cryostatio/cryostat#1423

To manually test, see cryostatio/cryostat#1423

PR #2179, opened 2026-04-06
feat(archives): implement time-range and lineage filter

andrewazores:
# Welcome to Cryostat! 👋

Based on #2172
Depends on #2172
See cryostatio/cryostat#893

## Description of the change:
Extends the idea in #2172 further, as well as borrowing the time range filter idea from #2156. This adds All Archives filters to all three similar tables (JFR, Thread Dump, and Heap Dump archives). There is a new filter for the archive timestamp so users can filter results to only display archives that were captured within a particular timeframe, and a new filter to only display results from the new "lineage labels". These look and behave similarly to recording labels. By clicking a lineage label on an archived data file the user can narrow the selection to only show results that share that same label - for example, to only show results from the same k8s Namespace, or from the same Deployment. These time and lineage filters persist and are shared across all three tables, so if the user is looking for some specific JFR archive results then it's also quick and easy to switch over to looking for thread or heap dumps that match the same search criteria.

## Motivation for the change:
This makes it much quicker and easier to search and filter archives by discovery details so that looking for archived data captured from a particular Deployment or Pod is a better experience.

## How to manually test:
2. *...*


jtolentino1 (approved):
LGTM. Tested locally and the archive filtering UI and logic looks good in archives and diagnostics under all archives

PR #2192, opened 2026-04-15
feat(triggers): Add Mbean selector to Trigger Creation Form

### LINKED ISSUE #2092: [Task] SmartTriggers creation form

Josh-Matsuoka:
# Welcome to Cryostat! 👋

Fixes: [#<issue number>](https://github.com/cryostatio/cryostat-web/issues/2092)

Depends on: https://github.com/cryostatio/cryostat-web/pull/2155 https://github.com/cryostatio/cryostat-core/pull/667 https://github.com/cryostatio/cryostat-agent/pull/849 https://github.com/cryostatio/cryostat/pull/1467

## Description of the change:
- Adds an Mbean selector component to the Trigger Creation Form displaying all currently available Mbean attributes. 
- Currently filters them to only the ones available through mbeanMetrics since internally that's currently what Smart Triggers are evaluated against. In future if/once they're extended to check against generic mbean attributes we can unfilter them.

## How to manually test:
2. *...*


src/app/Shared/Services/api.types.ts
```diff
@@ -445,6 +445,21 @@ export interface SmartTrigger {
...
+  name: string;
+  type: string;
+  description: string;
+  parentBean: string;
+  isReadable: boolean;
+  isWritable: boolean;
+  isIs: boolean;
+}
+
+export interface MbeanAttributeMap {
```
andrewazores, line 458: nit: `MBean`, capital B

src/app/Triggers/SmartTriggers.tsx
```diff
@@ -633,16 +646,70 @@ export interface CreateSmartTriggersModalProps {
   onAccept: (s: string) => void;
 }
 
+export interface MbeanSelectorOption {
```
andrewazores, line 649: same here

src/app/Triggers/SmartTriggers.tsx
```diff
@@ -633,16 +646,70 @@ export interface CreateSmartTriggersModalProps {
...
+      'ThreadCount',
+      'DaemonThreadCount',
+      'Arch',
+      'AvailableProcessors',
+      'Version',
+      'SystemCpuLoad',
+      'SystemLoadAverage',
+      'ProcessCpuLoad',
+      'TotalPhysicalMemorySize',
+      'FreePhyiscalMemorySize',
```
andrewazores, line 684: typo? Phyiscal -> Physical

src/app/Triggers/SmartTriggers.tsx
```diff
@@ -676,6 +743,66 @@ export const CreateSmartTriggersModal: React.FC<CreateSmartTriggersModalProps> =
...
+    if (template && template.name && template.type) {
+      return `${template.name},${template.type}`;
+    }
+    return '';
+  }, [formData]);
+
+  const MbeanOptions = React.useMemo(() => {
+    var attributes: MbeanSelectorOption[] = [];
+    mbeans.forEach((m: MbeanAttributeMap) => {
+      m.attributes.forEach((i: MBeanAttributeInfo) => {
```
andrewazores, line 781: If `supportedTriggerAttributes` is just a hardcoded list of the MBeans we expect to see present in the remote JVM, then do we even need to send a query at this point... ? It seems like just returning the supported attributes list would do the same thing more simply, until we actually have some wider support on the backend (agent side) for processing other MBean metrics in the CEL expressions. Or am I missing something here?

andrewazores (commented):
Just a little confused about the MBean query list and filtering for the supported attributes.

src/app/Triggers/SmartTriggers.tsx
Josh-Matsuoka: You're correct that hardcoding it would achieve the same thing, I wanted to lay the groundwork here for generic mbean attributes down the line but for now we can get rid of the query if you'd prefer and just hardcode the list options.

andrewazores: If it's currently functionally equivalent then I think for now let's just go with the hardcoded list so that we can get that UI enhancement in, and work on the broader-scoped change that affects multiple components when we're further out from a dev freeze. Maybe the hardcoded version of this can go directly into #2155 and this PR can build on top of that do to the enhanced filtering based on actual live data pulled from the Agent.

Josh-Matsuoka: Sounds good, I'll add it to that PR

PR #2232, opened 2026-05-05
fix(diagnostics): prefill selected thread dump for analysis

### LINKED ISSUE #915: [Bug] Thread Dumps views misnamed/missing
### Current Behavior

Image

The selected plugin view is named "Analyze Thread Dumps", but it should be named "Thread Dump Archives" to match Cryostat's own UI.

Image

Clicking the "Analyze" overflow action in the Cryostat UI routes the user to the (proper) Analyze Thread Dumps view with the selected thread dump pre-selected. Clicking the same action in the plugin UI results in a console 404 page.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```Markdown
- OS:
- Environment:
- Version:
```

### Anything else?

_No response_

jtolentino1:
# Welcome to Cryostat! 👋

Fixes: cryostatio/cryostat-openshift-console-plugin#915

## Description of the change:
This change preserves and consumes the selected archived thread dump when navigating to the thread dump analysis page.

Changes:
- Add `jvmId` and `threadDumpId` query parameters when clicking **Analyze Thread Dump** from the thread dump table row action
- Keep the existing React Router `location.state` prefill path, and add query parameter fallback support for environments
where router state is not preserved
- Make `ThreadDumpSelector` reflect the selected dump from props instead of only using its initial local state
- Avoid clearing the selected dump/analysis result when the target service emits the same target during initial page load
- Add focused tests for location-state prefill, query-param prefill, and row-action navigation

## Motivation for the change:
In the OpenShift Console plugin, clicking **Analyze Thread Dump** from Thread Dump Archives navigates to the analysis page, but the selected dump is not pre-filled. Users then have to manually select the same dump again before analysis runs.

Passing the dump identifiers through the URL and making the analysis page consume them makes the flow work reliably in both standalone Cryostat Web and embedded OpenShift Console routing.

## How to manually test:
1. Start Cryostat with at least one discoverable target.
2. Create a thread dump for the target.
3. Navigate to **Thread Dumps**.
4. Open the row action menu for the archived thread dump and click **Analyze Thread Dump**.
5. Verify the app navigates to **Analyze Thread Dumps**.
6. Verify the thread dump selector is already populated with the selected dump ID.
7. Verify the analysis result loads automatically without manually selecting the dump.
8. In the OpenShift Console plugin, repeat the same flow from **Thread Dump Archives** and verify there is no 404 and no manual selection is required.

### DIFF
--- a/src/app/Diagnostics/Analysis/ThreadDumpAnalysis.tsx
+++ b/src/app/Diagnostics/Analysis/ThreadDumpAnalysis.tsx
@@ -62,6 +62,9 @@ import { ThreadDumpSelector } from './ThreadDumpSelector';
 
 export interface ThreadDumpAnalysisProps {}
 
+const isSameTarget = (a: NullableTarget, b: NullableTarget): boolean =>
+  a?.connectUrl === b?.connectUrl && a?.jvmId === b?.jvmId;
+
 interface ThreadRowData {
   threadInfo: ThreadInfo;
   isExpanded: boolean;
@@ -247,9 +250,13 @@ export const ThreadDumpAnalysis: React.FC<ThreadDumpAnalysisProps> = ({ ...props
   React.useEffect(() => {
     addSubscription(
       context.target.target().subscribe((t) => {
-        setTarget(t);
-        setAnalysisResult(undefined);
-        setSelectedThreadDump('');
+        setTarget((currentTarget) => {
+          if (currentTarget && !isSameTarget(currentTarget, t)) {
+            setAnalysisResult(undefined);
+            setSelectedThreadDump('');
+          }
+          return t;
+        });
       }),
     );
   }, [addSubscription, context.target, setTarget]);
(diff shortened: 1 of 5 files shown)


andrewazores:
I think there's something funny that happens (maybe already pre-existing, or maybe new to this PR) when moving from the thread dump archives table to the analysis view in this workflow:

1. Set up `smoketest.bash -t quarkus-cryostat-agent` or probably any other scenario with 2+ discovered targets
2. Use the Diagnostics > Capture view to capture thread dumps from at least 2 targets. Take note of which target was selected last and note the captured thread dump IDs
3. Go to Diagnostics > Analyze > Thread Dump Archives > All Targets, expand the sub-table of the _other_ target (not the one that is currently selected in ex. the target context selector from the Capture view)
4. Click the action menu item to "Analyze Thread Dump" on this other target's thread dump file
5. The UI navigates to the analysis view and the thread dump ID selector dropdown reflects the correct dump ID, but the context selector still has the original target selection. Clicking the thread dump ID selector reveals that the current value is not a selectable option

Screenshot_2026-05-05_15-40-08

image

image

image

I'm also noticing now that there is no loading spinner or other loading indication - when I click on the "Analyze Thread Dump" action menu item and get navigated away to the analysis view, there's a noticeable delay where the view just shows its "Select a thread dump to analyze" empty state before suddenly updating with the analysis contents.

jtolentino1:
@andrewazores   

Thanks, good catch. I think this is exposed by this PR...

I'll do some troubleshooting and put up a fix

andrewazores (approved):

andrewazores:
@Mergifyio backport cryostat-v4.2

PR #2311, opened 2026-06-16
feat(diagnostics): Add frontend for Heap dump analysis

### LINKED ISSUE #1127: [Story] Heap Dump analysis
### Describe the feature

Heap Dump analysis almost certainly needs to be performed on the Cryostat server side, and some analysis result provided to the client via an API endpoint.

The easiest path is probably to use what JMC exposes, since we already consume JMC as a dependency anyway:

Image

https://github.com/openjdk/jmc/blob/master/application/org.openjdk.jmc.joverflow/src/main/java/org/openjdk/jmc/joverflow/heap/parser/HeapDumpReader.java

https://github.com/openjdk/jmc/blob/master/application/org.openjdk.jmc.joverflow/src/main/java/org/openjdk/jmc/joverflow/heap/model/Snapshot.java

Analysis can be requested on an existing heap dump file located in Cryostat's storage. Cryostat would then download that file from storage and use these JMC classes to perform the analysis and return the result as some JSON structure. Cryostat's web-client UI would implement some simple display for these analysis results, perhaps something similar to the JFR Automated Analysis Reports UI. The Cryostat backend can also do some caching of the analysis result, similar to what it does for JFR Automated Analysis Reports.

### Anything other information?

_No response_

Josh-Matsuoka:
# Welcome to Cryostat! 👋

Fixes: cryostatio/cryostat#1127

## Description of the change:

Adds a frontend for visualization of heap dump data.

## How to manually test:
1. Pull and build with all related PRs
2. Trigger a heap dump in an agent target
3. Navigate to Analyze Heap Dumps, check the visualizations

~~Depends on cryostatio/cryostat#1447~~


andrewazores:
Screenshot_2026-06-17_16-14-33
Screenshot_2026-06-17_16-14-09

1. "Diagnostics > Analyze > Heap Dumps" should probably be renamed "Diagnostics > Analyze > Heap Dump Archives" to be symmetric with Thread Dump Archives. The page titles should also match - currently on the "Thread Dumps Archives" view, the title above the tabs just says "Thread Dumps".
2. The navigation layout being "thread dump archives, analyze thread dumps, analyze heap dumps, heap dump archives" seems odd. Maybe "thread dump archives, analyze thread dumps, heap dump archives, analyze heap dumps"?
3. The heap dump archives table is missing the action menu item to jump to the analysis view

andrewazores:
The right side legends of these charts seem to be cut off with no way to view the full label text, so it's not very meaningful as it stands. Maybe the charts should be left-justified within their container cards so that there's more space to the right for the labels, and the container cards could also be full page width.

image

For the `quarkus-cryostat-agent` sample application, which is a very small test case, this view is incredibly long and would probably benefit from some pagination - even if it's just client-side. Having all of the content in browser memory is probably not terrible, but making the browser render all of that content and having such a long scrollable viewport is not ideal and makes it rather sluggish to respond to click events.

image

Same here regarding length/scrolling/pagination. Also, I like how the Problem Fields tab has sortable columns. I think the Histogram should, too. The columns names here are also truncated and there seems to be no way to see what the full column name actually is. If these columns are in bytes, which I think they are, it might also be good to use the `formatBytes` utility to make it a little more readable.

image

Same thing here regarding pagination and sortable columns and maybe `formatBytes` (unclear currently, I can't tell what the displayed values are).

image

image

image

High Size Objects is pretty good, same as Problem Fields, in that it's sortable and the column names are hoverable to see the full name, but it probably could use pagination and `formatBytes` still.

image

Weak Hashmaps don't show up in this simple test so the same comments above may apply here as well.

image

andrewazores:
Maybe along with pagination some text filtering would also be nice. The Event Types table is a good template to follow for what I'm hoping to see.

andrewazores:
Can we have some descriptive text or some tooltips to explain certain things? Maybe users who are familiar with investigating heap dumps on a regular basis or who've used JMC (or maybe other similar tools) to analyze heap dumps before will understand what all of this means, but I don't think looking at this view it's really obvious what problem we're reporting to the user:

image

What is the problem with these fields, or why are they problems? What is the overhead of a problem field? Is there a reason to even render results for classes with the NO_FIELDS result type?

image

andrewazores:
image

It still seems to me that these charts are not utilizing the available screen space effectively - for example, there are very many pie chart slices on the top chart, but they nearly all have a truncated label in the legend causing them to look like they have identical names: `jdk.internal.reflect.DelegatingClassLoader@0x5f62`.

image

Looks like a broken translation key.

image

This tab has a different layout from the rest - there is no `Collections` bold text title above the table. I think this is actually fine - there is already a tab title, so the bold text (seen in Object Histogram above or Problem Fields below) is redundant.

image

image

Expanding the Duplicate Strings row items leads to a similar problem as before - the page now has hundreds or thousands of nested row elements to render, so the UI render performance takes a significant hit and the page becomes incredibly long to scroll. This inner element should probably also be paginated.

And I see that you added tooltip components in a prior commit, but for some reason I'm not finding how to actually get the tooltips to display in the UI. I thought hovering my cursor over ex. the Problem Fields cells in the Problem Type column would do it, but it still just displays ex. `SOME_FIELDS_EMPTY` with no apparent explanation of what this means.

Josh-Matsuoka:
I've moved the charts to their own Tabs and changed the offsets/rendering ratio so the legend should be more legible, the legend is also now sorted as well.

Dup Strings now also has a subtable with filtering/pagination.

There were also a few additional types of problem fields given by the analysis that now have their own tabs

andrewazores:
Awesome, that last update looks and works much better. Please rebase.

Josh-Matsuoka:
Pushed an update to handle async report generation.

The endpoint in the server now returns either a job ID or the JSON report depending on the caching state (Similar to how the existing reports endpoint functions). I've added a heap dump reports service along the same lines as the report service to handle tracking job IDs and retrieving reports.

In practice nothing changes for the user, the view will be loading until the report is generated and retrieved then it will update accordingly. If you watch the network requests you'll see 2: the first gets a 200 response to the initial report request containing a job ID and a little later a second request getting a 202 containing the report once the notification fires.

src/app/Shared/Services/HeapDumpReport.service.tsx
```diff
@@ -0,0 +1,233 @@
...
+              }
+            }),
+            tap({
+              next: (report) => {
+                try {
+                  sessionStorage.setItem(this.analysisKey(jvmId), JSON.stringify(report));
+                  sessionStorage.setItem(this.analysisKeyTimestamp(jvmId), Date.now().toString());
+                } catch (err) {
+                  if (isQuotaExceededError(err)) {
+                    this.notifications.warning('Report Caching Failed', err.message);
```
andrewazores: I tried running through a smoketest scenario in a private browsing window (so localstorage and sessionstorage should be as empty as possible), and even with the small quarkus-cryostat-agent test application and a ~200MiB heap, it seems I always hit this case where the report exceeds the browser's sessionstorage quota. It seems like in practice these report documents are probably always going to be too large to cache this way.

andrewazores (approved):

